PR #218, opened 2023-03-31
Add support for stopping Scans to Orchestrator

chrisgacsal:
## Description

* Add `ScanWatcher` to look up for Scans which status has changed to `Aborted` and updated all the corresponding `ScanResult` status to `ABORTED` so the Scanner Worker and the CLI could detect cancellation events.
* Update Scanner Worker to detect if the `ScanResult `state is changed to `Aborted`.

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

## Test Plan

Manual steps to test aborting ongoing scans in live environment.

<details>
  <summary>Expand</summary>

### Scan configuration

Create Scan Configuration file

```shell
cat <<EOF > scanconfig.json
{
  "name": "test",
  "scanFamiliesConfig": {
     "sbom": {
       "enabled": true
     },
     "vulnerabilities": {
       "enabled": true
     },
     "exploits": {
       "enabled": true
     }
  },
  "scheduled": {
    "cronLine": "0 */4 * * *",
    "operationTime": "2023-01-20T15:46:18+00:00"
  },
  "scope": {
    "allRegions": true,
    "objectType": "AwsScanScope",
    "instanceTagSelector": [
      {
        "key": "ScanConfig",
        "value": "test"
      }
    ]
  }
}
EOF
```

Apply Scan Configuration to API

```shell
curl -sSf -X POST 'http://localhost:8888/api/scanConfigs' -H 'Content-Type: application/json' \
  -d @scanconfig.json \
| jq -r -e '.id' > scanconfig.id
```

Get Scan Configuration object from API

```shell
curl -sSf -X GET 'http://localhost:8888/api/scanConfigs/'"$(cat scanconfig.id)"'' \
| jq -r -e '.' > scanconfig.api.json
```

### Start Scan

Start Scan using Scan Config

```shell
jq -r -e '{maxParallelScanners, name, scanFamiliesConfig, scheduled, scope} | .scheduled.operationTime = (now|todate)' \
  scanconfig.api.json \
| curl -sSf -X PUT -H 'Content-Type: application/json' 'http://localhost:8888/api/scanConfigs/'"$(cat scanconfig.id)"'' \
  -d @-
```

Get ongoing Scan from API using ScanConfig id 

```shell
curl -sSf -G 'http://localhost:8888/api/scans' \
  --data-urlencode "\$filter=scanConfig/id eq '$(cat scanconfig.id)' and state ne 'Done' and state ne 'Failed'" \
| jq -r -e '.items[] | first' > scan.api.json
```

### Abort Scan in progress

```shell
cat <<EOF > scan-aborted.json
{
  "state": "Aborted"
}
EOF
```

```shell
jq -r -e '.id' scan.api.json > scan.id \
&& curl -sSf -X PATCH -H 'Content-Type: application/json' \
  "http://localhost:8888/api/scans/$(cat scan.id)" \
  -d @scan-aborted.json \
| jq -r -e '.' > scan-aborted.api.json
```

</details>

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -66,20 +66,24 @@ var rootCmd = &cobra.Command{
 	RunE: func(cmd *cobra.Command, args []string) error {
 		logger.Infof("Running...")
 
+		// Main context which remains active even if the scan is aborted allowing post-processing operations
+		// like updating scan result state
+		ctx := cmd.Context()
+
 		cli, err := newCli()
 		if err != nil {
 			return fmt.Errorf("failed to initialize CLI: %w", err)
 		}
 
-		ctx, cancel := context.WithCancel(cmd.Context())
+		// Create context used to signal to operations that the scan is aborted
+		abortCtx, cancel := context.WithCancel(ctx)
 		defer cancel()
 
 		// Start watching for abort event
 		cli.WatchForAbort(ctx, cancel, DefaultWatcherInterval)
 
-		// Collect non-critical errors for reporting them back to the backend
 		if waitForServerAttached {
-			if err := cli.WaitForVolumeAttachment(ctx); err != nil {
+			if err := cli.WaitForVolumeAttachment(abortCtx); err != nil {
 				err = fmt.Errorf("failed to wait for block device being attached: %w", err)
 				if e := cli.MarkDone(ctx, []error{err}); e != nil {
 					logger.Error(fmt.Errorf("failed to update scan result stat to completed with errors: %w", e))
@@ -89,7 +93,7 @@ var rootCmd = &cobra.Command{
 		}
 
 		if mountVolume {
-			mountPoints, err := cli.MountVolumes(ctx)
+			mountPoints, err := cli.MountVolumes(abortCtx)
 			if err != nil {
 				err = fmt.Errorf("failed to mount attached volume: %w", err)
 				logger.Error(err)
@@ -107,10 +111,10 @@ var rootCmd = &cobra.Command{
 		}
 
 		logger.Infof("Running scanners...")
-		res, familiesErr := families.New(logger, config).Run(ctx)
+		res, familiesErr := families.New(logger, config).Run(abortCtx)
 
 		logger.Infof("Exporting results...")
-		errs := cli.ExportResults(ctx, res, familiesErr)
+		errs := cli.ExportResults(abortCtx, res, familiesErr)
 
 		if len(familiesErr) > 0 {
 			errs = append(errs, fmt.Errorf("at least one family failed to run"))
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -96,7 +96,6 @@ var rootCmd = &cobra.Command{
 			mountPoints, err := cli.MountVolumes(abortCtx)
 			if err != nil {
 				err = fmt.Errorf("failed to mount attached volume: %w", err)
-				logger.Error(err)
 				if e := cli.MarkDone(ctx, []error{err}); e != nil {
 					logger.Error(fmt.Errorf("failed to update scan result stat to completed with errors: %w", e))
 				}
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/ghodss/yaml"
 	kubeclarityutils "github.com/openclarity/kubeclarity/shared/pkg/utils"
@@ -40,6 +41,8 @@ import (
 	"github.com/openclarity/vmclarity/shared/pkg/families/secrets"
 	"github.com/openclarity/vmclarity/shared/pkg/families/vulnerabilities"
 )
+
+const DefaultWatcherInterval = 2 * time.Minute
 
 var (
 	cfgFile string
@@ -63,24 +66,38 @@ var rootCmd = &cobra.Command{
 	RunE: func(cmd *cobra.Command, args []string) error {
 		logger.Infof("Running...")
 
-		ctx, cancel := context.WithCancel(cmd.Context())
-		defer cancel()
-
 		cli, err := newCli()
 		if err != nil {
 			return fmt.Errorf("failed to initialize CLI: %w", err)
 		}
 
+		ctx, cancel := context.WithCancel(cmd.Context())
+		defer cancel()
+
+		// Start watching for abort event
+		cli.WatchForAbort(ctx, cancel, DefaultWatcherInterval)
+
+		// Collect non-critical errors for reporting them back to the backend
 		if waitForServerAttached {
 			if err := cli.WaitForVolumeAttachment(ctx); err != nil {
-				return fmt.Errorf("failed to wait for block device being attached: %w", err)
+				err = fmt.Errorf("failed to wait for block device being attached: %w", err)
+				logger.Error(err)
+				if e := cli.MarkDone(ctx, []error{err}); e != nil {
+					logger.Error(fmt.Errorf("failed to update scan result stat to completed with errors: %w", e))
+				}
+				return err
 			}
 		}
 
 		if mountVolume {
 			mountPoints, err := cli.MountVolumes(ctx)
 			if err != nil {
-				return fmt.Errorf("failed to mount attached volume: %v", err)
+				err = fmt.Errorf("failed to mount attached volume: %w", err)
+				logger.Error(err)
+				if e := cli.MarkDone(ctx, []error{err}); e != nil {
+					logger.Error(fmt.Errorf("failed to update scan result stat to completed with errors: %w", e))
+				}
+				return err
 			}
 			setMountPointsForFamiliesInput(mountPoints, config)
 		}
@@ -90,17 +107,17 @@ var rootCmd = &cobra.Command{
 			return fmt.Errorf("failed to inform server %v scan has started: %w", server, err)
 		}
 
-		res, familiesErr := families.New(logger, config).Run()
+		logger.Infof("Running scanners...")
+		res, familiesErr := families.New(logger, config).Run(ctx)
 
 		logger.Infof("Exporting results...")
-		var generalErrors []error
-		exportErrors := cli.ExportResults(ctx, res, familiesErr)
-		generalErrors = append(generalErrors, exportErrors...)
+		errs := cli.ExportResults(ctx, res, familiesErr)
+
 		if len(familiesErr) > 0 {
-			generalErrors = append(generalErrors, fmt.Errorf("at least one family failed to run"))
+			errs = append(errs, fmt.Errorf("at least one family failed to run"))
 		}
 
-		err = cli.MarkDone(ctx, generalErrors)
+		err = cli.MarkDone(ctx, errs)
 		if err != nil {
 			return fmt.Errorf("failed to inform the server %v the scan was completed: %w", server, err)
 		}
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -81,7 +81,6 @@ var rootCmd = &cobra.Command{
 		if waitForServerAttached {
 			if err := cli.WaitForVolumeAttachment(ctx); err != nil {
 				err = fmt.Errorf("failed to wait for block device being attached: %w", err)
-				logger.Error(err)
 				if e := cli.MarkDone(ctx, []error{err}); e != nil {
 					logger.Error(fmt.Errorf("failed to update scan result stat to completed with errors: %w", e))
 				}
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -86,7 +86,7 @@ var rootCmd = &cobra.Command{
 			if err := cli.WaitForVolumeAttachment(abortCtx); err != nil {
 				err = fmt.Errorf("failed to wait for block device being attached: %w", err)
 				if e := cli.MarkDone(ctx, []error{err}); e != nil {
-					logger.Error(fmt.Errorf("failed to update scan result stat to completed with errors: %w", e))
+					logger.Errorf("Failed to update scan result stat to completed with errors: %v", e)
 				}
 				return err
 			}
@@ -97,7 +97,7 @@ var rootCmd = &cobra.Command{
 			if err != nil {
 				err = fmt.Errorf("failed to mount attached volume: %w", err)
 				if e := cli.MarkDone(ctx, []error{err}); e != nil {
-					logger.Error(fmt.Errorf("failed to update scan result stat to completed with errors: %w", e))
+					logger.Errorf("Failed to update scan result stat to completed with errors: %v", e)
 				}
 				return err
 			}
--- a/cli/pkg/cli/cli.go
+++ b/cli/pkg/cli/cli.go
@@ -130,7 +130,7 @@ func (c *CLI) ExportResults(ctx context.Context, res *results.Results, errs fami
 
 func (c *CLI) WatchForAbort(ctx context.Context, cancel context.CancelFunc, interval time.Duration) {
 	go func() {
-		timer := time.NewTimer(interval)
+		timer := time.NewTicker(interval)
 		defer timer.Stop()
 
 		for {
@@ -144,11 +144,7 @@ func (c *CLI) WatchForAbort(ctx context.Context, cancel context.CancelFunc, inte
 					cancel()
 					return
 				}
-				timer.Reset(interval)
 			case <-ctx.Done():
-				if !timer.Stop() {
-					<-timer.C
-				}
 				log.Debugf("Stop watching for abort event as context is cancelled")
 				return
 			}
--- a/cli/pkg/cli/cli.go
+++ b/cli/pkg/cli/cli.go
@@ -59,7 +59,7 @@ func (c *CLI) MountVolumes(ctx context.Context) ([]string, error) {
 			if err := device.Mount(mountDir); err != nil {
 				return nil, fmt.Errorf("failed to mount device: %v", err)
 			}
-			log.Infof("device %v on %v is mounted", device.DeviceName, mountDir)
+			log.Infof("Device %v on %v is mounted", device.DeviceName, mountDir)
 			mountPoints = append(mountPoints, mountDir)
 		}
 		if ctx.Err() != nil {
@@ -138,7 +138,7 @@ func (c *CLI) WatchForAbort(ctx context.Context, cancel context.CancelFunc, inte
 			case <-timer.C:
 				aborted, err := c.IsAborted(ctx)
 				if err != nil {
-					log.Errorf("failed to retrieve scan result state: %v", err)
+					log.Errorf("Failed to retrieve scan result state: %v", err)
 				}
 				if aborted {
 					cancel()
@@ -149,7 +149,7 @@ func (c *CLI) WatchForAbort(ctx context.Context, cancel context.CancelFunc, inte
 				if !timer.Stop() {
 					<-timer.C
 				}
-				log.Debugf("stop watching for abort evetn as context is cancelled")
+				log.Debugf("Stop watching for abort event as context is cancelled")
 				return
 			}
 		}
--- a/cli/pkg/cli/cli.go
+++ b/cli/pkg/cli/cli.go
@@ -19,6 +19,7 @@ package cli
 import (
 	"context"
 	"fmt"
+	"time"
 
 	uuid "github.com/satori/go.uuid"
 	log "github.com/sirupsen/logrus"
@@ -125,6 +126,34 @@ func (c *CLI) ExportResults(ctx context.Context, res *results.Results, errs fami
 	}
 
 	return result
+}
+
+func (c *CLI) WatchForAbort(ctx context.Context, cancel context.CancelFunc, interval time.Duration) {
+	go func() {
+		timer := time.NewTimer(interval)
+		defer timer.Stop()
+
+		for {
+			select {
+			case <-timer.C:
+				aborted, err := c.IsAborted(ctx)
+				if err != nil {
+					log.Errorf("failed to retrieve scan result state: %v", err)
+				}
+				if aborted {
+					cancel()
+					return
+				}
+				timer.Reset(interval)
+			case <-ctx.Done():
+				if !timer.Stop() {
+					<-timer.C
+				}
+				log.Debugf("stop watching for abort evetn as context is cancelled")
+				return
+			}
+		}
+	}()
 }
 
 func isSupportedFS(fs string) bool {
--- a/cli/pkg/state/local.go
+++ b/cli/pkg/state/local.go
@@ -35,6 +35,7 @@ func (l *LocalState) MarkInProgress(context.Context) error {
 func (l *LocalState) MarkDone(_ context.Context, errs []error) error {
 	if len(errs) > 0 {
 		log.Errorf("scan has been completed with errors: %v", errs)
+		return nil
 	}
 	log.Info("scan has been completed")
 	return nil
--- a/cli/pkg/state/local.go
+++ b/cli/pkg/state/local.go
@@ -32,8 +32,11 @@ func (l *LocalState) MarkInProgress(context.Context) error {
 	return nil
 }
 
-func (l *LocalState) MarkDone(context.Context, []error) error {
-	log.Info("scanning is finished")
+func (l *LocalState) MarkDone(_ context.Context, errs []error) error {
+	if len(errs) > 0 {
+		log.Errorf("scan has been completed with errors: %v", errs)
+	}
+	log.Info("scan has been completed")
 	return nil
 }
 
--- a/cli/pkg/state/local.go
+++ b/cli/pkg/state/local.go
@@ -37,6 +37,10 @@ func (l *LocalState) MarkDone(context.Context, []error) error {
 	return nil
 }
 
+func (l *LocalState) IsAborted(context.Context) (bool, error) {
+	return false, nil
+}
+
 func NewLocalState() (*LocalState, error) {
 	return &LocalState{}, nil
 }
--- a/cli/pkg/state/local.go
+++ b/cli/pkg/state/local.go
@@ -28,7 +28,7 @@ func (l *LocalState) WaitForVolumeAttachment(context.Context) error {
 }
 
 func (l *LocalState) MarkInProgress(context.Context) error {
-	log.Info("scanning is in progress")
+	log.Info("Scanning is in progress")
 	return nil
 }
 
@@ -37,7 +37,7 @@ func (l *LocalState) MarkDone(_ context.Context, errs []error) error {
 		log.Errorf("scan has been completed with errors: %v", errs)
 		return nil
 	}
-	log.Info("scan has been completed")
+	log.Info("Scan has been completed")
 	return nil
 }
 
--- a/cli/pkg/state/vmclarity.go
+++ b/cli/pkg/state/vmclarity.go
@@ -137,6 +137,26 @@ func (v *VMClarityState) MarkDone(ctx context.Context, errors []error) error {
 	return nil
 }
 
+func (v VMClarityState) IsAborted(ctx context.Context) (bool, error) {
+	scanResult, err := v.client.GetScanResult(ctx, v.scanResultID, models.GetScanResultsScanResultIDParams{
+		Select: utils.PointerTo("id,status"),
+	})
+	if err != nil {
+		return false, fmt.Errorf("failed to get scan result: %w", err)
+	}
+
+	state, ok := scanResult.GetGeneralState()
+	if !ok {
+		return false, errors.New("failed to get general state of scan result")
+	}
+
+	if state == models.ABORTED {
+		return true, nil
+	}
+
+	return false, nil
+}
+
 func NewVMClarityState(client *backendclient.BackendClient, id ScanResultID) (*VMClarityState, error) {
 	if client == nil {
 		return nil, errors.New("backend client must not be nil")
--- a/runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
+++ b/runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
@@ -105,7 +105,7 @@ func (w *Watcher) GetAbortedScans(ctx context.Context) ([]ScanReconcileEvent, er
 }
 
 func (w *Watcher) Reconcile(ctx context.Context, event ScanReconcileEvent) error {
-	w.logger.Infof("reconciling scan event: %v", event)
+	w.logger.Infof("Reconciling scan event: %v", event)
 
 	selector := "id,state,stateReason"
 	params := models.GetScansScanIDParams{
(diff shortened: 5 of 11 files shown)


cli/cmd/root.go
```diff
@@ -63,24 +66,38 @@ var rootCmd = &cobra.Command{
...
+
+		// Start watching for abort event
+		cli.WatchForAbort(ctx, cancel, DefaultWatcherInterval)
+
+		// Collect non-critical errors for reporting them back to the backend
 		if waitForServerAttached {
 			if err := cli.WaitForVolumeAttachment(ctx); err != nil {
-				return fmt.Errorf("failed to wait for block device being attached: %w", err)
+				err = fmt.Errorf("failed to wait for block device being attached: %w", err)
+				logger.Error(err)
```
sambetts-cisco, line 84: If we update the local state manager to log the errors passed to MarkDone, then we don't need to remember to log here.

shared/pkg/families/manager.go
```diff
@@ -77,17 +78,43 @@ func New(logger *log.Entry, config *Config) *Manager {
...
+				log.Debugf("received result from family %q: %v", family, r)
+				if r.err != nil {
+					familyErrors[family.GetType()] = fmt.Errorf("failed to run family %v: %w", family.GetType(), r.err)
+				} else {
+					familyResults.SetResults(r.result)
+				}
+				break inner
+			default:
+				continue
+			}
```
sambetts-cisco, line 117: we don't want to send the context all the way down to the family/job manager? then each family would return failure

I guess because that logic is in kubeclarity we'd need to make a change there first to add context ?

shared/pkg/families/manager.go
```diff
@@ -77,17 +78,43 @@ func New(logger *log.Entry, config *Config) *Manager {
...
 		}
```
sambetts-cisco, line 118: I don't think we need a for loop here, otherwise we're busy waiting instead of blocking on the channels. We also won't need the go labels either.

shared/pkg/families/manager.go
```diff
@@ -77,17 +78,43 @@ func New(logger *log.Entry, config *Config) *Manager {
...
+				ret,
+				err,
+			}
+		}()
+
+	inner:
+		for {
+			select {
+			case <-ctx.Done():
+				log.Warningf("received context cancelled while family %q is still running. Aborting...", family)
```
sambetts-cisco, line 105: I think if a particular family is aborted while its still running it should be counted as a family error e.g.

```
familyErrors[family.GetType()] = fmt.Errorf("failed to run family %v: aborted", family.GetType(), r.err)
```

sambetts-cisco (changes requested):

runtime_scan/pkg/scanner/job_managment.go
```diff
@@ -225,32 +209,89 @@ func (s *Scanner) worker(ctx context.Context, queue chan *scanData, workNumber i
...
+				s.Lock()
+				data.success = !scanStatusHasErrors(scanResultStatus)
+				data.completed = true
+				s.Unlock()
+				return
+			}
+		case <-ctx.Done():
+			if !timer.Stop() {
+				<-timer.C
+			}
```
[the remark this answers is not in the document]
chrisgacsal: Totally agree!

shared/pkg/families/manager.go
chrisgacsal: That's correct. I have started the work in KubeClarity, but it will take a decent effort to sort this out there. I wanted to have a short term solution here in order to get the initial version of this functionality done. Otherwise, I'll keep refactoring here and there while the aborting functionality moving nowhere. :)

cli/cmd/root.go
chrisgacsal: Fixed.

shared/pkg/families/manager.go
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
```diff
@@ -0,0 +1,220 @@
...
+		fallthrough
+	default:
+	}
+
+	return nil
+}
+
+func (w *Watcher) getScansByState(ctx context.Context, s models.ScanState) (models.Scans, error) {
+	filter := fmt.Sprintf("state eq '%s'", s)
+	selector := "id,state,stateReason,targetIDs"
```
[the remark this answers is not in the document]
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
```diff
@@ -0,0 +1,220 @@
...
+		scans = &models.Scans{}
+	}
+
+	return *scans, err
+}
+
+func (w *Watcher) reconcileAborted(ctx context.Context, event ScanReconcileEvent) error {
+	filter := fmt.Sprintf("scan/id eq '%s' and status/general/state ne '%s' and status/general/state ne '%s'",
+		event.ScanID, models.ABORTED, models.DONE)
+	selector := "id,scan,status,target"
```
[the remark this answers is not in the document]
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
```diff
@@ -0,0 +1,220 @@
...
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			sr := models.TargetScanResult{
+				Status: &models.TargetScanStatus{
+					General: &models.TargetScanState{
+						State: runtimeScanUtils.PointerTo(models.ABORTED),
+					},
+				},
+			}
```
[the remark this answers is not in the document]
chrisgacsal: Uhh. I don't think this is the right behavior for PATCHing resources. What if the scanner job updates other part of the Status field of the ScanResult *after* we retrieved it from the API and *before* sending the PATCH request? Would that information lost? 

Anyway, I'll update the PR to send the whole content of the Status field back with our updates, but this does not feel right.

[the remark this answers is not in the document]
sambetts-cisco: I think I misread the spec https://www.rfc-editor.org/rfc/rfc7396. The code *should* do recursive merges as long as the target of the merge is a json object:

> If the patch is anything other than an object, the result will always be to replace the entire target with the entire patch.

Regarding:
 
> What if the scanner job updates other part of the Status field of the ScanResult after we retrieved it from the API and before sending the PATCH request? Would that information lost?

This is why we need to add a resource version / etag on to our objects. Any update like this should read the etag when it GET's the object, and the include the etag as a header in the PATCH request to ensure that we don't end up with lost updates.
In scenario's where the object has change on the server the etag should cause the request to get rejected, and then this logic should retry to ensure that it making the correct change according the current state of the object.

shared/pkg/families/manager.go
chrisgacsal: Keep forgetting that `select` is blocking without the `default` case. Fixed.

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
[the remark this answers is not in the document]
chrisgacsal: So in our case patching only the General State should be enough. Does the backend support the RFC from above?

chrisgacsal:
The PR is ready for review. I have added a test plan with manual steps to verify the functionality to abort running scans.

shared/pkg/families/manager.go
```diff
@@ -77,17 +78,35 @@ func New(logger *log.Entry, config *Config) *Manager {
...
-	familyErrors := make(map[types.FamilyType]error)
+type familyResult struct {
+	result interfaces.IsResults
+	err    error
+}
 
+func (m *Manager) Run(ctx context.Context) (*results.Results, RunErrors) {
+	familyErrors := make(RunErrors)
 	familyResults := results.New()
+	result := make(chan familyResult, len(m.families))
```
sambetts-cisco, line 89: I'm not sure but from a memory management perspective it might be better to move this channel into the loop, so that we can clean it up when we're done for example:
```
for _, family := range m.families {
    result := make(chan familyResult)
    go func() {
        ret, err := family.Run(familyResults)
        result <- familyResult{
           ret,
           err,
        }
    }()
    
    select {
       case <-ctx.Done():
          // drain and close result channel
          go func() {
              <-result
              close(result)
          }()
          // handle aborted error
        case <-result:
           // handle result
           close(result)
    }
}
```

sambetts-cisco (approved):
Overall looking great, I've got a couple of suggestions, happy to punt them to later.

cli/pkg/state/local.go
```diff
@@ -32,11 +32,18 @@ func (l *LocalState) MarkInProgress(context.Context) error {
 	return nil
 }
 
-func (l *LocalState) MarkDone(context.Context, []error) error {
-	log.Info("scanning is finished")
+func (l *LocalState) MarkDone(_ context.Context, errs []error) error {
+	if len(errs) > 0 {
+		log.Errorf("scan has been completed with errors: %v", errs)
+	}
+	log.Info("scan has been completed")
```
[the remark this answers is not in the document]
chrisgacsal: Actually there is a missing `return nil`. Will fix it in this PR.

cli/cmd/root.go
```diff
@@ -63,24 +66,41 @@ var rootCmd = &cobra.Command{
...
 			}
 		}
 
 		if mountVolume {
-			mountPoints, err := cli.MountVolumes(ctx)
+			mountPoints, err := cli.MountVolumes(abortCtx)
 			if err != nil {
-				return fmt.Errorf("failed to mount attached volume: %v", err)
+				err = fmt.Errorf("failed to mount attached volume: %w", err)
+				logger.Error(err)
```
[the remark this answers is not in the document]
chrisgacsal: You are right, there is not need for logging here. Will fix it in this PR.

shared/pkg/families/manager.go
```diff
@@ -77,17 +78,35 @@ func New(logger *log.Entry, config *Config) *Manager {
...
-			familyErrors[family.GetType()] = fmt.Errorf("failed to run family %v: %w", family.GetType(), err)
-		} else {
-			familyResults.SetResults(ret)
+		go func() {
+			ret, err := family.Run(familyResults)
+			result <- familyResult{
+				ret,
+				err,
+			}
+		}()
```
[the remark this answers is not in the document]
chrisgacsal: Alright. My plan is to add support for `context.Context` to families.Manager in a separate PR which I have almost finished.

shared/pkg/families/manager.go
chrisgacsal: So the reason why `result` channel is created outside the for loop is because I think that certain families can be grouped together and run them parallel.
But I will add suggested solution and will deal with making families run parallel later.

chrisgacsal: Fixed.

shared/pkg/families/manager.go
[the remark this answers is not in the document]
chrisgacsal: Fixed.

cli/cmd/root.go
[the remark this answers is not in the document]
chrisgacsal: Fixed.

cli/pkg/state/local.go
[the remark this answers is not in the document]
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
```diff
@@ -0,0 +1,220 @@
...
+		Select: &selector,
+	}
+
+	scanResults, err := w.client.GetScanResults(ctx, params)
+	if err != nil {
+		return fmt.Errorf("getting ScanResult(s) for Scan with %s id failed: %v", event.ScanID, err)
+	}
+
+	if scanResults.Items != nil && len(*scanResults.Items) > 0 {
+		var retryIsNeeded bool
```
akpsgit, line 171: Not sure how retry is reflected here or to the user.

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
```diff
@@ -0,0 +1,220 @@
...
+		r[i] = ScanReconcileEvent{
+			ScanID: *scan.Id,
+		}
+	}
+
+	return r, nil
+}
+
+func (w *Watcher) Reconcile(ctx context.Context, event ScanReconcileEvent) error {
+	w.logger.Infof("reconciling scan event: %v", event)
```
akpsgit, line 108: Reconciling

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	state, ok := scan.GetState()
+	if !ok {
+		return fmt.Errorf("cannot determine state of Scan with %s id", event.ScanID)
+	}
+
+	switch state {
+	case models.ScanStateDone, models.ScanStateFailed:
+		w.logger.Infof("reconciling scan event is skipped as Scan is already finished: %v", event)
```
akpsgit, line 127: Reconciling

cli/pkg/cli/cli.go
```diff
@@ -127,6 +128,34 @@ func (c *CLI) ExportResults(ctx context.Context, res *results.Results, errs fami
...
+	go func() {
+		timer := time.NewTimer(interval)
+		defer timer.Stop()
+
+		for {
+			select {
+			case <-timer.C:
+				aborted, err := c.IsAborted(ctx)
+				if err != nil {
+					log.Errorf("failed to retrieve scan result state: %v", err)
```
akpsgit, line 141: Failed

cli/pkg/cli/cli.go
```diff
@@ -127,6 +128,34 @@ func (c *CLI) ExportResults(ctx context.Context, res *results.Results, errs fami
...
+				if aborted {
+					cancel()
+					return
+				}
+				timer.Reset(interval)
+			case <-ctx.Done():
+				if !timer.Stop() {
+					<-timer.C
+				}
+				log.Debugf("stop watching for abort evetn as context is cancelled")
```
akpsgit, line 152: Stop

cli/pkg/state/local.go
```diff
@@ -32,11 +32,19 @@ func (l *LocalState) MarkInProgress(context.Context) error {
 	return nil
 }
 
-func (l *LocalState) MarkDone(context.Context, []error) error {
-	log.Info("scanning is finished")
+func (l *LocalState) MarkDone(_ context.Context, errs []error) error {
+	if len(errs) > 0 {
+		log.Errorf("scan has been completed with errors: %v", errs)
```
akpsgit, line 37: Scan

cli/pkg/state/local.go
```diff
@@ -32,11 +32,19 @@ func (l *LocalState) MarkInProgress(context.Context) error {
...
+		return nil
+	}
+	log.Info("scan has been completed")
```
akpsgit, line 40: Scan

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
```diff
@@ -0,0 +1,220 @@
...
+					Status: &models.TargetScanStatus{
+						General: &models.TargetScanState{
+							State: runtimeScanUtils.PointerTo(models.ABORTED),
+						},
+					},
+				}
+
+				err := w.client.PatchScanResult(ctx, sr, id)
+				if err != nil {
+					w.logger.Errorf("failed to patch ScanResult with id: %s", id)
```
akpsgit, line 193: Failed

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
```diff
@@ -0,0 +1,220 @@
...
+		Filter: &filter,
+		Select: &selector,
+	}
+
+	scanResults, err := w.client.GetScanResults(ctx, params)
+	if err != nil {
+		return fmt.Errorf("getting ScanResult(s) for Scan with %s id failed: %v", event.ScanID, err)
+	}
+
+	if scanResults.Items != nil && len(*scanResults.Items) > 0 {
```
akpsgit, line 170: maybe better to return if scanResults.Items == nil || len(*scanResults.Items) == 0

cli/cmd/root.go
```diff
@@ -63,24 +66,40 @@ var rootCmd = &cobra.Command{
...
+		// Start watching for abort event
+		cli.WatchForAbort(ctx, cancel, DefaultWatcherInterval)
+
 		if waitForServerAttached {
-			if err := cli.WaitForVolumeAttachment(ctx); err != nil {
-				return fmt.Errorf("failed to wait for block device being attached: %w", err)
+			if err := cli.WaitForVolumeAttachment(abortCtx); err != nil {
+				err = fmt.Errorf("failed to wait for block device being attached: %w", err)
+				if e := cli.MarkDone(ctx, []error{err}); e != nil {
+					logger.Error(fmt.Errorf("failed to update scan result stat to completed with errors: %w", e))
```
akpsgit, line 89: Failed

cli/cmd/root.go
```diff
@@ -63,24 +66,40 @@ var rootCmd = &cobra.Command{
...
 		}
 
 		if mountVolume {
-			mountPoints, err := cli.MountVolumes(ctx)
+			mountPoints, err := cli.MountVolumes(abortCtx)
 			if err != nil {
-				return fmt.Errorf("failed to mount attached volume: %v", err)
+				err = fmt.Errorf("failed to mount attached volume: %w", err)
+				if e := cli.MarkDone(ctx, []error{err}); e != nil {
+					logger.Error(fmt.Errorf("failed to update scan result stat to completed with errors: %w", e))
```
akpsgit, line 100: Failed

runtime_scan/pkg/scanner/job_managment.go
```diff
@@ -114,19 +115,41 @@ func (s *Scanner) jobBatchManagement(ctx context.Context) {
...
+					scan.StateReason = utils.PointerTo(models.ScanStateReasonUnexpected)
+					break
+				}
+
+				var isAborted bool
+				if scanState == models.ScanStateAborted {
+					isAborted = true
+				}
+
+				if isAborted {
```
akpsgit, line 133: maybe just if scanState == models.ScanStateAborted?

cli/pkg/cli/cli.go
FrimIdan: `evetn`?

cli/pkg/cli/cli.go
```diff
@@ -127,6 +128,34 @@ func (c *CLI) ExportResults(ctx context.Context, res *results.Results, errs fami
...
+				}
+				if aborted {
+					cancel()
+					return
+				}
+				timer.Reset(interval)
+			case <-ctx.Done():
+				if !timer.Stop() {
+					<-timer.C
+				}
```
FrimIdan, line 151: can you explain why there is a need to be blocked on timer here?

cli/pkg/state/vmclarity.go
```diff
@@ -137,6 +137,26 @@ func (v *VMClarityState) MarkDone(ctx context.Context, errors []error) error {
...
+	scanResult, err := v.client.GetScanResult(ctx, v.scanResultID, models.GetScanResultsScanResultIDParams{
+		Select: utils.PointerTo("id,status"),
+	})
+	if err != nil {
+		return false, fmt.Errorf("failed to get scan result: %w", err)
+	}
+
+	state, ok := scanResult.GetGeneralState()
+	if !ok {
+		return false, errors.New("failed to get general state of scan result")
```
FrimIdan, line 150: please be consistent with `fmt.Errorf`/`errors.New`

cli/pkg/state/vmclarity.go
```diff
@@ -137,6 +137,26 @@ func (v *VMClarityState) MarkDone(ctx context.Context, errors []error) error {
 	return nil
 }
 
+func (v VMClarityState) IsAborted(ctx context.Context) (bool, error) {
+	scanResult, err := v.client.GetScanResult(ctx, v.scanResultID, models.GetScanResultsScanResultIDParams{
+		Select: utils.PointerTo("id,status"),
```
FrimIdan, line 142: why we are not doing filter on ABORTED state with the odata?

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
```diff
@@ -0,0 +1,220 @@
...
+		Queue:             queue,
+		ReconcileFunction: w.Reconcile,
+	}
+	reconciler.Start(ctx)
+}
+
+func (w *Watcher) GetAbortedScans(ctx context.Context) ([]ScanReconcileEvent, error) {
+	scans, err := w.getScansByState(ctx, models.ScanStateAborted)
+	if err != nil || scans.Items == nil || len(*scans.Items) <= 0 {
+		return nil, err
```
FrimIdan, line 93: please wrap the error, in case `err==nil`

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
FrimIdan: we want to print that info message every time?

cli/pkg/state/vmclarity.go
chrisgacsal: @FrimIdan You mean starting error messages with capital letter?

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
chrisgacsal: @akpsgit The `retryIsNeeded` is used here to inform the the `common.Reconciler` that the reconciliation is failed.
So if your suggestion is that this needs to be a silent error (e.g. without logging/returning error) and let the `common.Poller` pick it up the aborted scan again in case the `retryIsNeeded == true` than it is fine by me.

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
chrisgacsal: I'll set it to debug level, so it won't be log during normal operation.

cli/pkg/state/vmclarity.go
FrimIdan: I mean the use of `errors.New` and `fmt.Errorf`

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
chrisgacsal: Nope. We still need to update the Scan [here](https://github.com/openclarity/vmclarity/pull/218/files#diff-ffbf779aa4010b4a63a4c4576e2765b61ba56f7db4d4249002b9e27f037ab4e6R206-R217) which is temporarily disabled as it conflicts with the code in `Scanner.jobBatchManagement`. I've started to refactor the whole Scan lifecycle management by adopting the reconciler logic and moving Scan related operations to `ScanWatcher`.

So I'd keep it as is.

runtime_scan/pkg/scanner/job_managment.go
chrisgacsal: Fixed.

cli/pkg/cli/cli.go
chrisgacsal: @FrimIdan https://pkg.go.dev/time#Timer.Stop

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
chrisgacsal: Could you please elaborate on this? If `err != nil` returned by `getScansByState()` than `GetAbortedScans()` returnes error as well, otherwise not.

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
chrisgacsal: Fixed.

cli/pkg/cli/cli.go
chrisgacsal: Fixed.

cli/pkg/cli/cli.go
chrisgacsal: Fixed.

cli/pkg/state/local.go
chrisgacsal: Fixed.

cli/pkg/state/local.go
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
chrisgacsal: Fixed.

cli/cmd/root.go
chrisgacsal: Fixed.

cli/cmd/root.go
chrisgacsal: Fixed.

cli/pkg/state/vmclarity.go
chrisgacsal: The `models.GetScanResultsScanResultIDParams` does not support setting `filter` field for `GetScanResult`.

Using `GetScanResults` would allow filtering based on status like this: `/api/scanResults?$filter=id eq 'SCAN_RESULT_ID' and status/general/state eq 'ABORTED'`. However  the API returns and empty list if there is no `ScanResult`which matches the filter. The problem with this approach is that empty list could mean two things: either the `ScanResult` does not exist at all or the `ScanResult` does exist but has different state than ABORT. We need to be able to differentiate between the two cases as former must raise an error while the later means the `ScanResult` in question is in `ABORTED` state.

So I think the current implementation is the right way of checking whether the `ScanResult` is aborted or not.

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
akpsgit: just when  (retryIsNeeded == true) nothing that is related to retry happening (e.g. to do a retry, notify in the error text that a retry is needed (to be done by the user?), etc.). 
or if we don't want to handle the retry now, maybe we can change the retryIsNeeded var to patchErr and concatenate it to the returned error. WDYT?

cli/cmd/root.go
```diff
@@ -86,7 +86,8 @@ var rootCmd = &cobra.Command{
 			if err := cli.WaitForVolumeAttachment(abortCtx); err != nil {
 				err = fmt.Errorf("failed to wait for block device being attached: %w", err)
 				if e := cli.MarkDone(ctx, []error{err}); e != nil {
-					logger.Error(fmt.Errorf("failed to update scan result stat to completed with errors: %w", e))
+					//nolint:stylecheck
+					logger.Error(fmt.Errorf("Failed to update scan result stat to completed with errors: %w", e))
```
FrimIdan, line 90: why you need the `//nolint:stylecheck`?
why not just 
```
logger.Errorf("Failed to update scan result stat to completed with errors: %v", err)
```

you don't need to wrap the error since it is not returned, right?

cli/pkg/state/vmclarity.go
FrimIdan: > The models.GetScanResultsScanResultIDParams does not support setting filter field for GetScanResult.

I think that this is just a missing implementation, why wouldn't we want to support it?

shared/pkg/families/manager.go
```diff
@@ -77,17 +78,40 @@ func New(logger *log.Entry, config *Config) *Manager {
...
-		} else {
-			familyResults.SetResults(ret)
+		result := make(chan familyResult)
+		go func() {
+			ret, err := family.Run(familyResults)
+			result <- familyResult{
+				ret,
+				err,
+			}
+		}()
```
FrimIdan, line 98: @chrisgacsal I'm not sure how the family run in parallel related to the current PR, but there is a problem here. There are families that depends on results from other families.
For example, the vulnerability family sometimes expect to get   `input_from_sbom`, How the parallel scan handle this?
Did you tried to run sbom & vulnerability together, the code is setting `familiesConfig.Vulnerabilities.InputFromSbom = true` in this case and I would expect to get a failure in the vulnerability family run.

cli/pkg/state/vmclarity.go
FrimIdan: I guess we can close it for now.

cli/cmd/root.go
chrisgacsal: Will fix.

FrimIdan:
LGTM

runtime_scan/pkg/orchestrator/scanwatcher/watcher.go
chrisgacsal: So the confusion around `retryIsNeeded` was that the name of the variable which suggested that there is a retry logic which is triggered based on the value of this variable. It has been renamed to `reconciliationFailed` which provides a better explanation for the use of this variable.

runtime_scan/pkg/scanner/job_managment.go
```diff
@@ -225,32 +227,83 @@ func (s *Scanner) worker(ctx context.Context, queue chan *scanData, workNumber i
...
+
+	switch state {
+	case models.INIT:
+		job, err = s.runJob(ctx, data)
+		if err != nil {
+			s.Lock()
+			data.success = false
+			data.completed = true
+			s.Unlock()
+			return nil, fmt.Errorf("failed to run scan job for target %s: %v", data.targetInstance, err)
```
akpsgit, line 251: data.targetInstance.TargetID

runtime_scan/pkg/scanner/job_managment.go
```diff
@@ -225,32 +227,83 @@ func (s *Scanner) worker(ctx context.Context, queue chan *scanData, workNumber i
...
+		}
+		fallthrough
+	case models.ATTACHED, models.INPROGRESS, models.ABORTED:
+		s.waitForResult(ctx, data, ks)
+		if data.timeout {
+			return nil, fmt.Errorf("scan job for target %s timed out: %v", data.targetInstance, err)
```
akpsgit, line 257: data.targetInstance.TargetID

runtime_scan/pkg/scanner/job_managment.go
chrisgacsal: Fixed.

runtime_scan/pkg/scanner/job_managment.go
chrisgacsal: Fixed.

akpsgit (approved):

PR #273, opened 2023-04-25
Add ability to configure timeout for Scans

sambetts-cisco:
## Description

Through the scan configuration a maximum number of seconds (`timeoutSeconds`) can be configured.

When a scan is triggered from a configuration this number of seconds will be added to the start time and saved as `timeoutTime` on the Scan object.

If while the Scan is running the current time exceeds the `timeoutTime` then the scan is automatically failed and any remaining AssetScans will be terminated.

If no `timeoutTime` is provided on a scan it will be automatically timed out 48 hours after the start time of the scan.

## Type of Change

[ ] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


api/openapi.yaml
```diff
@@ -1056,10 +1056,18 @@ components:
...
           format: date-time
+        timeoutTime:
+          description: |
+            A time in the future when this Scan will be automatically aborted
+            if it is still in progress.
+          type: string
+          format: date-time
         scanConfig:
           $ref: '#/components/schemas/ScanConfigRelationship'
+          readOnly: true
```
sambetts-cisco, line 1067: This shouldn't be readOnly otherwise you can't set the `scanConfig: {id: ...}` field. The ScanConfigRelationship enforces that all fields except ID are readOnly.

api/openapi.yaml
```diff
@@ -1056,10 +1056,18 @@ components:
...
         scanConfigSnapshot:
           $ref: '#/components/schemas/ScanConfigSnapshot'
+          readOnly: true
```
sambetts-cisco, line 1070: This shouldn't be readOnly otherwise the Orchestrator can't set the field.

api/openapi.yaml
```diff
@@ -1056,10 +1056,18 @@ components:
         endTime:
           type: string
           format: date-time
+        timeoutTime:
+          description: |
+            A time in the future when this Scan will be automatically aborted
+            if it is still in progress.
+          type: string
+          format: date-time
```
sambetts-cisco, line 1064: I'm torn whether to make this a timeoutTime or timeoutSeconds. From a API UX perspective timeoutSeconds is better because you don't need to calculate a time before posting it to the API, and then we can compare those seconds to the startTime to know if we've reached timeout.

I think timeoutSeconds will be better when we have to ability to POST a Scan directly without a ScanConfig.

sambetts-cisco (commented):
I think we should drop the TimeoutTime logic now, its not great from an API UX perspective and StartTime + TimeoutSeconds satisfies the same flow in the code.

chrisgacsal:
> I think we should drop the TimeoutTime logic now, its not great from an API UX perspective and StartTime + TimeoutSeconds satisfies the same flow in the code.

Agreed. It felt weird to me as well. I thought that having the `TimeoutTime` could be helpful to have this field for API consumers so they do not need to calculate the actual timeout on their own. But doing the math on the client side should not be an issue.

api/openapi.yaml
chrisgacsal: My bad, reverting this.

api/openapi.yaml
chrisgacsal: I'll remove it completely and keep the `timeoutSeconds` as part of the `ScanConfigSnapshot`. `Scan` objects must still have the `ScanConfigSnapshot` attribute even if the there is no parent `ScanConfig` for them.

sambetts-cisco: 👍 sounds good. The re-model should make this a little less weird

chrisgacsal: Fixed.

api/openapi.yaml
```diff
@@ -1237,6 +1237,11 @@ components:
           $ref: '#/components/schemas/ScanFamiliesConfig'
         scope:
           $ref: '#/components/schemas/ScanScopeType'
+        timeoutSeconds:
```
akpsgit, line 1240: need to update the odata schema for the api changes

chrisgacsal: Nice catch, keep forgetting to do that. Will fix it now.

runtime_scan/pkg/orchestrator/scanwatcher/helpers.go
```diff
@@ -28,9 +28,17 @@ func isScanTimedOut(scan *models.Scan, timeout time.Duration) bool {
 	if scan == nil || scan.StartTime == nil {
 		return false
 	}
-	deadline := (*scan.StartTime).Add(timeout)
 
-	return time.Now().UTC().After(deadline)
+	now := time.Now().UTC()
+
+	// Set timeoutTime to global timeout provided to the ScanWatcher controller
+	timeoutTime := scan.StartTime.Add(timeout)
```
akpsgit, line 35: maybe it's better to do it in an else of the if, also rename the param timeout->globalTimeout following your comment?

chrisgacsal: Honestly, I'd keep it as is since it is easy to understand even without the comments. I prefer to not use if-else until it is absolutely necessary.

api/openapi.yaml
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/scanconfigwatcher/helpers.go
```diff
@@ -31,6 +31,7 @@ func newScanFromScanConfig(scanConfig *models.ScanConfig) *models.Scan {
 			ScanFamiliesConfig:  scanConfig.ScanFamiliesConfig,
 			Scheduled:           scanConfig.Scheduled,
 			Scope:               scanConfig.Scope,
+			TimeoutSeconds:      scanConfig.TimeoutSeconds,
```
akpsgit, line 34: maybe worth checking that this doesn't come negative from the api because you're doing calculations on it, or maybe doing it with min value in the API (not sure what it should be...)

chrisgacsal: I'll check how we can add validation to this on the API side.

chrisgacsal: @akpsgit I added validation to `isScanTimedOut` helper function so we only use `TimeoutSeconds` with value bigger than zero.

runtime_scan/pkg/orchestrator/scanwatcher/helpers.go
```diff
@@ -28,9 +28,17 @@ func isScanTimedOut(scan *models.Scan, timeout time.Duration) bool {
 	if scan == nil || scan.StartTime == nil {
 		return false
 	}
-	deadline := (*scan.StartTime).Add(timeout)
 
-	return time.Now().UTC().After(deadline)
+	now := time.Now().UTC()
+	// Use the provided timeout to calculate the timeoutTime by default.
+	timeoutTime := scan.StartTime.Add(timeout)
```
sambetts-cisco, line 34: I think it would be more readable to switch this to use the DefaultScanTimeout const here instead of passing  timeout into isScanTimedOut. Its just one more level of indirection which makes this hard to understand. WDYT?

chrisgacsal: There are two reasons why we pass the timeout parameter here:
* the default timeout parameter is comes from the `scanwatcher.Config` which can be set by `SCAN_TIMEOUT` which has the `DefaultScanTimeout` as a default.
* the `isScanTimedOut` is a helper function which is not method of the `ScanWatcher` and it should not be

I am leaning towards make the `isScanTimedOut` as a method for `models.Scan` with the same semantics.

sambetts-cisco: hmm, I see, I didn't realise the default was configurable through the scanwatcher.Config. I agree moving it as a helper function on the Scan makes sense. Also I think we need to change the variable naming a bit to make it obvious that its the default timeout being passed into the function:

```
func (s *Scan) IsTimedOut(defaultTimeout time.Duration) {
   timeoutDuration := defaultTimeout
   if timeoutSeconds := s.GetTimeoutSeconds(); timeout > 0 {
       timeoutDuration = time.Duration(timeoutSeconds) * time.Second
   }
   
   endTime := s.StartTime.Add(timeoutDuration)
   return time.Now().After(endTime)
}
```

sambetts-cisco (commented):
LGTM, thanks @chrisgacsal

pbalogh-sa (approved):

PR #289, opened 2023-04-30
extend target info

fishkerez:
## Description

Extend target info with the following properties:
Tags 
Image
instance type (micro, large, etc.)
Platform (linux distro)
add ../vpc/security-group to the location which is only the region currently
Launch time

## Type of Change

[* ] New Feature


ui_backend/pkg/rest/controllers.go
```diff
@@ -79,15 +80,15 @@ func createRegionFindingsFromTargets(targets *backendmodels.Targets) []models.Re
...
+func getTargetRegion(target backendmodels.Target) (string, error) {
 	discriminator, err := target.TargetInfo.ValueByDiscriminator()
 	if err != nil {
 		return "", fmt.Errorf("failed to get value by discriminator: %w", err)
 	}
 
 	switch info := discriminator.(type) {
 	case backendmodels.VMInfo:
-		return info.Location, nil
+		return strings.Split(info.Location, "/")[0], nil
```
FrimIdan, line 91: This is assuming you know how the location was created, and true only for AWS, is this safe/right to get the region?
Maybe better to also save the region in a separate value on the target if it is needed? Also, region is for AWS, what about other cloud-providers? I think we call it  `getTargetLocation` to be generic.

runtime_scan/pkg/provider/aws/instance.go
```diff
@@ -32,21 +32,52 @@ type InstanceImpl struct {
...
 func (i *InstanceImpl) GetID() string {
 	return i.id
 }
 
 func (i *InstanceImpl) GetLocation() string {
-	return i.region
+	ret := i.region + "/" + i.vpcID
+	for _, group := range i.securityGroups { // TODO how do we want to represent multiple sg?
+		ret = ret + "/" + group
+	}
```
FrimIdan, line 53: maybe we should not put sg as part of the location, after all if instance can have multiple sg it doesn't say it exists in multiple locations.

runtime_scan/pkg/provider/aws/client.go
```diff
@@ -398,15 +409,31 @@ func (c *Client) getInstancesFromDescribeInstancesOutput(result *ec2.DescribeIns
 				continue
 			}
 			ret = append(ret, &InstanceImpl{
-				ec2Client: c.ec2Client,
-				id:        *instance.InstanceId,
-				region:    regionID,
+				ec2Client:        c.ec2Client,
+				id:               *instance.InstanceId,
+				region:           regionID,
+				availabilityZone: *instance.Placement.AvailabilityZone,
```
FrimIdan, line 415: are we sure `instance.Placement` or `instance.Placement.AvailabilityZone` are not nil?

akpsgit: Actually, why do we even need AZ?

runtime_scan/pkg/provider/aws/instance.go
akpsgit: it's not correct to concat the security groups as it's not one within the other. 
I think that it's better to move the Security groups out of the location into a new property of the instance (list of SGs). WDYT?

ui_backend/pkg/rest/controllers.go
akpsgit: This dashboard is specific to AWS. We can discuss if we want to use general terminology there.

FrimIdan: if the widget is AWS specific maybe we should add it to the api path name or description

runtime_scan/pkg/provider/aws/instance.go
FrimIdan: sgtm

runtime_scan/pkg/provider/aws/client.go
akpsgit: Agreed to add a validation function of the instance object, print the invalid field and skip the instance.

runtime_scan/pkg/types/types.go
```diff
@@ -38,6 +39,11 @@ type ScanJobRunConfig struct {
 
 type Instance interface {
 	GetID() string
+	GetImage() string
+	GetType() string
+	GetLaunchTime() time.Time
+	GetPlatform() string
+	GetTags() []Tag
```
akpsgit, line 46: Please add get SGs

ui_backend/pkg/rest/controllers.go
akpsgit: Need to move this to a function (with unit test). Need to do the split only for AWS provider, for other providers need to just return the location

api/openapi.yaml
```diff
@@ -1453,6 +1453,11 @@ components:
           items:
             $ref: '#/components/schemas/Tag'
           nullable: true
+        securityGroups:
+          type: array
+          items:
+            $ref: '#/components/schemas/AwsSecurityGroup'
```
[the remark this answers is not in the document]
akpsgit: Agree, we should create a new object.

runtime_scan/pkg/provider/aws/client.go
```diff
@@ -419,17 +423,39 @@ func (c *Client) getInstancesFromDescribeInstancesOutput(result *ec2.DescribeIns
...
 	}
 	return ret
 }
 
-func getSecurityGroupsNames(sg []ec2types.GroupIdentifier) []string {
+func validateInstanceFields(instance ec2types.Instance) error {
+	if instance.InstanceId == nil {
+		return fmt.Errorf("instance id does not exist")
+	}
+	if instance.Placement == nil || instance.Placement.AvailabilityZone == nil {
```
akpsgit, line 441: maybe we should split this if in 2 to distinguish?

akpsgit (approved):

PR #308, opened 2023-05-10
Cli cicd mode

pbalogh-sa:
## Description

Please provide a meaningful description of what this change will do, or is for. Bonus points for including links to
related issues, other PRs, or technical references.

Note that by _not_ including a description, you are asking reviewers to do extra work to understand the context of this
change, which may lead to your PR taking much longer to review, or result in it not being reviewed at all.

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

- [ ] I have read the [contributing guidelines](/CONTRIBUTING.md)


cli/cmd/root.go
```diff
@@ -150,21 +173,66 @@ func init() {
...
+	}
+}
+
+func getConfigFromBackend() *families.Config {
+	if server == "" {
+		panic("Missing backend")
+	}
+	client, err := backendclient.Create(server)
+	if err != nil {
+		logrus.Errorf("failed to create VMClarity API client: %v", err)
```
FrimIdan, line 212: an error should be returned. you can't continue after the failure

cli/cmd/root.go
```diff
@@ -150,21 +173,66 @@ func init() {
...
+	if len(*scanConfigs.Items) == 0 {
+		panic(fmt.Sprintf("There is no scan config with name=%s", scanConfigName))
+	}
+
+	scanConfig := families.CreateFamilyConfigFromModel(
+		(*scanConfigs.Items)[0].ScanFamiliesConfig,
+		families.LoadAddresses("localhost"),
+		families.LoadPaths(),
+	)
+	scanConfigID = *(*scanConfigs.Items)[0].Id
```
FrimIdan, line 230: is it a get or a set function?

cli/cmd/root.go
```diff
@@ -207,13 +292,17 @@ func initLogger() {
...
 	var err error
 
 	if config == nil {
 		return nil, errors.New("families config must not be nil")
 	}
 
-	if server != "" {
+	updater = cicdupdater.NewDefaultUpdater()
+
+	if (server != "" && !cicdMode) || (cicdMode && exportCICDResults) {
```
FrimIdan, line 306: what if `(cicdMode && exportCICDResults)` but server is empty?

cli/cmd/root.go
```diff
@@ -222,6 +311,22 @@ func newCli() (*cli.CLI, error) {
 			return nil, fmt.Errorf("failed to create VMClarity API client: %w", err)
 		}
 
+		if exportCICDResults && scanResultID == "" {
+			i, err := cicdinitiator.NewInitiator(client, config, scanConfigID, scanConfigName, input, inputType)
```
FrimIdan, line 316: looks like `client, config, scanConfigID, scanConfigName, input, inputType` can be put in a `cicdinitiatorConfig`

cli/cmd/root.go
```diff
@@ -222,6 +311,22 @@ func newCli() (*cli.CLI, error) {
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to create VMClarity initiator: %w", err)
+			}
+			var scanID string
+			scanID, scanResultID, err = i.InitResults(context.TODO())
+			if err != nil {
+				return nil, fmt.Errorf("failed to init scan result: %w", err)
+			}
```
FrimIdan, line 324: seems like all the logic to get the `scanID` and `scanResultID` can be extracted to a function, the `initiator` is not used after that, right?

cli/cmd/root.go
```diff
@@ -252,64 +357,81 @@ func newCli() (*cli.CLI, error) {
...
+	return familiesConfig
+}
+
+func setInput(input, inputType string, familiesConfig *families.Config) *families.Config {
+	if familiesConfig.SBOM.Enabled {
+		familiesConfig.SBOM.Inputs = append(familiesConfig.SBOM.Inputs, sbom.Input{
+			Input:     input,
+			InputType: inputType,
+		})
+	}
```
FrimIdan, line 378: Seems like it's more an append input and not set

cli/pkg/cicd/initiator/initiator.go
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package initiator
+
+import "context"
+
+type Initiator interface {
+	InitResults(ctx context.Context) (string, string, error)
```
FrimIdan, line 21: better to add comment on the purpose of the method and what it returns.

cli/pkg/cicd/initiator/vmclarity.go
```diff
@@ -0,0 +1,231 @@
...
+		return "VMInfo", nil
+	default:
+		return "", errors.New("unsupported target type")
+	}
+}
+
+func (i *VMClarityInitiator) InitResults(ctx context.Context) (string, string, error) {
+	targetID, err := i.createTarget(ctx)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to init results: %v", err)
```
FrimIdan, line 84: please fix the error message in all calls, they all have `failed to init results...`

cli/pkg/cicd/updater/vmclarity.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (u *VMClarityUpdater) UpdateScan(ctx context.Context) error {
+	scan, err := u.updatedScanSummary(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to update scan results summary: %v", err)
+	}
+
+	scan.EndTime = utils.PointerTo(time.Now())
+	scan.State = utils.PointerTo(models.ScanStateDone)
+	scan.StateMessage = utils.PointerTo("All scan jobs completed")
```
FrimIdan, line 54: `All scan jobs completed` - we need to make it const that will be used in all places.

cli/pkg/cicd/vminfoprovider/aws.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return "", "", fmt.Errorf("failed to get availability-zone: %v", err)
+	}
+	defer azResp.Body.Close()
+	availabilityZone, err := io.ReadAll(azResp.Body)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to read availability-zone response body: %v", err)
+	}
+	az := string(availabilityZone)
+	return string(instanceID), az[:len(az)-1], nil
```
FrimIdan, line 51: `az[:len(az)-1]` - can you comment why?

api/openapi.yaml
```diff
@@ -1477,6 +1483,50 @@ components:
...
+          type: string
+        imageName:
+          type: string
+        imageVersion:
+          type: string
+        arch:
+          type: string
+        imageID:
+          type: string
+        location:
```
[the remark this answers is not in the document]
pbalogh-sa: In the case of AWS the AMIs are stored in regions The same image has different AMI IDs in different regions.

cli/cmd/root.go
```diff
@@ -150,21 +173,66 @@ func init() {
...
 	// will be global for your application.
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.vmclarity.yaml)")
-	rootCmd.PersistentFlags().StringVar(&output, "output", "", "set output directory path. Stdout is used if not set.")
+	rootCmd.PersistentFlags().StringVar(&output, "output", "", "set output directory path. Stdout is used if not set")
 	rootCmd.PersistentFlags().StringVar(&server, "server", "", "VMClarity server to export scan results to, for example: http://localhost:9999/api")
+	rootCmd.PersistentFlags().StringVar(&scanConfigName, "scan-config-name", "", "use an existing scan config that is defined in the VMClarity server")
 	rootCmd.PersistentFlags().StringVar(&scanResultID, "scan-result-id", "", "the ScanResult ID to export the scan results to")
 	rootCmd.PersistentFlags().BoolVar(&mountVolume, "mount-attached-volume", false, "discover for an attached volume and mount it before the scan")
 	rootCmd.PersistentFlags().BoolVar(&waitForServerAttached, "wait-for-server-attached", false, "wait for the VMClarity server to attach the volume")
+	rootCmd.PersistentFlags().StringVar(&input, "input", "", "input for families")
```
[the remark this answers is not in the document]
pbalogh-sa: The input can be a different type in the case of CICD mode. For example, it can be a directory and in the future we have to support container image, container info, and vm image input type. It makes sense only CICD mode because in orchestrated mode we are scanning VMs at the moment.

cli/cmd/root.go
pbalogh-sa: It is get config from the backend and convert it to families config.

cli/cmd/root.go
pbalogh-sa: It can't be empty because of the validation of the flags:
	`validateRequiredFlagForDefinedFlag(rootCmd, "export-cicd-results", "server")`
	
If the export is set the server must be set as well.

cli/pkg/cicd/vminfoprovider/aws.go
pbalogh-sa: cut the end of the AZ to get just the region. I add a comment

cli/pkg/cicd/initiator/vmclarity.go
```diff
@@ -0,0 +1,231 @@
...
+		Summary:   createInitScanSummary(),
+	}
+
+	createdScan, err := i.client.PostScan(ctx, *scan)
+	if err != nil {
+		var conErr backendclient.ScanConflictError
+		if errors.As(err, &conErr) {
+			logrus.Infof("Scan already exist. scan id=%v.", *conErr.ConflictingScan.Id)
+			return *conErr.ConflictingScan.Id, nil
+		}
```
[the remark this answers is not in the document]
pbalogh-sa: Shouldn't

cli/cmd/root.go
pbalogh-sa: Yes, I put it into a function.

cli/pkg/cicd/initiator/initiator.go
pbalogh-sa: I've added some comments.

cli/pkg/cicd/initiator/vmclarity.go
```diff
@@ -0,0 +1,231 @@
...
+		if errors.As(err, &conErr) {
+			logrus.Infof("Scan results already exist. scan result id=%v.", *conErr.ConflictingScanResult.Id)
+			return *conErr.ConflictingScanResult.Id, nil
+		}
+		return "", fmt.Errorf("failed to post scan result: %v", err)
+	}
+	return *createdScanResult.Id, nil
+}
+
+func createInitScanResultSummary() *models.ScanFindingsSummary {
```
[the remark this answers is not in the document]
pbalogh-sa: Now we get an exception if the summary is not set because the elements of it are pointers.

cli/cmd/root.go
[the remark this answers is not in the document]
sambetts-cisco: IMO the input for all this should be the VMClarity CLI input yaml, ideally the CLI should have little to no command line arguments so that backwards compatibility is easier to maintain. For example the yaml could look something like:

```
asset:
  type: VM
  ...

loadFamilySettingsFromScanConfig: <ID or Name>
  
sbom:
  inputs:
    - type: ROOTFS
       path: /
```

If we want to provide the option to run VMClarity CLI without a yaml input, then IMO CLI flags we provide should be direct overrides of the options in the yaml "helm style" e.g. `--set asset.type=VM`

cli/pkg/cli/cli.go
```diff
@@ -39,6 +40,7 @@ const (
 type CLI struct {
 	state.Manager
 	presenter.Presenter
+	updater.Updater
```
sambetts-cisco, line 43: I don't think there is a need for a new concept here, I think almost everything can be hidden inside the existing state.Manager and presenter.Presenter concepts, if we create a new "CICDStateManagerPresenter" which wraps the VMClarity state manager and presenter, the "WaitForVolumeAttachment" and "IsAborted" functions can be empty.

"MarkInProgress" can:
- Create the Asset as Defined by the user
- Create a Scan with just that Asset defined as the scanned asset
- Create a ScanResult with that Scan ID
- Call the VMClarity state manager MarkInProgress with the ScanResult ID

then "MarkDone" can:
- Call the normal VMClarity state manager MarkDone
- Update the Scan with the summary from the ScanResult and mark it as completed.

All the presenter functions will be passed straight through except the ScanResult ID from MarkInProgress will be injected. This way the work that Erez is doing to allow for incremental status updates etc will still work even in the CI/CD flow.

If we don't need the incremental updates stuff and we don't want a Scan in the system until we've succeeded, we can leave MarkInProgress empty, and then cache all the results as they arrive through our presenter interface, and then do everything we need to do in the "MarkDone".

runtime_scan/pkg/config/config.go
```diff
@@ -70,11 +62,8 @@ type ScannerConfig struct {
 	// address to use.
 	ScannerBackendAddress string
 
-	ExploitsDBAddress string
-
-	TrivyServerAddress string
-
-	GrypeServerAddress string
+	// Scanner server and exploit DB addresses for families
+	FamiliesAddresses families.Addresses
```
sambetts-cisco, line 66: Why are we changing this in this PR?

shared/pkg/families/config.go
sambetts-cisco, line 1: Not sure this belongs in shared, this requires families logic to know about the VMClarity API concepts. If we want to put it anywhere "common" it should be in the VMClarity CLI package, and then the runtime_scan can import it from there.

shared/pkg/families/config.go
```diff
@@ -16,15 +16,48 @@
...
+
+	GitleaksBinaryPath            = "GITLEAKS_BINARY_PATH"
+	ClamBinaryPath                = "CLAM_BINARY_PATH"
+	FreshclamBinaryPath           = "FRESHCLAM_BINARY_PATH"
+	AlternativeFreshclamMirrorURL = "ALTERNATIVE_FRESHCLAM_MIRROR_URL"
+	LynisInstallPath              = "LYNIS_INSTALL_PATH"
+	ChkrootkitBinaryPath          = "CHKROOTKIT_BINARY_PATH"
+	ExploitDBAddress              = "EXPLOIT_DB_ADDRESS"
+	TrivyServerAddress            = "TRIVY_SERVER_ADDRESS"
+	GrypeServerAddress            = "GRYPE_SERVER_ADDRESS"
```
sambetts-cisco, line 58: families logic should only receive a configuration struct from the outside, it should not be doing any loading of configurations itself, that is the responsibility of the CLI.

The CLI code uses viper + yaml, we should not be using environment variables like this anywhere in this flow, and if we want defaults then we should be handling it as part of loading that configuration. E.g. setting a default for rootkits.scannerconfig.chkrootkit.binarypath.

sambetts-cisco (changes requested):

cli/cmd/root.go
```diff
@@ -150,21 +175,63 @@ func init() {
 	// Cobra supports persistent flags, which, if defined here,
 	// will be global for your application.
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.vmclarity.yaml)")
-	rootCmd.PersistentFlags().StringVar(&output, "output", "", "set output directory path. Stdout is used if not set.")
+	rootCmd.PersistentFlags().StringVar(&output, "output", "", "set output directory path. Stdout is used if not set")
 	rootCmd.PersistentFlags().StringVar(&server, "server", "", "VMClarity server to export scan results to, for example: http://localhost:9999/api")
+	rootCmd.PersistentFlags().StringVar(&scanConfigName, "scan-config-name", "", "use an existing scan config that is defined in the VMClarity server")
```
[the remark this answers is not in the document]
sambetts-cisco: I'm suggesting that we support both, some people might prefer to use the ID in case someone on the server side renames the config and breaks everything

cli/pkg/cicd/vminfoprovider/aws.go
```diff
@@ -0,0 +1,54 @@
...
+	azResp, err := http.Get("http://169.254.169.254/latest/meta-data/placement/availability-zone") // nolint:noctx
+	if err != nil {
+		return "", "", fmt.Errorf("failed to get availability-zone: %v", err)
+	}
+	defer azResp.Body.Close()
+	availabilityZone, err := io.ReadAll(azResp.Body)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to read availability-zone response body: %v", err)
+	}
+	az := string(availabilityZone)
```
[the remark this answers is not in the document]
pbalogh-sa: thx, I missed it somehow.

backend/pkg/database/gorm/target.go
```diff
@@ -281,6 +281,24 @@ func (t *TargetsTableHandler) checkUniqueness(target models.Target) (*models.Tar
...
+		if err != nil {
+			return nil, err
+		}
+		if len(targets) > 0 {
+			var apiTarget models.Target
+			if err := json.Unmarshal(targets[0].Data, &apiTarget); err != nil {
+				return nil, fmt.Errorf("failed to convert DB model to API model: %w", err)
+			}
+			return &apiTarget, &common.ConflictError{
+				Reason: fmt.Sprintf("Target directory exists with same name=%q and location=%q", info.DirName, info.Location),
```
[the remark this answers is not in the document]
FrimIdan: I think it was not addressed

cli/cmd/root.go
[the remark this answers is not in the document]
FrimIdan: agree, not sure why we need different cli flags for that mode

cli/cmd/root.go
```diff
@@ -150,21 +174,64 @@ func init() {
...
+		}
+	}
 }
 
-// initConfig reads in config file and ENV variables if set.
-func initConfig() {
-	logrus.Infof("init config")
+func getConfigFromBackend(cliConf *cliconfig.Config) *cliconfig.Config {
+	if server == "" {
+		panic("Missing backend")
```
FrimIdan, line 205: please don't use `panic`

cli/cmd/root.go
FrimIdan: I'm a bit confused, the same object type is an input an output for the function.
If the purpose is to set the `scanConfig` on the `cliConf` I would split it into "get" and do the "set" outside.
And IMO, the input for the "get" function doesn't need to be the `cliConf` but only what is needed from it.

cli/cmd/root.go
FrimIdan: I would add a comment here, it just seems weird

cli/pkg/cicd/initiator/vmclarity.go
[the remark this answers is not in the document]
FrimIdan: Sure, not related to this task, @akpsgit / @sambetts-cisco maybe we need to open an issue to remove the need for an empty initialization if needed.

shared/pkg/utils/utils.go
```diff
@@ -27,6 +27,7 @@ import (
 const (
 	DefaultResourceReadyWaitTimeoutMin   = 3
 	DefaultResourceReadyCheckIntervalSec = 3
+	AllScanJobsCompleted                 = "All scan jobs completed"
```
FrimIdan, line 30: please rename to `AllScanJobsCompletedMessage`

cli/cmd/root.go
pbalogh-sa: This piece of code was removed because of a small refactor. So it isn't relevant anymore.

api/openapi.yaml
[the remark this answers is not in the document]
pbalogh-sa: `vmImage`type is removed now because we don't support yet

akpsgit:
#334

runtime_scan/pkg/config/config.go
pbalogh-sa: I changed back

shared/pkg/families/config.go
pbalogh-sa: I removed from families

cli/pkg/cli/cli.go
pbalogh-sa: The update was removed from the cli

cli/pkg/cicd/vminfoprovider/aws.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// cut the last character from the availability-zone in order to return only the region
+	return string(instanceID), az[:len(az)-1], nil
+}
```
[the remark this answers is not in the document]
pbalogh-sa: Now we can set the instanceID and location using `--set asset.instanceID=instance --set asset.location=location`
If they are not set the incanceID and location will be withdrawn from the instance.

api/openapi.yaml
```diff
@@ -1442,12 +1442,18 @@ components:
         - $ref: '#/components/schemas/VMInfo'
         - $ref: '#/components/schemas/PodInfo'
         - $ref: '#/components/schemas/DirInfo'
+        - $ref: '#/components/schemas/VMImage'
+        - $ref: '#/components/schemas/ContainerInfo'
```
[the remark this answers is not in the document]
pbalogh-sa: The new target types were removed now.

backend/pkg/database/gorm/target.go
```diff
@@ -281,6 +282,24 @@ func (t *TargetsTableHandler) checkUniqueness(target models.Target) (*models.Tar
...
+		if len(targets) > 0 {
+			var apiTarget models.Target
+			if err := json.Unmarshal(targets[0].Data, &apiTarget); err != nil {
+				return nil, fmt.Errorf("failed to convert DB model to API model: %w", err)
+			}
+			return &apiTarget, &common.ConflictError{
+				Reason: fmt.Sprintf("Target directory exists with same name=%q and location=%q", *info.DirName, *info.Location),
+			}
+		}
+		return nil, nil // nolint:nilnil
```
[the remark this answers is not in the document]
pbalogh-sa: new targets were removed

api/openapi.yaml
```diff
@@ -1512,6 +1518,50 @@ components:
...
+      required:
+        - objectType
+
+    ContainerInfo:
+      type: object
+      properties:
+        objectType:
+          type: string
+        containerName:
+          type: string
```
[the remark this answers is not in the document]
pbalogh-sa: new targets were removed

shared/pkg/backendclient/client.go
```diff
@@ -619,3 +619,34 @@ func (b *BackendClient) PostFinding(ctx context.Context, finding models.Finding)
 		return nil, fmt.Errorf("failed to create a finding. status code=%v", resp.StatusCode())
 	}
 }
+
```
chrisgacsal, line 622: This bloats the `backendclient` unnecessarily https://github.com/openclarity/vmclarity/pull/308/files#diff-189663ef212b0528411bdd1bc250e2c49471f3ae97cf265d63bf6d087212b314R623-R652
(btw, do not get having the `GetScanResultSummary` either)

shared/pkg/families/config.go
```diff
@@ -16,15 +16,35 @@
...
+	chkrootkitConfig "github.com/openclarity/vmclarity/shared/pkg/families/rootkits/chkrootkit/config"
+	rootkitsCommon "github.com/openclarity/vmclarity/shared/pkg/families/rootkits/common"
 	"github.com/openclarity/vmclarity/shared/pkg/families/sbom"
 	"github.com/openclarity/vmclarity/shared/pkg/families/secrets"
+	"github.com/openclarity/vmclarity/shared/pkg/families/secrets/common"
+	gitleaksconfig "github.com/openclarity/vmclarity/shared/pkg/families/secrets/gitleaks/config"
 	"github.com/openclarity/vmclarity/shared/pkg/families/vulnerabilities"
 )
 
+const (
```
chrisgacsal, line 43: Are these default values? If so, please add `Default` prefix to their names. Otherwise these do not belong here in my opinion.

chrisgacsal (changes requested):

PR #314, opened 2023-05-18
Ensure that the reconcile queue tracks processing items

sambetts-cisco:
## Description

This commit adds some new state and functions to the orchestration queue so that it keeps track of Dequeued items until that item is marked as Done by the person who Dequeued it.

This information is used to prevent items from being re-queued while the system is processing an item because the poller feeding the queue may pick up an item which is in the process of being updated.

This also makes it safe to have multiple reconciler threads being fed from a single queue because we have a guarentee that the same item isn't in the queue twice and therefore we can't end up with mutiple reconcilers trying to process the same item.

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


FrimIdan:
> This information is used to prevent items from being re-queued while the system is processing an item because the poller feeding the queue may pick up an item which is in the process of being updated.

Not sure I understand, why that an item that was de-queued will be re-queued?

sambetts-cisco:
> @sambetts-cisco Not sure I understand, why that an item that was de-queued will be re-queued? Also, why we are not removing it from the `inqueue` once done is called? why we need both `inqueue` and `processing`?

The poller searching for items in specific states continues to run while the reconciler processes an item, which can result in the poller re-adding the item which is being processed to the end of the queue, because the queue doesn't know about this item any longer.

By keeping track of the items which are being processed we can avoid readding these items and avoid an extra reconcile when we have a single reconciler, if we enable parallel reconcilers for some objects then its required to avoid two reconcilers trying to reconcile the same item.

I original implemented this re-using the inqueue map, but that broke the behaviour I wanted for the "Has" function which is meant to check if something is in the queue, so I needed to maintain the list of processing items separately which is why I added the `processing` map.

FrimIdan:
> I original implemented this re-using the inqueue map, but that broke the behaviour I wanted for the "Has" function which is meant to check if something is in the queue, so I needed to maintain the list of processing items separately which is why I added the `processing` map.

```
Has returns true if items is currently in the queue
```
So the `Has` is for checking if the item is in the queue or in processing or both? From the code I see that it both in the queue and in processing so not sure, again, why we can't keep it in the queue until it was done processing.

FrimIdan:
> `Has` is for checking if an item is in the queue itself, dequeued items would return false even though they continue to be processed.

```
	_, inQueue := q.inqueue[item]
	_, isProcessing := q.processing[item]
	return inQueue || isProcessing
```

Maybe i'm missing something but if the item continue to be processed the `Has` will return true.

sambetts-cisco:
>  Maybe i'm missing something but if the item continue to be processed the Has will return true.

*face plam* I must have changed that at some point during development and now I don't remember why...

FrimIdan (dismissed):

runtime_scan/pkg/orchestrator/common/reconciler.go
```diff
@@ -52,6 +52,7 @@ func (r *Reconciler[T]) Start(ctx context.Context) {
 				}
 				cancel()
 			}
+			r.Queue.Done(item)
```
FrimIdan, line 55: @sambetts-cisco If `Dequeue` returned an error, should we call `Done` here?

sambetts-cisco: I actually fixed that in the next PR, I think this is a bug... item will be an empty T so it won't break because of the implementation but its not good either. I'll push a revision thanks for catching!

FrimIdan (approved):

akpsgit (approved):

PR #318, opened 2023-05-23
Add EnqueueAfter and RequeueAfter option to queue

sambetts-cisco:
## Description

This commit adds an option for an item to be enqueued after a specified
duration. During that duration the item can not be enqueued if requested
through Enqueue to ensure that the item is not enqueued before that
duration is completed.

This allows reconcilers to delay reconciliation for an item when it
knows it will be performing an operation which takes longer than the
polling cycle preventing the poller from requeuing that item.

RequeueAfter is a wrapper around Done() followed by EnqueueAfter() done
under the same lock to prevent race condition when an item could be
requeued between Done and the EnqueueAfter call.

Also adds a new error type, RequeueAfterError, which can be returned by
a reconcile function to indicate that an item should be requeued after a
certain duration. This should only be used as an optimisation and the
pollers and reconcilers should be designed to pick up correctly from a
restart of the system.

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/runtime_scan/pkg/orchestrator/common/queue.go
+++ b/runtime_scan/pkg/orchestrator/common/queue.go
@@ -19,15 +19,18 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"time"
 )
 
 type Enqueuer[T comparable] interface {
 	Enqueue(item T)
+	EnqueueAfter(item T, d time.Duration)
 }
 
 type Dequeuer[T comparable] interface {
 	Dequeue(ctx context.Context) (T, error)
-	Done(T)
+	Done(item T)
+	RequeueAfter(item T, d time.Duration)
 }
 
 type Queue[T comparable] struct {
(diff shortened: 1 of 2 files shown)


chrisgacsal (approved):
LGTM! :shipit:

runtime_scan/pkg/orchestrator/common/queue.go
```diff
@@ -108,9 +117,49 @@ func (q *Queue[T]) Enqueue(item T) {
...
+
+// Internal enqueueAfter function that it can be reused by the public
+// EnqueueAfter and public RequeueAfter functions, these should not be called
+// without obtaining a lock on Queue first.
+func (q *Queue[T]) enqueueAfter(item T, d time.Duration) {
+	_, inQueue := q.inqueue[item]
+	_, isProcessing := q.processing[item]
+	_, isWaitingForEnqueue := q.waitingForEnqueue[item]
+	if inQueue || isProcessing || isWaitingForEnqueue {
+		// item is already known by the queue so there is nothing to do
```
FrimIdan, line 142: @sambetts-cisco maybe a warning is needed here, it is probably a "bug" that you try to `enqueueAfter` for an already known item.

PR #320, opened 2023-05-24
Adopt reconciler logic for orchestrator

chrisgacsal:
## Description
The main purpose for this PR to adopt the reconciler (from the `runtime_scan/orchestrator/common` package)
based workflow for managing API objects such as ScanConfig, Scan, Target, ScanResult.
This also required to refactor the Provider as all operations must be idempotent as the Scan life-cycle is broken up
into multiple stages where controllers handle a single transition state during a reconciling event.

### Orchestrator (and Scanner)

The Orchestrator is responsible for managing controllers for each API object type: ScanConfig, Scan, ScanResult.

#### ScanConfigWatcher

The `ScanConfigWatcher` is responsible for scheduling/handling `Scans` based on `ScanConfig`.
When the `ScanConfig` has both the `Schedule.operationTime` and `Schedule.CronLine` defined and the latter represents
a single point in time (Quartz), the latter is used for defining the actual `operationTime` for new `Scans`.

#### ScanWatcher

The `ScanWatcher` periodically polls the API to detect if the `Scan` objects are scheduled to be run. In case there is
new `Scan` scheduled, it invokes the Provider to get the `Targets` in scope. When `Targets` are available it creates
`ScanResult` object for each. It monitors in-progress `Scans` and updates their Summary based on the information
gathered from the corresponding `ScanResult` objects.

#### ScanResultWatcher

The `ScanResultWatcher` monitors the API to find `ScanResult` objects which are in transient state based on
their `status.General.State` and `resourceCleanup` fields. It is responsible for invoking `RunTargetScan` method of the
`Provider` in order to set up the scanner instance and start the scan job. The `ScanResult` is marked as `DONE` with errors
(it is considered failed) if the `Provider` was unable to set up the scanner instance due to a permanent error.
It also does perform cleanup of the resources created for scan job by invoking the `RemoveTargetScan` method of the provider
according to the `DeleteJobPolicy` configuration.

#### Interface

The `Provider` interface has been split into two (`Discoverer` and `Scanner`) in order to separate responsibilities for
discovering scan scopes and actually performing scans.

Both `Discoverer` and `Scanner` interfaces have been reworked in order to have better separation of concerns between
the `Provider` and the caller (`Orchestrator`), like leaking provider specific data back to the caller.

The new `Scanner` interface also requires any implementation to be idempotent.

### VMClarity-Scanner (aka cli)

The `vmclarity-scanner` does not have timeout for waiting for volume attachment anymore, instead it monitors
the corresponding `ScanResult` and raises an error if the state for `ScanResult` is changed to `DONE` meaning that
scanning cannot be performed due to external (mostly due to provider) error.

### Bugfixes/Improvements

* fix missing `/dev` mount for `vmclarity-scanner` container which prevented it to detect when the scanner volume is mounted
* make `provider/aws` implementation conform with the new `Provider` interface including the requirement for idempotency

### Issues

* https://github.com/openclarity/vmclarity/issues/336
* https://github.com/openclarity/vmclarity/issues/335

## Type of Change

[x] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[x] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


sambetts-cisco (dismissed):
LGTM, all comments as minor that can be addressed in follow ups. I think the AWS provider can probably be cleaned up a lot now that the interface is simpler (remove all the wrapping interfaces for instance/snapshot etc)

installation/aws/VmClarity.cfn
```diff
@@ -623,6 +623,11 @@ Resources:
               "ForAllValues:StringEquals":
                 "aws:TagKeys":
                   - Owner
+                  - Name
+                  - VMClarity.ScanID
```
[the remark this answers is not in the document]
chrisgacsal: My plan is to remove it in a separate PR, right after this one got merged.

installation/aws/VmClarity.cfn
```diff
@@ -776,6 +814,15 @@ Resources:
...
           # ##########################
+
+          # ##########################
+          #
+          - Effect: "Allow"
+            Action:
+              - "sts:DecodeAuthorizationMessage"
+            Resource: "*"
+          #
+          # ##########################
```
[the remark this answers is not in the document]
chrisgacsal: I'll remove this now as it was used only for debugging. It was not meant to be part of the change.

runtime_scan/pkg/provider/aws/client.go
```diff
@@ -126,419 +140,705 @@ func (c *Client) DiscoverInstances(ctx context.Context, scanScope *models.ScanSc
...
+		"TargetInstanceID": vmInfo.InstanceID,
+		"TargetLocation":   vmInfo.Location,
+		"ScannerLocation":  config.Region,
+		"Provider":         string(c.Kind()),
+	})
+
+	// Note(chrisgacsal): In order to speed up the initialization process the scanner instance and the volume are created
+	//                    in parallel.
+
+	// Create scanner instance in same region/location as the VMClarity server deployed
```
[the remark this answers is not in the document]
chrisgacsal: You are right. Removing it now.

runtime_scan/pkg/provider/aws/volume.go
```diff
@@ -22,111 +22,208 @@ import (
...
+		return false, FatalError{
+			Err: fmt.Errorf("volume is not ready due to its state: %s", volState),
+		}
+	default:
+	}
+
+	return false, nil
+}
+
+func (v *Volume) WaitForAttached(ctx context.Context, timeout time.Duration, interval time.Duration) error {
```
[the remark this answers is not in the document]
chrisgacsal: Yup. Forgot to remove it. Fixing it now.

shared/pkg/utils/errors.go
```diff
@@ -0,0 +1,69 @@
...
+type (
+	singleWrappedError = interface{ Unwrap() error }
+	multiWrappedError  = interface{ Unwrap() []error }
+)
+
+// UnwrapErrors returns a slice of errors by unwrapping err error.
+// Returned slice is nil if err is nil.
+// Slice with the original err as a single element is returned if err does not implement `Unwrap() error`
+// or `Unwrap []error` interfaces.
+// UnwrapErrors does not perform recursive lookup, so only the provider err is unwrapped.
```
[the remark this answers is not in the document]
chrisgacsal: Yup. Originally it was part of the `provider` package. Will update it now.

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+}
+
+func (w ScheduleWindow) String() string {
+	return fmt.Sprintf("start: %s, end: %s", w.start.Format(time.RFC3339), w.end.Format(time.RFC3339))
+}
+
+// nolint:gomnd
+func NewScheduleWindow(now time.Time, size time.Duration) *ScheduleWindow {
+	return &ScheduleWindow{
```
FrimIdan, line 76: nit: it might be better to save the `size` to avoid calculating `w.end.Sub(w.start)` every  `Next()` and `Prev()`

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
```diff
@@ -0,0 +1,264 @@
...
+				OperationTime: operationTime,
+				Window:        window,
+			}, nil
+		}
+		return &ScanConfigSchedule{
+			State:         ScheduleStateOverdue,
+			OperationTime: operationTime,
+			Window:        window,
+		}, nil
+	}
```
FrimIdan, line 257: Can you add a comment here? The `window.Before(operationTime.Time())` is a bit confusing for me, does it means that the operation time is before the window or after the window?
As I understand it, the time is before the window, so not sure why it is `Overdue`?

runtime_scan/pkg/orchestrator/scanconfigwatcher/watcher.go
```diff
@@ -0,0 +1,250 @@
...
+
+	return nil
+}
+
+func (w *Watcher) reconcileDue(ctx context.Context, scanConfig *models.ScanConfig, schedule *ScanConfigSchedule) error {
+	if err := w.createScan(ctx, scanConfig); err != nil {
+		return fmt.Errorf("failed to reconcile new Scan for ScanConfig. ScanConfigID=%s: %w", *scanConfig.Id, err)
+	}
+	nextOperationTime := schedule.OperationTime.NextAfter(schedule.Window.Next().Start())
+	// FIXME: disable ScanConfig if it was a oneshot
```
FrimIdan, line 177: what about this one?

runtime_scan/pkg/orchestrator/scanresultwatcher/event.go
```diff
@@ -0,0 +1,42 @@
...
+func (e *ScanResultReconcileEvent) ToFields() log.Fields {
+	return log.Fields{
+		"ScanResultID": e.ScanResultID,
+		"ScanID":       e.ScanID,
+		"TargetID":     e.TargetID,
+	}
+}
+
+func (e ScanResultReconcileEvent) String() string {
+	return fmt.Sprintf("ScanResultID=%s ScanID=%s TargetID=%s", e.ScanID, e.ScanID, e.TargetID)
```
FrimIdan, line 41: The first `e.ScanID` should be `e.ScanResultID`

runtime_scan/pkg/provider/aws/instance.go
```diff
@@ -18,139 +18,80 @@ package aws
...
+
+	Metadata provider.ScanMetadata
+
+	ec2Client *ec2.Client
 }
 
-func (i *InstanceImpl) GetID() string {
-	return i.id
+func (i *Instance) Location() string {
+	return i.Region + "/" + i.VpcID
```
FrimIdan, line 53: you should use `LocationSeparator`

cli/pkg/state/vmclarity.go
```diff
@@ -40,29 +40,36 @@ type VMClarityState struct {
...
+	timer := time.NewTicker(DefaultWaitForVolRetryInterval)
 	defer timer.Stop()
 
 	for {
 		select {
 		case <-timer.C:
 			status, err := v.client.GetScanResultStatus(ctx, v.scanResultID)
 			if err != nil {
-				return fmt.Errorf("failed to get scan result status: %w", err)
+				logger.Errorf("failed to get scan result status: %v", err)
```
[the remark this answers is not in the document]
chrisgacsal: I wanted to make sure that the `scanner` won't fail in case there is a transient error where the `scanner` cannot reach the `backend` API. e.g. VMClarity server is being upgraded therefore it is not available for short period of time.

runtime_scan/pkg/orchestrator/scanconfigwatcher/watcher.go
chrisgacsal: I deliberately skipped fixing this as it is non-critical and  it would require rerunning E2E manually to validate the change, therefore I'll address this in a separate PR.

runtime_scan/pkg/provider/aws/instance.go
chrisgacsal: Nice catch. Fixing this now.

runtime_scan/pkg/orchestrator/scanresultwatcher/event.go
chrisgacsal: Nice catch. Fixing it now.

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
```diff
@@ -167,10 +196,17 @@ func isCronPeriodic(c *cronexpr.Expression) bool {
 type ScheduleState int8
 
 const (
+	// ScheduleStateDisabled means the ScanConfig isi disabled no Scan needs to be scheduled for it.
```
sambetts-cisco, line 199:
```suggestion
	// ScheduleStateDisabled means the ScanConfig is disabled no Scan needs to be scheduled for it.
```

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
FrimIdan: I missed the part that `ScheduleStateOverdue` will be handled later and will update the operation time.
Thanks for the clarification.

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
```diff
@@ -167,10 +196,17 @@ func isCronPeriodic(c *cronexpr.Expression) bool {
...
 	ScheduleStateDisabled ScheduleState = iota
+	// ScheduleStateUnscheduled means the ScanConfig does not define a schedule.
 	ScheduleStateUnscheduled
+	// ScheduleStateNotDue means the OperationTime for ScanConfig is in the future based on the current ScheduleWindow.
 	ScheduleStateNotDue
+	// ScheduleStateDue means the OperationTime for ScanConfig is in the current ScheduleWindow.
 	ScheduleStateDue
+	// ScheduleStateOverdue means the OperationTime for ScanConfig is in the past compared to
+	// the current ScheduleWindow, but it has recurring schedule, therefore the new OperationTime needs
+	// to be calculated which is in the next ScheduleWindow.
```
sambetts-cisco, line 209:
```suggestion
	// the current ScheduleWindow, but it has recurring schedule, therefore the new OperationTime needs
	// to be calculated from the cron expression.
```

I'm not sure "in the next ScheduleWindow" in the right wording here, I've provided a suggestion.

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
chrisgacsal: I'd keep it as is as it does not take a lot to compute Add/Sub on `int`.

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
```diff
@@ -241,14 +281,19 @@ func NewScanConfigSchedule(scanConfig *models.ScanConfig, window *ScheduleWindow
 		}, nil
 	}
 
+	// Check whether the operationTime is before the window meaning that it is in the past compared to the window.
 	if window.Before(operationTime.Time()) {
```
sambetts-cisco, line 285: As mentioned I think if we fix these functions to align with the time package it'll read much better this will become:

if window.After(operationTime.Time()) {
}

which is much better "pseudo english": "if the schedule window is after the operation time; then ... "

cli/pkg/state/vmclarity.go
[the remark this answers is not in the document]
sambetts-cisco: it'll retry on the next tick anyway until its aborted by the context so I think logging is ok here, WDYT?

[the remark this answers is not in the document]
chrisgacsal: > So maybe a retry logic is better, it feels that if the scanner will just skip the error here it will cause unexpected behavior later.

There is already an implicit retry logic here as the `break` only returns from the `select` but not from the `for` loop. So it will retry when the next arrives on `timer.C` channel.

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
```diff
@@ -0,0 +1,309 @@
...
+		if next.Time().Equal(o.Time()) {
+			break
+		}
+		o = *next
+	}
+
+	return &o
+}
+
+// Time returns tim.Time represented by the OperationTime.
```
FrimIdan, line 141: `tim.Time` --> `time.Time`

cli/pkg/state/vmclarity.go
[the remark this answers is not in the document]
FrimIdan: cool, missed it. tnx.

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/scanconfigwatcher/schedule.go
```diff
@@ -45,21 +49,25 @@ func (w *ScheduleWindow) In(t time.Time) bool {
 	return true
 }
 
+// Before returns true if the t time is before the timeframe represented by w ScheduleWindow, otherwise it returns false.
```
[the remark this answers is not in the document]
chrisgacsal: Fixed.

sambetts-cisco (approved):

FrimIdan (approved):

PR #321, opened 2023-05-24
Remodel Scan and TargetScanResult so that they can be used separately

sambetts-cisco:
## Description

This begins the work to move to a world where the ScanConfig and Scan are purely orchestration concepts where ScanConfig is responsible for scheduling Scans. Scans are responsible for batch orchestrating assets scans, and Asset Scans (TargetScanResults) are responsible for actually scanning.

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[X] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


api/openapi.yaml
```diff
@@ -1576,6 +1575,17 @@ components:
...
+          type: string
+        scanFamiliesConfig:
+          $ref: '#/components/schemas/ScanFamiliesConfig'
+        maxParallelScanners:
+          type: 'integer'
+          minimum: 1
+          maximum: 20
+          description: "The maximum number of scanners that can run in parallel for each scan"
+        scannerInstanceCreationConfig:
+          $ref: '#/components/schemas/ScannerInstanceCreationConfig'
```
FrimIdan, line 1588: Why `TargetScanResult` needs this info?

api/openapi.yaml
```diff
@@ -1612,6 +1622,63 @@ components:
       #  - target
       #  - scan
 
+    TargetScanResultRelationship:
```
FrimIdan, line 1625: maybe a comment here and in `TargetScanResult` that if one the field is modified the other object should be modified accordantly? This are basically the same objects.

api/openapi.yaml
```diff
@@ -2093,10 +2160,8 @@ components:
           # on the client side.
           #
           # readOnly: true
-        scan:
-          $ref: '#/components/schemas/ScanRelationship'
-        asset:
-          $ref: '#/components/schemas/TargetRelationship'
+        assetScan:
+          $ref: '#/components/schemas/TargetScanResultRelationship'
```
FrimIdan, line 2164: why we need this on finding? 
why `finding` is holding the `summary` of the scan and `maxParallelScanners` for example?

api/openapi.yaml
```diff
@@ -2093,10 +2160,8 @@ components:
           # on the client side.
           #
           # readOnly: true
-        scan:
-          $ref: '#/components/schemas/ScanRelationship'
-        asset:
-          $ref: '#/components/schemas/TargetRelationship'
+        assetScan:
```
FrimIdan, line 2163: `assetScan` --> `targetScanResult`?

runtime_scan/pkg/orchestrator/configwatcher/scan_runner.go
```diff
@@ -32,35 +32,32 @@ import (
...
 	}
 
-	scanner := _scanner.CreateScanner(scw.scannerConfig, scw.providerClient, scw.backendClient, scanConfig, targetInstances, scanID)
+	scanner := _scanner.CreateScanner(scw.scannerConfig, scw.providerClient, scw.backendClient, scan, targetInstances)
 	go scanner.Scan(ctx)
 
 	return nil
 }
 
 // initNewScan Initialized a new scan, returns target instances and scan ID.
```
FrimIdan, line 46: please update comment

api/openapi.yaml
sambetts-cisco: Think of TargetScanResult as "AssetScan" its a scan of a specific asset and can be reconciled and created separately. i.e. like a pod. This is part of the configuration for how to run that scan.

api/openapi.yaml
sambetts-cisco: This is a the assetScan which found the finding, a finding shouldn't have a relationship to a scan directly because going forward AssetScans could exist without a Scan.

api/openapi.yaml
```diff
@@ -966,8 +982,17 @@ components:
...
+          $ref: '#/components/schemas/ScanFamiliesConfig'
+        scope:
+          $ref: '#/components/schemas/ScanScopeType'
+        maxParallelScanners:
+          type: 'integer'
+          minimum: 1
+          maximum: 20
+          description: "The maximum number of scanners that can run in parallel for each scan"
+        scannerInstanceCreationConfig:
+          $ref: '#/components/schemas/ScannerInstanceCreationConfig'
```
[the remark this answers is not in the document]
sambetts-cisco: I think I mentioned this in the VMClarity channel, I was thinking about doing something like:

ScanConfig becomes:
```
schedule:
disabled: 
scanConfig:
    scope:
    assetScanConfig:
        scanFamilyConfig:
        instanceCreationConfig:
        etc...
```

Scan becomes:
```
scope:
assetScanConfig:
    scanFamilyConfig:
    instanceCreationConfig:
```

WDYT? This will allow us to create reusable objects, but its a bigger remodel. Personally I think it would be better like this.

api/openapi.yaml
sambetts-cisco: right now the concurrency is handled by the reconciler for a specific AssetScan, it looks to see if there are other AssetScans with the same Scan ID, and otherwise it won't schedule. Its something I want to revisit.

api/openapi.yaml
FrimIdan: > a finding shouldn't have a relationship to a scan directly because going forward AssetScans could exist without a Scan

But `TargetScanResultRelationship` holds info that is not relevant to finding, DYT?

api/openapi.yaml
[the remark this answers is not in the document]
FrimIdan: ok, we can do it later, just thought since you've now added the same 3/4 fields to different objects we can make them as a separate object.

api/openapi.yaml
FrimIdan: Not sure I understand, can you point me to the code?
AFAIK the `maxParallelScanners` is used by the job manager to handle the max number of workers that spin jobs.

api/openapi.yaml
sambetts-cisco: This is a relationship to the AssetScan/TargetScanResult which found the finding its not embeded information unless you $expand. But personally I think all the AssetScan information is relavant, because you may wish to check the configuration of the scan which found the finding etc.

PR #379, opened 2023-06-07
refactor: ScanResult states

chrisgacsal:
## Description

* use CamelCase for naming
* make `ScanResult` states more generic and granular to be applicable for different type of assets (VM, Dir, Pod, etc)

  | Old Name | New Name | Description | 
  |---|---|---|
  | `INIT` | `Pending` | Initial state for `ScanResult` waiting for being scheduled |
  | - | `Scheduled` | `ScanResult` which has been scheduled on `Provider` |
  | `ATTACHED` | `ReadyToScan` | `Provider` acknowledged that scan for `ScanResult` is ready to run |
  | `ABORTED` | `Aborted` | `ScanResult` has been aborted and the `vmclarity-scanner` needs to cancel all running scanners and shutdown gracefully |
  | `IN_PROGRESS` | `InProgress` | `vmclarity-scanner` is running scanners on Asset |
  | `DONE` | `Done` | `vmclarity-scanner` has finished with running all scanners on Asset |
  | `NOT_SCANNED` | `NotScanned` | Scanners are skipped for `ScanResult` |

* modify `ScanResultWatcher` by splitting the reconcile logic to handle `Pending` and `Scheduled` states separately
* rename `WaitForAttachment` in `vmclarity-scanner` to `WaitForReadyState` as the corresponding `ATTACHED` state has been changed to `ReadyToScan`.

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[x] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


cli/cmd/root.go
```diff
@@ -150,7 +150,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&server, "server", "", "VMClarity server to export scan results to, for example: http://localhost:9999/api")
 	rootCmd.PersistentFlags().StringVar(&scanResultID, "scan-result-id", "", "the ScanResult ID to export the scan results to")
 	rootCmd.PersistentFlags().BoolVar(&mountVolume, "mount-attached-volume", false, "discover for an attached volume and mount it before the scan")
-	rootCmd.PersistentFlags().BoolVar(&waitForServerAttached, "wait-for-server-attached", false, "wait for the VMClarity server to attach the volume")
+	rootCmd.PersistentFlags().BoolVar(&waitForReadyState, "wait-for-ready", false, "wait until the ScanResult state is set to ReadyToScan")
```
FrimIdan, line 153: Do we need this flag? When we will not want to wait for scan to be ready to be scanned?

sambetts-cisco: we don't want to wait when the CLI is standalone without server, and CICD case when there is no ScanResult to wait on.

 Perhaps we decide to WaitForReady based on being given the Server ID and ScanResult ID?

chrisgacsal: Both the `ScanResult` and `ReadyToScan` are part of the public API, so I would not describe them as internal. I think this help the user to understand better the use of this flag. Its kinda the same as the `--scan-result-id` flag.
But I have no preference on this one. I am good with changing the help string to your suggestion.

chrisgacsal: > Perhaps we decide to WaitForReady based on being given the Server ID and ScanResult ID?

This could work as well. I think this is the way to go.

FrimIdan: > This could work as well. I think this is the way to go.

So in another PR?

chrisgacsal: Nope, I'll update the PR with the suggestion. :)

chrisgacsal: Done. The `WaitForReadyState` will be called always and the `state.Manager` implementation will decide whether it needs to actually wait until the state is set to `ReadyToScan` or just skips it. I think it is better to delegate this to the  state manager.

FrimIdan: where is the code that the manager decide upon? or it is just because `state.NewLocalState()` is used when `server` is empty? I hope it works well for the cicd mode.

cli/cmd/root.go
```diff
@@ -86,14 +85,12 @@ var rootCmd = &cobra.Command{
...
 
-		if waitForServerAttached {
-			if err := cli.WaitForVolumeAttachment(abortCtx); err != nil {
-				err = fmt.Errorf("failed to wait for block device being attached: %w", err)
-				if e := cli.MarkDone(ctx, []error{err}); e != nil {
-					logger.Errorf("Failed to update scan result stat to completed with errors: %v", e)
-				}
-				return err
+		if err := cli.WaitForReadyState(abortCtx); err != nil {
+			err = fmt.Errorf("failed to wait for block device being attached: %w", err)
```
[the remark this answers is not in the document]
chrisgacsal: Good catch. Fixed it.

sambetts-cisco (approved):

api/openapi.yaml
```diff
@@ -1760,12 +1760,13 @@ components:
         state:
           type: string
           enum:
-            - NOT_SCANNED
-            - INIT
-            - ATTACHED
-            - IN_PROGRESS
-            - ABORTED
-            - DONE
+            - NotScanned
```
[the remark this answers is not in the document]
chrisgacsal: Good idea, but I'd do it in a separate PR if that is fine with you.

PR #388, opened 2023-06-09
refactor: provider configuration

chrisgacsal:
### Refactor

* removed `Provider` specific configuration form `orchestrator.Config`
* added `VMCLARITY_AWS` prefix to environment variables related to `aws` package
* switch from using `viper.Get*` getters to `viper.Unmarshal` to initialize `aws.Config`

### Documentation

* extended documentation with the configuration options for `Orchestrator` and `Provider` components.

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[x] Breaking Change  
[x] Refactor  
[x] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/installation/aws/VmClarity.cfn
+++ b/installation/aws/VmClarity.cfn
@@ -229,21 +229,46 @@ Resources:
               content:
                 Fn::Sub:
                   - |
-                    AWS_REGION=${AWS::Region}
-                    SCANNER_AWS_REGION=${AWS::Region}
-                    AWS_SUBNET_ID=${VmClarityScannerSubnet}
-                    AWS_SECURITY_GROUP_ID=${VmClarityScannerSecurityGroup}
-                    AWS_INSTANCE_TYPE=${ScannerInstanceType}
-                    SCANNER_KEY_PAIR_NAME=${KeyName}
-                    AWS_JOB_IMAGE_ID=${JobImageID}
+                    ##
+                    ## Orchestrator configuration
+                    ##
+                    # Host for the VMClarity backend server
                     BACKEND_REST_HOST=__BACKEND_REST_HOST__
+                    # Port number for the VMClarity backend server
                     BACKEND_REST_PORT=8888
+                    # Container image for Scanner instance
                     SCANNER_CONTAINER_IMAGE=${ScannerContainerImage}
+                    # Trivy server address
                     TRIVY_SERVER_ADDRESS=http://__BACKEND_REST_HOST__:9992
+                    # Grype server address
                     GRYPE_SERVER_ADDRESS=__BACKEND_REST_HOST__:9991
-                    DELETE_JOB_POLICY=${AssetScanDeletePolicy}
+                    # FreshClam mirror URL
                     ALTERNATIVE_FRESHCLAM_MIRROR_URL=http://__BACKEND_REST_HOST__:1000/clamav
-                  - JobImageID: !FindInMap
+                    # Resource cleanup policy
+                    DELETE_JOB_POLICY=${AssetScanDeletePolicy}
+                    # Provider to use
+                    PROVIDER=aws
+
+                    ##
+                    ## Provider configuration
+                    ##
+                    # The AWS region where the provider is deployed
+                    AWS_REGION=${AWS::Region}
+                    # Region where the Scanner instance needs to be created
+                    VMCLARITY_AWS_SCANNER_REGION=${AWS::Region}
+                    # SubnetID where the Scanner instance needs to be created
+                    VMCLARITY_AWS_SUBNET_ID=${VmClarityScannerSubnet}
+                    # SecurityGroupId which needs to be attached to the Scanner instance
+                    VMCLARITY_AWS_SECURITY_GROUP_ID=${VmClarityScannerSecurityGroup}
+                    # Name of the SSH KeyPair to use for Scanner instance launch
+                    VMCLARITY_AWS_KEYPAIR_NAME=${KeyName}
+                    # The AMI image used for creating Scanner instance
+                    VMCLARITY_AWS_SCANNER_AMI_ID=${ScannerImageID}
+                    # The instance type used for Scanner instance
+                    VMCLARITY_AWS_SCANNER_INSTANCE_TYPE=${ScannerInstanceType}
+                    # Block device name used for attaching Scanner volume to the Scanner instance
+                    #VMCLARITY_AWS_BLOCK_DEVICE_NAME=xvhd
+                  - ScannerImageID: !FindInMap
                       - AWSRegionArch2AMI
                       - !Ref "AWS::Region"
                       - !FindInMap
--- a/runtime_scan/pkg/orchestrator/config.go
+++ b/runtime_scan/pkg/orchestrator/config.go
@@ -19,31 +19,28 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/spf13/viper"
 
+	"github.com/openclarity/vmclarity/api/models"
 	"github.com/openclarity/vmclarity/runtime_scan/pkg/orchestrator/discovery"
 	"github.com/openclarity/vmclarity/runtime_scan/pkg/orchestrator/scanconfigwatcher"
 	"github.com/openclarity/vmclarity/runtime_scan/pkg/orchestrator/scanresultprocessor"
 	"github.com/openclarity/vmclarity/runtime_scan/pkg/orchestrator/scanresultwatcher"
 	"github.com/openclarity/vmclarity/runtime_scan/pkg/orchestrator/scanwatcher"
-	"github.com/openclarity/vmclarity/runtime_scan/pkg/provider/aws"
 )
 
 const (
-	ScannerAWSRegion              = "SCANNER_AWS_REGION"
 	DeleteJobPolicy               = "DELETE_JOB_POLICY"
 	ScannerContainerImage         = "SCANNER_CONTAINER_IMAGE"
-	ScannerKeyPairName            = "SCANNER_KEY_PAIR_NAME"
 	GitleaksBinaryPath            = "GITLEAKS_BINARY_PATH"
 	ClamBinaryPath                = "CLAM_BINARY_PATH"
 	FreshclamBinaryPath           = "FRESHCLAM_BINARY_PATH"
 	AlternativeFreshclamMirrorURL = "ALTERNATIVE_FRESHCLAM_MIRROR_URL"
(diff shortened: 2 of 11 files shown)


chrisgacsal:
I've verified the changes by running E2E test with success.

chrisgacsal:
I've verified the changes by re-running E2E test with success.

runtime_scan/pkg/provider/aws/config.go
```diff
@@ -15,40 +15,89 @@
...
+	// BlockDeviceName contains the block device name used for attaching Scanner volume to the Scanner instance
+	BlockDeviceName string
 }
 
-func setConfigDefaults() {
-	viper.SetDefault(AWSJobImageID, defaultAWSJobImageID)
-	viper.SetDefault(AWSInstanceType, defaultAWSInstanceType)
+func (c *Config) Validate() error {
+	if c.ScannerRegion == "" {
+		return fmt.Errorf("parameter Region must not be nil")
```
sambetts-cisco, line 56: I think "must be provided" would be better than "must not be nil" so as to not leak through go language?

installation/aws/VmClarity.cfn
```diff
@@ -229,21 +229,46 @@ Resources:
...
+                    # Resource cleanup policy
+                    DELETE_JOB_POLICY=${AssetScanDeletePolicy}
+                    # Provider to use
+                    PROVIDER=aws
+
+                    ##
+                    ## Provider configuration
+                    ##
+                    # The AWS region where the provider is deployed
+                    AWS_REGION=${AWS::Region}
```
sambetts-cisco, line 256: Is this needed for the AWS client still?

runtime_scan/pkg/provider/aws/client.go
```diff
@@ -37,12 +37,21 @@ import (
...
-func New(ctx context.Context, config *Config) (*Client, error) {
+func New(ctx context.Context) (*Client, error) {
+	config, err := NewConfig()
+	if err != nil {
+		return nil, fmt.Errorf("invalid configuration. Provider=AWS: %w", err)
+	}
+
+	if err = config.Validate(); err != nil {
+		return nil, fmt.Errorf("failed to validate provider configuration. Provider=AWS: %w", err)
+	}
```
sambetts-cisco, line 51: Is there much point in having this separate from NewConfig()?

runtime_scan/pkg/provider/aws/config.go
```diff
@@ -15,40 +15,89 @@
...
+	v := viper.New()
+
+	v.AllowEmptyEnv(true)
+	v.AutomaticEnv()
+
+	_ = v.BindEnv("ScannerRegion", EnvScannerRegion)
+	_ = v.BindEnv("SubnetID", EnvSubnetID)
+	_ = v.BindEnv("SecurityGroupID", EnvSecurityGroupID)
+	_ = v.BindEnv("KeyPairName", EnvKeyPairName)
+	_ = v.BindEnv("ScannerImage", EnvScannerImage)
```
sambetts-cisco, line 89: not sure we want to be eating the errors here right? I'm not sure what the errors might be but it would probably be a good idea to handle them

runtime_scan/pkg/provider/aws/config.go
```diff
@@ -15,40 +15,89 @@
...
 }
 
-func LoadConfig() *Config {
-	setConfigDefaults()
+func NewConfig() (*Config, error) {
+	// Avoid modifying the global instance
+	v := viper.New()
+
+	v.AllowEmptyEnv(true)
+	v.AutomaticEnv()
```
sambetts-cisco, line 83: I don't think we need AutomaticEnv if we're using BindEnv explicitly. Alternatively we could switch to:
```
v.SetEnvPrefix("VMCLARITY_AWS")
v.AutomaticEnv()
```
and avoid the need to define the environment variables and use BindEnv entirely. Only difference would be that `VMCLARITY_AWS_SCANNER_REGION` would become `VMCLARITY_AWS_SCANNERREGION` because of rule used by viper is:

> It will check for an environment variable with a name matching the key uppercased and prefixed with the EnvPrefix if set.

As we're now using Unmarshal, adding a new configuration option would then be as simple as adding a new item o the struct, no need to change anywhere else.

sambetts-cisco (changes requested):

installation/aws/VmClarity.cfn
chrisgacsal: Yes. This is for the AWS client which might be different than the Scanner region.

runtime_scan/pkg/provider/aws/config.go
chrisgacsal: I am aware of this, but to make the Env var readably, we need a `StringReplacer` which I am thinking about to add, so there will be no need for defining env vars, but will be automatically generated by `viper`. The downside of this that it is not straightforward to ppl who are not familiar with `viper`.

runtime_scan/pkg/provider/aws/config.go
chrisgacsal: We are eating errors as `BindEnv` always returns a `nil` in this case.

runtime_scan/pkg/provider/aws/client.go
chrisgacsal: The reason for having validation not being part of the `NewConfig`, because there might be cases where:
* the Config is updated after creating it with `NewConfig` therefore we do not want to fail the config creation
* the caller of `NewConfig` does custom validation

runtime_scan/pkg/provider/aws/config.go
chrisgacsal: Alright. Will update the wording here.

runtime_scan/pkg/provider/aws/config.go
sambetts-cisco: No need for a string replacer or anything custom, viper will just use the ToUppered struct key name + the prefix from SetEnvPrefix('VMCLARITY_AWS') so if the struct looks like:

```
struct {
  Foo string
}
```

then viper will look for `VMCLARITY_AWS_FOO` in the environment. This is all natively supported by viper.

As I said the only downside is that `VMCLARITY_AWS_SCANNER_REGION` would become `VMCLARITY_AWS_SCANNERREGION` because the struct looks like:

```
struct {
 ScannerRegion string
}
```

so `ScannerRegion` becomes `SCANNERREGION`.

I'd need to double check but viper also uses mapstructure tags, so it might be possible to do:

```
struct {
  ScannerRegion string `map_structure:'scanner_region'`
}
```

and keep the underscore, need to experiment though.

runtime_scan/pkg/orchestrator/config.go
```diff
@@ -123,18 +118,25 @@ func setConfigDefaults(backendHost string, backendPort int, backendBaseURL strin
...
 }
 
 func LoadConfig(backendHost string, backendPort int, baseURL string) (*Config, error) {
 	setConfigDefaults(backendHost, backendPort, baseURL)
 
-	c := &Config{
-		AWSConfig: aws.LoadConfig(),
+	var providerKind models.CloudProvider
+	switch strings.ToLower(viper.GetString(ProviderKind)) {
+	case "aws":
```
FrimIdan, line 131: why not to use `strings.ToLower(models.AWS)`?

runtime_scan/pkg/provider/aws/config.go
chrisgacsal: > As I said the only downside is that VMCLARITY_AWS_SCANNER_REGION would become VMCLARITY_AWS_SCANNERREGION because the struct looks like.

Thats the exact reason for me want to use a `StringReplacer` as the generated env vars would be hard to read.
Will add a `StringReplacer` in the next update.

runtime_scan/pkg/orchestrator/config.go
```diff
@@ -67,20 +64,21 @@ const (
...
 
 const (
-	DefaultScannerAWSRegion         = "us-east-1"
-	DefaultAttachedVolumeDeviceName = "xvdh"
-
 	DefaultTrivyServerTimeout = 5 * time.Minute
 	DefaultGrypeServerTimeout = 2 * time.Minute
 
 	DefaultControllerStartupDelay = 15 * time.Second
+	DefaultProviderKind           = "aws"
```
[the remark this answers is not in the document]
chrisgacsal: Will fix this now.

runtime_scan/pkg/provider/aws/config.go
chrisgacsal: Also `StringReplacer` is part of the public API while the fact that `viper` uses `mapstructure` under the hood is an implementation detail which we should not directly rely on.

sambetts-cisco: mapstructure tags are part of viper's public API, they call it out in their README in the unmarshalling section: https://github.com/spf13/viper#unmarshaling

chrisgacsal: Alright. Whatever.

runtime_scan/pkg/orchestrator/config.go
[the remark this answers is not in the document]
chrisgacsal: Fixed.

runtime_scan/pkg/orchestrator/config.go
chrisgacsal: Fixed.

runtime_scan/pkg/provider/aws/config.go
chrisgacsal: Updated the code to use `mapstructure` struct tags for  unmarshaling the `Config` struct, but it turned out that it is not working. It seems there is a bug related to this: https://github.com/spf13/viper/issues/1545
@sagikazarmark

sagikazarmark: Right...try setting a default for the value in Viper.

chrisgacsal: @sambetts-cisco Managed to make it work. Updated the PR.

runtime_scan/pkg/provider/aws/config.go
chrisgacsal: Fixed.

runtime_scan/pkg/provider/aws/config.go
sambetts-cisco: awesome, this looks way better from my perspective than having to maintain the environment variable constants. It also means that if we decide to use a yaml or json config, or even cobra flags we're already set up for it.

sambetts-cisco (approved):

PR #402, opened 2023-06-12
refactor: reconcile event

chrisgacsal:
## Description

* extend `ReconcileEvent` interface with `Hash` function returning `string` type which does implement the `comparable` interface required by `map` for types used as keys.
* change `common.Queue` to rely on the `ReconcileEvent` type instead of `comparable` to provide more flexibility on how  the items pushed to the queue are compared to eachother avoiding duplicated items on the queue.
* add logging of the items being pushed to the queue or popped from the queue for reconciling.

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[x] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/runtime_scan/pkg/orchestrator/common/queue.go
+++ b/runtime_scan/pkg/orchestrator/common/queue.go
@@ -22,18 +22,18 @@ import (
 	"time"
 )
 
-type Enqueuer[T comparable] interface {
+type Enqueuer[T ReconcileEvent] interface {
 	Enqueue(item T)
 	EnqueueAfter(item T, d time.Duration)
 }
 
-type Dequeuer[T comparable] interface {
+type Dequeuer[T ReconcileEvent] interface {
 	Dequeue(ctx context.Context) (T, error)
 	Done(item T)
 	RequeueAfter(item T, d time.Duration)
 }
 
-type Queue[T comparable] struct {
+type Queue[T ReconcileEvent] struct {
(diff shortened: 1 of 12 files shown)


runtime_scan/pkg/orchestrator/common/queue.go
```diff
@@ -55,25 +55,25 @@ type Queue[T comparable] struct {
...
+func NewQueue[T ReconcileEvent]() *Queue[T] {
 	return &Queue[T]{
 		itemAdded:         make(chan struct{}),
 		queue:             make([]T, 0),
-		inqueue:           map[T]struct{}{},
-		processing:        map[T]struct{}{},
-		waitingForEnqueue: map[T]struct{}{},
+		inqueue:           make(map[string]T),
+		processing:        make(map[string]T),
+		waitingForEnqueue: make(map[string]T),
```
FrimIdan, line 76: Do we need to store the item (`T`) in `inqueue`, `processing` and `waitingForEnqueue`? Or we can just use `map[string]struct{}{}`?
I didn't found where we are using the value, it seems like we just need a boolean answer.

sambetts-cisco (dismissed):

chrisgacsal:
Tested in E2E.

sambetts-cisco (approved):

PR #432, opened 2023-06-23
refactor: Orchestrator config

chrisgacsal:
## Description

* refactor `orchestrator.Config` to use `VMCLARITY_ORCHESTRATOR` prefix for environment variables
* use `viper.Unmarshal` instead of the `viper.Get*` functions
* add tests
* add environment variables to documentations (no complete)

## Type of Change

[ ] Bug Fix
[ ] New Feature  
[ ] Breaking Change  
[x] Refactor  
[x] Documentation  
[ ] Other (please describe)  

## Checklist


runtime_scan/pkg/orchestrator/config.go
```diff
@@ -33,150 +31,71 @@ import (
...
-	DiscoveryConfig           discovery.Config
-	ScanConfigWatcherConfig   scanconfigwatcher.Config
-	ScanWatcherConfig         scanwatcher.Config
-	ScanResultWatcherConfig   scanresultwatcher.Config
-	ScanResultProcessorConfig scanresultprocessor.Config
+	DiscoveryConfig           discovery.Config           `mapstructure:",squash"`
+	ScanConfigWatcherConfig   scanconfigwatcher.Config   `mapstructure:",squash"`
+	ScanWatcherConfig         scanwatcher.Config         `mapstructure:",squash"`
+	ScanResultWatcherConfig   scanresultwatcher.Config   `mapstructure:",squash"`
+	ScanResultProcessorConfig scanresultprocessor.Config `mapstructure:",squash"`
```
sambetts-cisco, line 51: I would much prefer to avoid squashing these so that the specific controllers didn't need to know about each others configurations or need to be aware of each others mapstructure prefixes. I think this can be achieved if we pass a prefix into the WithViper functions i.e. discovery.WithPrefix("discovery", v). This would potentially also allow us to reuse some parts of the configuration if we needed to at some point in the future i.e. the struct could look like:

```
DiscoveryConfigA discovery.Config `mapstructure:"discoveryA"`
DiscoveryConfigB discovery.Config `mapstructure:"discoveryB"`
```

the bind code could look like:

```
discovery.WithViper("discoveryA", v)
discovery.WithViper("discoveryB", v)
```

This way the configurations in each package have no idea they are in a larger configuration strucuture.

sambetts-cisco (changes requested):

api/models/provider.go
```diff
@@ -0,0 +1,37 @@
+// Copyright © 2023 Cisco Systems, Inc. and its affiliates.
```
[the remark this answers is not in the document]
sambetts-cisco: helper functions can live in the same directory and package as the generated code, and in fact *have* to live in the same package if we want them to be functions extending the generated types. Regenerating the code will not affect these files

api/models/provider.go
```diff
@@ -0,0 +1,37 @@
...
+// limitations under the License.
+
+package models
+
+import (
+	"fmt"
+	"strings"
+)
+
+func (p *CloudProvider) UnmarshalText(text []byte) error {
```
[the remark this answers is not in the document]
sambetts-cisco: This is a function that fulfils an interface defined by golang "encoding" https://pkg.go.dev/encoding#TextUnmarshaler. Its a type that unmarshal itself from a textual form into a go type which is what this function is doing.

runtime_scan/pkg/orchestrator/config.go
```diff
@@ -33,150 +31,71 @@ import (
...
-)
-
-const (
-	DefaultTrivyServerTimeout = 5 * time.Minute
-	DefaultGrypeServerTimeout = 2 * time.Minute
-
-	DefaultControllerStartupDelay = 15 * time.Second
+	DefaultEnvPrefix              = "VMCLARITY"
 	DefaultProviderKind           = models.AWS
+	DefaultControllerStartupDelay = 15 * time.Second
```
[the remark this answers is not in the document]
chrisgacsal: Nope, it is not. This was introduced in another PR, so I will live it as is.

runtime_scan/pkg/orchestrator/config.go
chrisgacsal: > I would much prefer to avoid squashing these so that the specific controllers didn't need to know about each others configurations or need to be aware of each others mapstructure prefixes.

They does not. Using `squash` is a workaround to prevent `viper` to prepend the inner struct name to the config key used for lookup during unmarshaling the `orchestrator.Config`.

Without using the `squash` tag the `viper` will use the `ScanWatcherConfig.scan_config_polling_interval` as key instead of the `scan_config_polling_interval`.


> This way the configurations in each package have no idea they are in a larger configuration strucuture.

They don't.

The whole configuration will need a rethink as soon as we want to support config files and cli args for defining configuration. This Pr is to make sure that the env var names are aligned and we use at least the logic for retrieving configuration from environment.

sambetts-cisco: > They does not. Using squash is a workaround to prevent viper to prepend the inner struct name to the config key used for lookup during unmarshaling the orchestrator.Config.

> Without using the squash tag the viper will use the ScanWatcherConfig.scan_config_polling_interval as key instead of the scan_config_polling_interval.

If we add a mapstructure tag with a name to the top level struct then the viper key will become: `scan_config.polling_interval`. Combined that will my suggestion of passing the prefix into the WithViper function then we can do:

```
func WithViper(v, prefix string) {
	config := "polling_interval"
	envVar := "polling_interval"
	if prefix != "" {
   		config := fmt.Sprintf("%s.%s", prefix, config)
   		envVar := fmt.Sprintf("%s_%s", prefix, envVar)
	}
	_ = BindEnv(config, envVar)
}
```

This allows the struct to be used directly as a top level thing, or within a large struct through passing a prefix.

> They don't.

They have to be a aware of it because they need to have a unique prefix, "scan_config" etc. I'm suggested that this unique prefix is passed through from the thing which has the knowledge of where it needs to be unique.

installation/aws/VmClarity.cfn
```diff
@@ -477,28 +477,26 @@ Resources:
                     ##
                     ## Orchestrator configuration
                     ##
-                    # Host for the VMClarity backend server
-                    APISERVER_HOST=apiserver
-                    # Port number for the VMClarity backend server
-                    APISERVER_PORT=8888
+                    # VMClarity API server address
+                    VMCLARITY_ORCHESTRATOR_APISERVER_ADDRESS=http://__CONTROLPLANE_HOST__:8888
```
fishkerez, line 481: Did you check that the orchestrator and the apiserver can communicate on the same host with the node public ip?

pkg/orchestrator/assetscanwatcher/delete_job_policy.go
```diff
@@ -25,21 +28,28 @@ const (
...
+	return nil
 }
 
-func GetDeleteJobPolicyType(policyType string) DeleteJobPolicyType {
-	deleteJobPolicy := DeleteJobPolicyType(policyType)
-	if !deleteJobPolicy.IsValid() {
-		log.Warnf("Invalid DeleteJobPolicy type %s. Falling back to default: %s", policyType, DeleteJobPolicyAlways)
-		deleteJobPolicy = DeleteJobPolicyAlways
-	}
+func DeleteJobPolicyOrDefault(p string) DeleteJobPolicyType {
```
fishkerez, line 50: I think this one is not used anymore

installation/aws/VmClarity.cfn
chrisgacsal: Yup, I've tested it. However the `__CONTROLPLANE_HOST__` variable is [substituted to the private IPv4](https://github.com/openclarity/vmclarity/blob/refactor-orchestrator-config/installation/aws/VmClarity.cfn#L210-L211) address of the host and not public.

It might be better to revert this config to `apiserver:8888` as all the VMClarity services are deployed in the same Docker network. Just to avoid unnecessary NAT in API <-> Orchestrator network comm.

Long term I'd probably deploy these services separately as systemd units with host networking.

pkg/orchestrator/assetscanwatcher/delete_job_policy.go
chrisgacsal: Fixed.

installation/aws/VmClarity.cfn
fishkerez: I think long term we need to understand if the orchestrator and the apiserver are deployed on the same host, and giving it address accordingly.

fishkerez: BTW, I see you left it with the "VMCLARITY_ORCHESTRATOR_APISERVER_ADDRESS=http://__CONTROLPLANE_HOST__:8888" 
Is that intentional, or you missed it?

chrisgacsal: > BTW, I see you left it with the "VMCLARITY_ORCHESTRATOR_APISERVER_ADDRESS=http://**CONTROLPLANE_HOST**:8888" Is that intentional, or you missed it?

Yeah, I wanted to discuss the course we should take here before reverting this change.

chrisgacsal: @fishkerez Updated the PR with reverting to using the `apiserver` as host in teh Orchestrator config.

fishkerez (approved):

PR #435, opened 2023-06-27
fix: Volume mount scanner host fs into scanner container

sambetts-cisco:
Containers can only be aware of device mount points which are within their containerized filesystem. This means that if a device is mounted somewhere on the host, and that location on the host is not exposed to the container, the container can not determine if its mount or not.

This commit resolves this issue by making the whole host filesystem available for the container to see by mounting it into a sub-directory called `/hostfs` within the containers filesystem.

Once the container can see the host filesystem, all volumes which are mounted by the host or by the scanner have a mount point defined in `lsblk`.

## Description

Please provide a meaningful description of what this change will do, or is for. Bonus points for including links to
related issues, other PRs, or technical references.

Note that by _not_ including a description, you are asking reviewers to do extra work to understand the context of this
change, which may lead to your PR taking much longer to review, or result in it not being reviewed at all.

## Type of Change

[X] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/runtime_scan/pkg/provider/cloudinit/cloud-init.tmpl.yaml
+++ b/runtime_scan/pkg/provider/cloudinit/cloud-init.tmpl.yaml
@@ -21,6 +21,7 @@ write_files:
       ExecStartPre=mkdir -p /var/opt/vmclarity
       ExecStartPre=docker pull {{ .ScannerImage }}
       ExecStart=docker run --rm --name %n --privileged \
+          -v /:/hostfs \
           -v /dev:/dev \
           -v /opt/vmclarity:/opt/vmclarity \
           -v /run:/run \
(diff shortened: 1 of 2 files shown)


runtime_scan/pkg/provider/cloudinit/cloud-init.tmpl.yaml
```diff
@@ -21,6 +21,7 @@ write_files:
       ExecStartPre=mkdir -p /var/opt/vmclarity
       ExecStartPre=docker pull {{ .ScannerImage }}
       ExecStart=docker run --rm --name %n --privileged \
+          -v /:/hostfs \
```
fishkerez, line 24: do we still need to mount /run and /dev explicitly with that change?

sambetts-cisco: yeah we need to keep those mounts, those directories are being mounted to the container's /run and /dev to allow the containers tools to operate against them. This change is mounting the rest of the root fs into /hostfs which doesn't affect the containers operation at all, but it just makes the container aware of all the directories

chrisgacsal (approved):
LGTM! :shipit:

PR #438, opened 2023-06-28
feat: Separate asset discovery from scanning

sambetts-cisco:
## Description

This commit removes the concept of scopes and instead replaces it with periodic discovery of assets. These assets will be created in the system and a ScanConfig can be configured with an ODATA $filter expression to choose the assets to scan.

This increases user visibility into the assets available to scan by VMClarity, it also gives the system an opportunity to estimate how many items a ScanConfig might scan based on a particular query.

It also opens the door for additional features such as the ability to scan a specific asset which has never been scanned before.

Assets are tracked with lastSeen timestamp which indicates when it was last seen by discovery. When they are not found in a discovery round they are marked as terminated. Terminated assets will not be included in scans even if they match the $filter configured by the user for the scan.

## Type of Change

[ ] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


api/openapi.yaml
```diff
@@ -1612,6 +1411,12 @@ components:
           # on the client side.
           #
           # readOnly: true
+        terminated:
```
akpsgit, line 1414: this is for all assets (e.g. also offline assets like directories). Maybe we should rename it (also with a comment), or move into the asset info?

runtime_scan/pkg/orchestrator/discovery/scope.go
```diff
@@ -44,17 +48,12 @@ func New(config Config) *ScopeDiscoverer {
 func (sd *ScopeDiscoverer) Start(ctx context.Context) {
```
akpsgit, line 48: shouldn't we rename the objects and the filename to asset discovery/er?

runtime_scan/pkg/orchestrator/discovery/scope.go
```diff
@@ -65,3 +64,74 @@ func (sd *ScopeDiscoverer) Start(ctx context.Context) {
 		}
 	}()
 }
+
+func (sd *ScopeDiscoverer) DiscoverAndCreateAssets(ctx context.Context) error {
+	discoveryTime := time.Now()
+
+	assetTypes, err := sd.providerClient.DiscoverTargets(ctx)
```
akpsgit, line 71: maybe better to do it paginated not to blowup the memory when there are many assets (separate task)?

api/openapi.yaml
sambetts-cisco: Yes this is for all assets, it not specific to VMs, any needs to be marked as removed at somepoint, i.e. the directory gets deleted, the container is rm'd, the container image is removed from the registry. Best word I could think of was "terminated" to represent this, perhaps "deleted" would be better.

runtime_scan/pkg/orchestrator/discovery/scope.go
```diff
@@ -65,3 +64,74 @@ func (sd *ScopeDiscoverer) Start(ctx context.Context) {
...
+			LastSeen:   &discoveryTime,
+		}
+		_, err := sd.backendClient.PostTarget(ctx, targetData)
+		if err == nil {
+			continue
+		}
+
+		var conflictError backendclient.TargetConflictError
+		if !errors.As(err, &conflictError) {
+			// If there is an error, and its not a conflict telling
```
akpsgit, line 89: typos the IDE alerted on.
```
			// If there is an error, and it's not a conflict telling
			// us that the target already exists, then we need to
			// keep track of it and log it as a failure to
			// complete discovery. We don't fail instantly here
			// because discovering the assets is a heavy operation,
			// so we want to give the best chance to create all the
			// assets in the DB before failing.
```

api/openapi.yaml
akpsgit: nm :)

runtime_scan/pkg/orchestrator/discovery/scope.go
sambetts-cisco: thats a great point, the AWS DiscoverAssets flow paginates the AWS call but it still results in all the models.AssetInfo's in memory. I should probably change that into a channel so the pagination is passed through.

runtime_scan/pkg/orchestrator/discovery/scope.go
```diff
@@ -65,3 +64,74 @@ func (sd *ScopeDiscoverer) Start(ctx context.Context) {
...
+
+	// Find all assets which were not updated or created by this discovery
+	// run by comparing their lastSeen time to this discovery's time stamp
+	//
+	// TODO(sambetts) when we add multiple providers/standalone support we
+	// need to filter these assets by provider so that we don't find assets
+	// which don't belong to us. We need to give the provider some kind of
+	// identity in this case.
+	targetResp, err := sd.backendClient.GetTargets(ctx, models.GetTargetsParams{
+		Filter: utils.PointerTo(fmt.Sprintf("lastSeen eq null or lastSeen lt %s", discoveryTime.Format(time.RFC3339))),
```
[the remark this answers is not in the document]
sambetts-cisco: Added not terminated to the query. Good point on marking it terminated if we fail to patch, perhaps I should skip the terminated processing if there are errors during the first part because we don't know if we've got completely accurate data?

The other thing I was thinking was to have the discovery be able to unset terminated, so if an asset disappears, but then reappears later, we can unset terminated effectively re-instating the asset. In that case if there was an error patching it, it would get marked terminated, but then the next discovery would un-terminate it?

runtime_scan/pkg/provider/aws/client.go
```diff
@@ -69,97 +69,28 @@ func (c Client) Kind() models.CloudProvider {
...
-	var filters []ec2types.Filter
-
-	awsScanScope, err := scanScope.AsAwsScanScope()
+func (c *Client) DiscoverTargets(ctx context.Context) ([]models.TargetType, error) {
+	regions, err := c.ListAllRegions(ctx)
 	if err != nil {
-		return nil, FatalError{
-			Err: fmt.Errorf("failed to cast ScanScopeType to AwsAccountScope: %w", err),
-		}
+		return nil, fmt.Errorf("failed to get regions: %w", err)
```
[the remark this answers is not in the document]
sambetts-cisco: not sure I can skip this error, DescribeRegions is a single API call which gets all the regions that are available from AWS. We could definitely change the logic below so that we process all the regions one after the other along with the channel idea from above that way we can show the error at the end.

runtime_scan/pkg/orchestrator/discovery/scope.go
```diff
@@ -65,3 +64,74 @@ func (sd *ScopeDiscoverer) Start(ctx context.Context) {
...
+		var conflictError backendclient.TargetConflictError
+		if !errors.As(err, &conflictError) {
+			// If there is an error, and its not a conflict telling
+			// us that the target already exists, then we need to
+			// keept track of it and log it as a failure to
+			// complete discovery. We don't fail instantly here
+			// because discovering the assets is a heavy operation
+			// so we want to give the best chance to create all the
+			// assets in the DB before failing.
+			errs = append(errs, fmt.Errorf("failed to post target: %v", err))
```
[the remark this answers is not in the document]
sambetts-cisco: The things that make the target unique are specific to each target type, I want this code to be agnostic to the target types being returned, otherwise we'll have to add code here every time we add a new target type. Perhaps we can put some kind of helper String() like interface on the AssetInfo type which prints that data out so then it lives with the type.

runtime_scan/pkg/provider/aws/client.go
```diff
@@ -790,10 +721,14 @@ func (c *Client) RemoveTargetScan(ctx context.Context, config *provider.ScanJobC
 func (c *Client) GetInstances(ctx context.Context, filters []ec2types.Filter, excludeTags []models.Tag, regionID string) ([]Instance, error) {
```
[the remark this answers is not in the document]
sambetts-cisco: I can remove it, I left it in case we needed it again in the future WDYT?

ui/src/layout/Assets/AssetsTable.js
```diff
@@ -43,6 +43,18 @@ const AssetsTable = () => {
...
             accessor: "assetInfo.location"
         },
+        {
+            Header: "LastSeen",
+            id: "lastSeen",
+            sortIds: ["lastSeen"],
+            accessor: original => formatDate(original.lastSeen)
+        },
+        {
+            Header: "Terminated",
```
akpsgit, line 53: On a fresh run it's not clear what are the values types. I think that better to state "Terminated On"

runtime_scan/pkg/provider/aws/client.go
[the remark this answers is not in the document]
fishkerez: If we don't use it I think we should not leave it. We can add it later if we will use it

runtime_scan/pkg/orchestrator/discovery/scope.go
sambetts-cisco: https://github.com/openclarity/vmclarity/issues/445

runtime_scan/pkg/orchestrator/discovery/scope.go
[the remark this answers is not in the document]
sambetts-cisco: https://github.com/openclarity/vmclarity/issues/446

runtime_scan/pkg/orchestrator/discovery/discoverer.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func New(config Config) *Discoverer {
+	return &Discoverer{
+		backendClient:  config.Backend,
+		providerClient: config.Provider,
+	}
+}
+
+func (sd *Discoverer) Start(ctx context.Context) {
```
akpsgit, line 48: small thing, sd -> d?

runtime_scan/pkg/provider/aws/client.go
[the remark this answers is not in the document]
akpsgit: sorry, the comment in a wrong place. Right, I meant to try to do it best effort and not return on getVMInfoFromInstance(), GetInstances()

akpsgit (approved):

api/openapi.yaml
```diff
@@ -1237,7 +1200,13 @@ components:
         scanFamiliesConfig:
           $ref: '#/components/schemas/ScanFamiliesConfig'
         scope:
-          $ref: '#/components/schemas/ScanScopeType'
+          description: |
```
fishkerez, line 1203: We should change the examples of scanConfig.json wherever we have them in the project.

PR #440, opened 2023-06-30
fix: Fix issues with postgres DB support

sambetts-cisco:
## Description

* Removes any timestamp casting from the indexes, these aren't allowed to be part of an expression index in postgres.

* Ensures that null values are handled correctly by JSONB_EACH_ARRAY. The postgres function expect the data must be an array, so we need to detect any other types in the SQL statement and switch it to empty array.

## Type of Change

[X] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/backend/pkg/database/gorm/database.go
+++ b/backend/pkg/database/gorm/database.go
@@ -132,8 +132,8 @@ func initDataBase(config types.DBConfig) (*gorm.DB, error) {
 	//	SELECT COUNT(*) FROM findings WHERE ((findings.Data->'$.findingInfo.objectType' = JSON_QUOTE('Vulnerability') AND datetime(findings.Data->>'$.foundOn') <= datetime('2023-06-11T14:24:28Z')) AND (findings.Data->'$.invalidatedOn' is NULL OR datetime(findings.Data->>'$.invalidatedOn') > datetime('2023-06-11T14:24:28Z')))
 	idb = db.Exec(fmt.Sprintf("CREATE INDEX IF NOT EXISTS findings_by_type_and_foundOn_and_invalidatedOn_idx ON findings((%s), (%s), (%s), (%s))",
 		SQLVariant.JSONExtract("Data", "$.findingInfo.objectType"),
-		SQLVariant.CastToDateTime(SQLVariant.JSONExtractText("Data", "$.foundOn")),
-		SQLVariant.CastToDateTime(SQLVariant.JSONExtractText("Data", "$.invalidatedOn")),
+		SQLVariant.JSONExtractText("Data", "$.foundOn"),
+		SQLVariant.JSONExtractText("Data", "$.invalidatedOn"),
 		SQLVariant.JSONExtract("Data", "$.invalidatedOn")))
 	if idb.Error != nil {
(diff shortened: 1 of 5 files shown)


backend/pkg/database/gorm/database.go
```diff
@@ -132,8 +132,8 @@ func initDataBase(config types.DBConfig) (*gorm.DB, error) {
 	//	SELECT COUNT(*) FROM findings WHERE ((findings.Data->'$.findingInfo.objectType' = JSON_QUOTE('Vulnerability') AND datetime(findings.Data->>'$.foundOn') <= datetime('2023-06-11T14:24:28Z')) AND (findings.Data->'$.invalidatedOn' is NULL OR datetime(findings.Data->>'$.invalidatedOn') > datetime('2023-06-11T14:24:28Z')))
 	idb = db.Exec(fmt.Sprintf("CREATE INDEX IF NOT EXISTS findings_by_type_and_foundOn_and_invalidatedOn_idx ON findings((%s), (%s), (%s), (%s))",
 		SQLVariant.JSONExtract("Data", "$.findingInfo.objectType"),
-		SQLVariant.CastToDateTime(SQLVariant.JSONExtractText("Data", "$.foundOn")),
-		SQLVariant.CastToDateTime(SQLVariant.JSONExtractText("Data", "$.invalidatedOn")),
+		SQLVariant.JSONExtractText("Data", "$.foundOn"),
+		SQLVariant.JSONExtractText("Data", "$.invalidatedOn"),
```
FrimIdan, line 136: I think that while not doing date-time casting the widget query will return to be slow - did you checked it?

sambetts-cisco: I need to double check, I'm going to deploy a version of this fix with SQLite enabled and test to see if the index gets used

sambetts-cisco: Double checked, and SQLite is using the index and it appears to be performing well without the datetime cast in the index itself.

This index might actually be better from a re-use perspective if we want to do a different comparison against the date fields because the index isn't specific about it needing to be a date too.

backend/pkg/database/odatasql/jsonsql/postgres.go
```diff
@@ -37,7 +37,7 @@ func (postgres) CastToDateTime(strTime string) string {
 }
 
 func (postgres) JSONEach(source string) string {
-	return fmt.Sprintf("JSONB_ARRAY_ELEMENTS((%s))", source)
+	return fmt.Sprintf("JSONB_ARRAY_ELEMENTS(CASE JSONB_TYPEOF(%s) WHEN 'array' THEN (%s) ELSE '[]' END)", source, source)
```
akpsgit, line 40: From the PR: 
Ensures that null values are handled correctly by JSONB_EACH_ARRAY. The postgres function expect the data must be an array, so we need to detect any other types in the SQL statement and switch it to empty string.

Maybe worth adding this comment? is it "to empty array"?

akpsgit (approved):

PR #443, opened 2023-07-04
feat: Remodel so that scan objects can be used separately

sambetts-cisco:
## Description

This commit remodels ScanConfigs, Scans and AssetScans so that Scans and AssetScans can be created independently from each other. This opens the door for a number of different use cases some examples are:

- creating a single Scan without a ScanConfig if you just want to scan once now.
- creating an AssetScan for a specific asset without a Scan to scan a specific asset now.
- creating an AssetScan with results for a asset without going through the orchestrator at all, for example when reporting results from a scan in CI/CD

## Type of Change

[ ] Bug Fix  
[X] New Feature  
[X] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


backend/pkg/database/gorm/database.go
```diff
@@ -121,9 +121,9 @@ func initDataBase(config types.DBConfig) (*gorm.DB, error) {
 	// The UI needs to find all the findings for a specific finding type
 	// and the asset scan processor needs to filter that list by a
```
fishkerez, line 122: should the comment also be updated? (filter that list by a specific asset scan)

pbalogh-sa (commented):
LGTM

backend/pkg/database/gorm/asset_scan.go
```diff
@@ -294,8 +284,14 @@ func (s *AssetScansTableHandler) UpdateAssetScan(assetScan models.AssetScan, par
 }
 
 func (s *AssetScansTableHandler) checkUniqueness(assetScan models.AssetScan) (models.AssetScan, error) {
+	// We only check unique if scan is set so return early if its not set.
```
akpsgit, line 287: early if it's not set

backend/pkg/database/gorm/asset_scan.go
```diff
@@ -294,8 +284,14 @@ func (s *AssetScansTableHandler) UpdateAssetScan(assetScan models.AssetScan, par
...
+	if assetScan.Scan == nil || assetScan.Scan.Id == "" {
+		return models.AssetScan{}, nil
+	}
+
+	// If Scan is set we need to check if there is another asset scan with
```
akpsgit, line 292: with with

ui/src/layout/Findings/utils.js
```diff
@@ -41,8 +29,8 @@ export const FindingsDetailsCommonFields = ({foundOn, invalidatedOn}) => (
         </TitleValueDisplayRow>
         {!!invalidatedOn &&
             <TitleValueDisplayRow>
-                <TitleValueDisplay title="Replaced by a newer scan on">{formatDate(invalidatedOn)}</TitleValueDisplay>
+                <TitleValueDisplay title="Invalidated on">{formatDate(invalidatedOn)}</TitleValueDisplay>
```
akpsgit, line 32: was it wrong? not sure what "invalidated" means to the user

ui/src/layout/Scans/ScanConfigWizardModal/index.js
```diff
@@ -12,16 +12,20 @@ import './scan-config-wizard-modal.scss';
...
 const ScanConfigWizardModal = ({initialData, onClose, onSubmitSuccess}) => {
-    const {id, name, scope, scanFamiliesConfig, scheduled, maxParallelScanners, scannerInstanceCreationConfig} = initialData || {};
+    const {id, name, scanTemplate, scheduled} = initialData || {};
+    const {scope, maxParallelScanners, assetScanTemplate} = scanTemplate || {};
     const {operationTime, cronLine} = scheduled || {};
+
+    const {scanFamiliesConfig, scannerInstanceCreationConfig} = assetScanTemplate || {}
     const {useSpotInstances} = scannerInstanceCreationConfig || {};
+
+    const {allRegions, regions, shouldScanStoppedInstances, instanceTagSelector, instanceTagExclusion} = scope || {};
```
akpsgit, line 22: unused according to IDE

ui/src/layout/Findings/Vulnerabilities/VulnerabilityDetails.js
```diff
@@ -31,12 +30,6 @@ const DetailsContent = ({data}) => {
                     title: "Asset details",
                     path: VULNERABILITY_DETAILS_PATHS.ASSET_DETAILS,
                     component: () => <AssetDetails assetData={asset} withAssetLink />
-                },
-                {
```
akpsgit, line 35: (General comment) I see that we now detach the finding from the scan. Don't know if worth removing it when it is associated with a scan. Or even better, should we associate it with the asset scan?

ui/src/layout/Findings/utils.js
sambetts-cisco: "Replaced by a new scan" to me means there was a scan which caused the finding to be invalidated, this might not be true if we update the finding graph in the future dynamically from the vulnerability databases etc.

"Replaced" also implies something was created to replace it, which also might not be true, a finding might be marked invalidated because it doesn't exist any more.

pbalogh-sa (approved):

akpsgit (approved):

PR #444, opened 2023-07-05
feat: Add nullable go type and use it for the fields in the API

ghost:
## Description

This introduces a new generic go type that we can use for all the types
in the API which need to be optional and nullable. This allows us to
distingish between unset, set but null, and set to a value. This is
important because in JSON PATCH these three states have different
meanings:

- unset: do nothing
- set to null: delete the existing data
- set to a value: replace the existing data

Traditionally in go there is no way to tell the difference between unset
and set to null. For example if you use a pointer type then json
unmarshal converts both to nil.ll.

## Type of Change

[X] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


ramizpolic:
Resolved by https://github.com/oapi-codegen/nullable

ramizpolic:
Closing due to unification

PR #449, opened 2023-07-06
feat: Add GCP installer using GCP Deployment Manager

sambetts-cisco:
## Description

This commit adds GCP Cloud Deployment Manager as a new method for installing VMClarity.

## Type of Change

[ ] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,16 @@ For a detailed installation guide, please see [AWS](installation/aws/README.md).
 2. Fill out the required fields in the wizard
 3. Once deployed, copy the VMClarity SSH address from the Outputs tab
 
+### GCP
+
+1. Change directory to `installation/gcp/dm`
+2. Copy `vmclarity-config.example.yaml` to `vmclarity-config.yaml`, update with required values.
+3. Deploy vmclarity using GCP deployment manager
+   ```
+   gcloud deployment-manager deployments create <vmclarity deployment name> --config vmclarity-config.yaml
+   ```
+4. Once deployed, copy the VMClarity SSH IP address from the CLI output.
+
 ## Access VMClarity UI
 1. Open an SSH tunnel to VMClarity server
     ```
--- a/installation/gcp/dm/.flake8
+++ b/installation/gcp/dm/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 90
--- a/installation/gcp/dm/.gitignore
+++ b/installation/gcp/dm/.gitignore
@@ -1,0 +1,1 @@
+vmclarity-config.yaml
--- a/installation/gcp/dm/Makefile
+++ b/installation/gcp/dm/Makefile
@@ -1,0 +1,5 @@
+lint:
+	flake8 *.py components/*.py
+
+format:
+	black *.py components/*.py
--- a/installation/gcp/dm/README.md
+++ b/installation/gcp/dm/README.md
@@ -1,0 +1,30 @@
+# Installing VMClarity on GCP with Deployment Manager
+
+1. Copy example configuration file to a new config
+
+   ```
+   cp vmclarity-config.example.yaml vmclarity-config.yaml
+   ```
+
+2. Edit configuration to add required fields.
+   Check vmclarity.py.schema for other optional parameters.
+
+3. Create a deployment using the gcloud CLI
+
+   ```
+   gcloud deployment-manager deployments create <vmclarity deployment name> --config vmclarity-config.yaml
+   ```
+
+4. Copy the VMClarity IP address from the output
+
+5. SSH into the VMClarity server and open a tunnel to the UI
+
+   ```
+   ssh -L 8888:localhost:8888 vmclarity@<VMClarity server IP address>
+   ```
+
+6. To update the VMClarity configuration, modify the vmclarity-config.yaml, then update the deployment:
+
+   ```
+   gcloud deployment-manager deployments update <vmclarity deployment name> --config vmclarity-config.yaml
+   ```
--- a/installation/gcp/dm/components/cloud-router.py
+++ b/installation/gcp/dm/components/cloud-router.py
@@ -1,0 +1,41 @@
+# Copyright © 2023 Cisco Systems, Inc. and its affiliates.
+# All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Creates the cloud NAT"""
+
+
+def GenerateConfig(context):
+    """Creates the cloud NAT"""
+
+    vmclarity_scanner_nat = {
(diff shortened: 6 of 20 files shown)


installation/gcp/dm/firewall-template.py
```diff
@@ -0,0 +1,35 @@
...
+    
+    resources = [
+        {
+            'name': f"allow-ssh-to-vmclarity-server-{context.env['deployment']}",
+            'type': 'compute.v1.firewall',
+            'properties': {
+                'network': '$(ref.' + context.properties['network'] + '.selfLink)',
+                'sourceRanges': ['0.0.0.0/0'],
+                'allowed': [{
+                    'IPProtocol': 'TCP',
```
fishkerez, line 29: Should you enable also http/https access to the server?

installation/gcp/dm/vmclarity-server.py
```diff
@@ -0,0 +1,70 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Creates the virtual machine with environment variables and startup script."""
+
+import vmclarity_install_script
+
+COMPUTE_URL_BASE = 'https://www.googleapis.com/compute/v1/'
+
+IMAGE_SELFLINK = "projects/ubuntu-os-cloud/global/images/ubuntu-2204-jammy-v20230630"
```
fishkerez, line 21: this image does not have cloud-init installed by default, so you might need to install it as part of the startup-script, or alternatively use the container optimized image

fishkerez:
Did not see any scanners network/firewall rules, the way we have in the rest of the providers (private network vs public and so on..). Is that on purpose?

installation/gcp/dm/firewall-template.py
sambetts-cisco: Not until we have RBAC, until then its the same as all the other providers, UI access is through the SSH tunnel

installation/gcp/dm/vmclarity-server.py
fishkerez: Did you check it? I remember checking it and it did not have

sambetts-cisco: Yeah I checked, without installing anything you can run `sudo journalctl -u cloud-init -u cloud-final` and see cloud-init running.

installation/gcp/dm/vmclarity-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+- path: vmclarity.py
+
+resources:
+- name: vmclarity
+  type: vmclarity.py
+  properties:
+    region: us-central1
+    zone: us-central1-f
+    machineType: e2-standard-2
+    sshPublicKey: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCyqIoOL9CcSxcNK/8YrDvgO2mfzKNWidKk6sYBMTEIVHqz3e1L/5kKi8sIh5ivSD+OiLlnmxfdc39l+5Ov7w9gODn/f57a85t8cPCS6Vths6wumwf4j5mYmwOS9ujmu2ZPJMc7P0i0CiTj12jniCglRm5C0v7W6VudBoRxpxxrz+n4LQyLiUicC6Qp/E7lyJFw1VYDLLSeakeapM+5M3dus00FLlWOOp7VvcE86K8h/31/d5QmqaLJHRX3oN9kiSXGXwijOgv1QYHO01/XFqZGa58yVGaSzT98o8gJDE4Bs5D8gpsepYUh6i18L1VH575F0v1MVG7qn5MmHp9nTh0T sambetts@SAMBETTS-M-H09X
```
pbalogh-sa, line 11: I think it would be better to commit vmclarity-config.yaml as an example using a comment instead of a specified ssh pubkey.

installation/gcp/dm/firewall-template.py
```diff
@@ -0,0 +1,50 @@
...
+            'name': f"allow-scanner-to-vmclarity-server-{context.env['deployment']}",
+            'type': 'compute.v1.firewall',
+            'properties': {
+                'network': '$(ref.' + context.properties['network'] + '.selfLink)',
+                'direction': 'INGRESS',
+                'sourceRanges': ['10.128.0.0/9'],
+                'targetTags': ["vmclarity-control-plane"],
+                'allowed': [{
+                    'IPProtocol': 'TCP',
+                    'ports': ["0-65535"]
```
fishkerez, line 45: Why enabling all ports instead the specific ports we listen on (8888)?

installation/gcp/dm/vmclarity-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+    backendContainerImage: docker.io/tehsmash/vmclarity-backend:gcp_support
+    scannerContainerImage: docker.io/tehsmash/vmclarity-cli:gcp_support
```
pbalogh-sa, line 13: same

installation/gcp/dm/network-template.py
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+
+"""Creates the network."""
+
+
+def GenerateConfig(context):
+  """Creates the network with environment variables."""
+
+  resources = [{
+      'name': context.env['name'],
```
fishkerez, line 22: Is that mean that the "cloud NAT" and the network will have the same name? (both using context.env['name'])

installation/gcp/dm/roles.py
```diff
@@ -0,0 +1,91 @@
...
+        {
+            'name': f"vmclarity-discoverer-snapshotter-{deployment_name_hash}",
+            'type': 'gcp-types/iam-v1:projects.roles',
+            'properties': {
+                'parent': f"projects/{context.env['project']}",
+                'roleId': f"vmclarity_discoverer_snapshotter_{deployment_name_hash}",
+                'role': {
+                    'title': f"VMClarity {context.env['deployment']} Discoverer Snapshotter",
+                    'description': (f"Role to allow vmclarity {context.env['deployment']} to discover "
+                                    f"and snapshot instances with the project"),
```
fishkerez, line 20: nit: snapshot instances **within** the project

installation/gcp/dm/vmclarity-config.yaml
```diff
@@ -0,0 +1,16 @@
+imports:
```
fishkerez, line 1: Do we need this file? Why are the values of  sshPublicKey and the images are hard coded?

pbalogh-sa (approved):

installation/gcp/dm/firewall-template.py
sambetts-cisco: This specifically allows access between the scanner and the control plane, I opted for the easier choice to allow all ports, then if we add any other components like grype-server etc the installer doesn't need changing. This is a little different to the other installers but we can align them if we want.

installation/gcp/dm/network-template.py
sambetts-cisco: context.env['name'] is the name passed into the template from the config/wrapper template. For example the wrapper vmclarity template uses this network template like:

```
resources [
   { 
       name: vmclarity-{context.env['deployment']}-network
       type: network-template.py
    }
 ]
 ```
 
context.env['name'] becomes `vmclarity-{context.env['deployment']}-network` in this case.

installation/gcp/dm/vmclarity-config.yaml
sambetts-cisco: I've removed this one, and left a vmclarity-config.example.yaml instead, and updated the README to describe how to use it.

fishkerez (approved):

PR #454, opened 2023-07-10
fix: Fix data race checking the queue length in dequeue

sambetts-cisco:
## Description

If we are working with q.queue we must hold the lock to make sure there is no race conditions. During Dequeue we weren't holding the lock when we checked the length of queue.

After this commit we hold the lock to check the length of q.queue. If the queue is empty we unlock the mutex to allow an item to be added to the queue while we wait for a itemAdded notification. The lock is then regained to dequeue the item that was added.

If the queue isn't empty we continue to hold the lock and move straight to dequeue the item defering the unlock until the function completes.

## Type of Change

[X] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/runtime_scan/pkg/orchestrator/common/queue.go
+++ b/runtime_scan/pkg/orchestrator/common/queue.go
@@ -82,12 +82,19 @@ func NewQueue[T ReconcileEvent]() *Queue[T] {
 // "Done" is called with the dequeued item to acknowledge its processing is
 // completed.
 func (q *Queue[T]) Dequeue(ctx context.Context) (T, error) {
+	// Grab the lock so that we can check the length of q.queue safely.
+	q.l.Lock()
 	if len(q.queue) == 0 {
+		// Unlock while we wait for an item to be added to the queue
+		q.l.Unlock()
+
 		// If the queue is empty, block waiting for the itemAdded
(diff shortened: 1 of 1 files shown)


runtime_scan/pkg/orchestrator/common/queue.go
```diff
@@ -100,8 +107,6 @@ func (q *Queue[T]) Dequeue(ctx context.Context) (T, error) {
 		default:
 		}
 	}
-
-	q.l.Lock()
 	defer q.l.Unlock()
```
ramizpolic, line 110: This can be moved on line 98

sambetts-cisco: This defer applies to two lock lines, if len(q.queue) == 0 then it defers the unlock of the Lock @97, if len(q.queue) > 0 then it defers the Lock @86 so I can't move this up otherwise we won't unlock the case when there are already items in the queue.

fishkerez (approved):

akpsgit (approved):

ramizpolic (approved):

PR #460, opened 2023-07-11
Extending asset vm info with root volume details of encryption and size (AWS and Azure support)

akpsgit:
Extending asset vm info with root volume details of encryption and size (AWS and Azure support)


api/openapi.yaml
```diff
@@ -1516,6 +1516,10 @@ components:
         launchTime:
           type: string
           format: date-time
+        rootVolumeSizeGB:
+          type: integer
+        rootVolumeEncrypted:
+          type: boolean
```
sambetts-cisco, line 1522: What do you think about nesting this info in a new object to avoid the prefixing? like:

```
rootVolume:
   sizeGB:
      type: integer
   encrypted:
      type: boolean
 ```
 ? 
 
 Do you think we should pull all the disk info off the instance? Like:
 
 ```
 disks:
    type: array
    items:
       type: object
       properties:
          rootVolume:
             description: This disk is the root volume, only one disk is allowed to have this marked true.
             type: boolean
          sizeGB:
              type: integer
          encrypted:
              type: boolean
 ```

File: runtime_scan/pkg/provider/aws/client.go [hunk 2]
```diff
@@ -797,6 +810,54 @@ func (c *Client) getInstancesFromDescribeInstancesOutput(ctx context.Context, re
...
+type rootVolumeInfo struct {
+	DeviceName string
+	SizeGB     int32
+	Encrypted  bool
+}
+
+func getRootVolumeInfo(ctx context.Context, client *ec2.Client, i ec2types.Instance, region string) (*rootVolumeInfo, error) {
+	logger := log.GetLoggerFromContextOrDiscard(ctx)
+	for _, mapping := range i.BlockDeviceMappings {
+		if getStringPointerValOrEmpty(mapping.DeviceName) == getStringPointerValOrEmpty(i.RootDeviceName) {
```
fishkerez, line 822: To avoid weird cases like RootDeviceName is not set or empty, I would check
```
 if i.RootDeviceName == nil {
  return nil, fmt.Errorf("rootDevice not set")
}

```
at the beginning of the function

runtime_scan/pkg/provider/azure/client.go
```diff
@@ -222,19 +223,48 @@ func processVirtualMachineListIntoAssetTypes(vmList armcompute.VirtualMachineLis
 	return ret, nil
 }
 
-func getVMInfoFromVirtualMachine(vm *armcompute.VirtualMachine) (models.AssetType, error) {
+type rootVolumeInfo struct {
```
fishkerez, line 226: Since you define this both here and in aws, can we have a general type "rootVolume" defined in types?

runtime_scan/pkg/provider/azure/client.go
```diff
@@ -222,19 +223,48 @@ func processVirtualMachineListIntoAssetTypes(vmList armcompute.VirtualMachineLis
...
+	SizeGB    int32
+	Encrypted bool
+}
+
+func (c *Client) getRootVolumeInfo(ctx context.Context, vm *armcompute.VirtualMachine) *rootVolumeInfo {
+	logger := log.GetLoggerFromContextOrDiscard(ctx)
+	ret := &rootVolumeInfo{
+		SizeGB:    *vm.Properties.StorageProfile.OSDisk.DiskSizeGB,
```
fishkerez, line 234: maybe use getInt32PointerValOrEmpty ?

runtime_scan/pkg/provider/azure/client.go
```diff
@@ -222,19 +223,48 @@ func processVirtualMachineListIntoAssetTypes(vmList armcompute.VirtualMachineLis
...
+		logger.Warnf("Failed to parse disk ID. DiskID=%v: %v", *vm.Properties.StorageProfile.OSDisk.ManagedDisk.ID, err)
+		return ret
+	}
+	osDisk, err := c.disksClient.Get(context.TODO(), osDiskID.ResourceGroupName, osDiskID.Name, nil)
+	if err != nil {
+		logger.Warnf("Failed to get OS disk. DiskID=%v: %v", *vm.Properties.StorageProfile.OSDisk.ManagedDisk.ID, err)
+		return ret
+	}
+	ret.Encrypted = isEncrypted(osDisk)
+	ret.SizeGB = *osDisk.Disk.Properties.DiskSizeGB
```
fishkerez, line 248: use getInt32PointerValOrEmpty

runtime_scan/pkg/provider/azure/client.go
```diff
@@ -222,19 +223,48 @@ func processVirtualMachineListIntoAssetTypes(vmList armcompute.VirtualMachineLis
...
+	}
+	osDiskID, err := arm.ParseResourceID(*vm.Properties.StorageProfile.OSDisk.ManagedDisk.ID)
+	if err != nil {
+		logger.Warnf("Failed to parse disk ID. DiskID=%v: %v", *vm.Properties.StorageProfile.OSDisk.ManagedDisk.ID, err)
+		return ret
+	}
+	osDisk, err := c.disksClient.Get(context.TODO(), osDiskID.ResourceGroupName, osDiskID.Name, nil)
+	if err != nil {
+		logger.Warnf("Failed to get OS disk. DiskID=%v: %v", *vm.Properties.StorageProfile.OSDisk.ManagedDisk.ID, err)
+		return ret
```
sambetts-cisco, line 245: I like the idea of not failing to discover the asset if we can't get its root volume info, however I'm not sure I like that we're setting Encrypted to false in this case. Perhaps encrypted should be an enum with three states, yes, no, and unknown?

sambetts-cisco (changes requested):

runtime_scan/pkg/provider/azure/client.go
akpsgit: we can but it's just an auxiliary struct and azure doesn't have a device name. WDYT?

sambetts-cisco: I would avoid tying the provider implementations together with shared structs, IMO this is an implementation detail and it just happens to look/be named the same because the same developer wrote it, but they aren't really the same.

fishkerez: Yeah but we are assuming there is a rootVolume for each provider, that has a size and encryption option, no? This is a general concept that exists also in the api on VMInfo, that is provider agnostic

sambetts-cisco: I think if @akpsgit takes my suggestion from above about making a rootVolume struct in the OAS we could reuse the models.RootVolume here potentially.

api/openapi.yaml
akpsgit: I can move it to a new object, but not sure that that's the time to make it a list, i think that it will block the UI as we need a UX for that. We can do it when there will be a need

runtime_scan/pkg/provider/azure/client.go
akpsgit: I like the idea but I think that it will raise questions what unknown means and what needs to be done for it to be known.
I think that we need to better define it (e.g. store and show the error why it's unknown). Maybe better to do in a separate task?

runtime_scan/pkg/provider/aws/client.go
akpsgit: good point but this is an internal function that runs right after validateInstanceFields. I think that we should validate it there. WDYT?
Do you think that better not to know about this asset in discovery (as validation failed) or only have no root volume info of (validate like you suggested)

runtime_scan/pkg/provider/azure/client.go
akpsgit: the problem is that aws has a root device name but azure doesn't, that's why i didn't make it common. Not sure about using an API struct here, I think that each provided will have it's own fields and maybe not worth make it common. If you think that it's better to use the API, i will remove the root device name and use it from the instance

runtime_scan/pkg/provider/azure/client.go
sambetts-cisco: I'm just concern that people might try to run a scan by filtering on whether the disk is encrypted or not, if we set it to "false" if we failed to determine it then they might get an error like "Failed to snapshot disk because its encrypted", but then our API says that the root disk is encrypted.

I'd rather be truthful and say and we were unable to determine it, perhaps unknown is the wrong word, maybe "undetermined"?

runtime_scan/pkg/provider/azure/client.go
sambetts-cisco: yeah probably better to leave getVMInfoFromVirtualMachine to convert it to the API objects 👍

api/openapi.yaml
sambetts-cisco: 👍

runtime_scan/pkg/provider/azure/client.go
akpsgit: Done

sambetts-cisco (approved):

sambetts-cisco (dismissed):

sambetts-cisco (dismissed):

sambetts-cisco (dismissed):

lgecse (approved):

PR #462, opened 2023-07-12
feat(docker): create docker provider

adamtagscherer:
## Description

This PR creates the docker provider which lets us to scan containers and images downloaded to the local registry.
After this PR is merged in we can move to creating the E2E tests.

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,539 @@
...
+--output /var/opt/vmclarity
+`, filepath.Base(scanConfigFilePath), config.VMClarityAddress, config.AssetScanID)
+
+	// TODO(paralta) Check if container already created
+	resp, err := c.dockerClient.ContainerCreate(
+		ctx,
+		&container.Config{
+			Image:      config.ScannerImage,
+			Entrypoint: []string{"sh", "-c", script},
+		},
```
ramizpolic, line 300: Can we have this a standard command rather than an entrypoint?
```go
&container.Config{
	Image: config.ScannerImage,
	Cmd:   []string{"/app/vmclarity-cli", "--config", fmt.Sprintf("/tmp/%s", filepath.Base(scanConfigFilePath), // add the rest
},

```

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,539 @@
...
+}
+
+func getScanConfigFileName(config *provider.ScanJobConfig) (string, error) {
+	scanName, err := getScanName(config)
+	if err != nil {
+		return "", provider.FatalError{Err: err}
+	}
+
+	// TODO (adamtagscherer): check if os.TempDir() doesn't create a new unique directory every time it's being called
+	return os.TempDir() + scanName + "_scanconfig.yaml", nil
```
ramizpolic, line 430: You can try using CreateTemp, e.g.
```go
os.CreateTemp("", fmt.Sprintf("*_%s_scanconfig.yaml", scanName))
```

cli/cmd/root.go
```diff
@@ -108,6 +108,20 @@ var rootCmd = &cobra.Command{
 				return err
 			}
 			setMountPointsForFamiliesInput(mountPoints, config)
+		} else {
+			// Check if volume already mounted
+			scanPath := "/mnt/snapshot"
```
ramizpolic, line 113: It would be good to define this as a constant somewhere

runtime_scan/pkg/orchestrator/config.go
```diff
@@ -136,10 +136,12 @@ func LoadConfig(backendHost string, backendPort int, baseURL string) (*Config, e
...
 	case strings.ToLower(string(models.Azure)):
 		providerKind = models.Azure
+	case strings.ToLower(string(models.Docker)):
+		providerKind = models.Docker
 	case strings.ToLower(string(models.AWS)):
-		fallthrough
-	default:
 		providerKind = models.AWS
+	default:
+		return nil, fmt.Errorf("unsupported provider: %s", providerKind)
```
ramizpolic, line 144: Would be good to use the same value that was used before, `providerKind = models.AWS`

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,539 @@
...
+		return nil, fmt.Errorf("failed to load provider configuration. Provider=%s: %w", models.Docker, err)
+	}
+
+	return &Client{
+		dockerClient: dockerClient,
+		config:       config,
+	}, nil
+}
+
+func (c Client) Kind() models.CloudProvider {
```
ramizpolic, line 67: Can we use pointers instead of structs for method declarations?

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+
+	for _, i := range images {
+		asset, err := c.getAssetFromImage(ctx, i.ID)
+		if err != nil {
+			errs <- provider.FatalError{
+				Err: fmt.Errorf("failed to create AssetType from ContainerImageInfo. Provider=%s: %w", models.Docker, err),
+			}
+			return
+		}
+		assets <- asset
```
ramizpolic, line 126: This should be run in a separate goroutine. I would move this into separate function, e.g. getImage) which
1. has signature `getImage(ctx context.Context, imageID string) (models.AssetType, error)`
2. returning error should be a regular error and not a provider error (this way we can control the errors in provider-specific interface methods)

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+
+func (c Client) getImages(ctx context.Context, assets chan models.AssetType, errs chan error, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	images, err := c.dockerClient.ImageList(ctx, types.ImageListOptions{})
+	if err != nil {
+		errs <- provider.FatalError{
+			Err: fmt.Errorf("failed to get images. Provider=%s: %w", models.Docker, err),
+		}
+		return
```
ramizpolic, line 115: Return a regular error and decide what to do with it from the caller scope

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return ret, nil
+}
+
+func (c Client) getImages(ctx context.Context, assets chan models.AssetType, errs chan error, wg *sync.WaitGroup) {
+	defer wg.Done()
```
ramizpolic, line 108: No need to pass down the sync.WaitGroup. We can run this method in an inline function where we have the access to wg by default, e.g.
```go
go func(){
	defer wg.Done()
	getImages(...)
}()
```

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+			}
+			return
+		}
+		assets <- asset
+	}
+}
+
+func (c Client) getContainers(ctx context.Context, assets chan models.AssetType, errs chan error, wg *sync.WaitGroup) {
+	defer wg.Done()
+
```
ramizpolic, line 132: Same notes like to getImages apply:
1. create separate single-fetch method
2. use that instead of doing everything in `getContainers`
3. get rid of provider-errors
4. remove unnecessary wg control

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+	for e := range errs {
+		if e != nil {
+			// nolint:typecheck
+			err = errors.Join(err, e)
+		}
+	}
+	if err != nil {
+		return nil, err
+	}
```
ramizpolic, line 102: The assumption that there will only be two errors will lead to a deadlock. Note that both methods `getImages` and `getContainers` write to assets and errors channels multiple times (for example, in the loop). When we have a situation in which 3 errors have been written to a channel and asset channel consumer is still waiting, this will block.

ramizpolic: I also think that there is no need to call getImages and getContainers in a spearate gorotine if these two will perform the requests concurrently to each image/container internally. Fetching the list of containers/images is a relatively minor operation compared to batch processing all of them.

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+		if err != nil {
+			_ = fmt.Errorf("failed to remove helper container. Provider=%s: %w", models.Docker, err)
+		}
+	}()
+	err = c.dockerClient.CopyToContainer(ctx, response.ID, "/data", readCloser, types.CopyToContainerOptions{})
+	if err != nil {
+		return provider.FatalError{
+			Err: fmt.Errorf("failed to copy data to container. Provider=%s: %w", models.Docker, err),
+		}
+	}
```
ramizpolic, line 257: I think this can be moved to a separate method as well, e.g. `prepareVolume` or something more descriptive?

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+
+	readCloser, err := c.export(ctx, config)
+	if err != nil {
+		return provider.FatalError{
+			Err: fmt.Errorf("failed to export target. Provider=%s: %w", models.Docker, err),
+		}
+	}
+
+	// Create an ephemeral container to populate volume with export output
+	_, err = c.dockerClient.ImagePull(ctx, "alpine", types.ImagePullOptions{})
```
ramizpolic, line 222: If we will be pulling VMClarity CLI tool, then we can also try using that image to perform bootstrapping rather than doing it with some other image. Maybe consider this at the end.

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+	}
+
+	return nil
+}
+
+func (c Client) createScanContainer(ctx context.Context, config *provider.ScanJobConfig) (string, error) {
+	// Check if scan container already created
+	scanName, err := getScanName(config)
+	if err != nil {
+		return "", provider.FatalError{Err: err}
```
ramizpolic, line 285: Return regular error

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+	if err != nil {
+		return err
+	}
+
+	if _, err := os.Stat(scanConfigFilePath); errors.Is(err, os.ErrNotExist) {
+		err = os.WriteFile(scanConfigFilePath, []byte(config.ScannerCLIConfig), 0644)
+		if err != nil {
+			return err
+		}
+	}
```
ramizpolic, line 374: Non `os.ErrNotExist` errors are not handled (unsure if it can occur, but either way should be handled too)

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+	if len(containers) == 0 {
+		return "", provider.FatalError{
+			Err: fmt.Errorf("scan container not found. Provider=%s: %w", models.Docker, err),
+		}
+	}
+	if len(containers) > 1 {
+		return "", provider.FatalError{
+			Err: fmt.Errorf("found more than one scan container. Provider=%s: %w", models.Docker, err),
+		}
+	}
```
ramizpolic, line 359: Same error cases here

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+	return &ret
+}
+
+func getScanName(config *provider.ScanJobConfig) (string, error) {
+	id, err := getAssetId(config)
+	if err != nil {
+		return "", provider.FatalError{
+			fmt.Errorf("failed to get scan name. Provider=%s: %w", models.Docker, err),
+		}
```
ramizpolic, line 464: Error handling

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+	}
+	return "scan_" + config.ScanID + "_asset_" + strings.Replace(id, ":", "_", -1), nil
+}
+
+func getScanConfigFileName(config *provider.ScanJobConfig) (string, error) {
+	scanName, err := getScanName(config)
+	if err != nil {
+		return "", provider.FatalError{Err: err}
```
ramizpolic, line 472: Error handling

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+func getAssetId(config *provider.ScanJobConfig) (string, error) {
+	objectType, err := config.AssetInfo.Discriminator()
+	if err != nil {
+		return "", provider.FatalError{
+			Err: fmt.Errorf("failed to get asset object type. Provider=%s: %w", models.Docker, err),
+		}
+	}
+
+	switch objectType {
+	case "ContainerInfo":
```
ramizpolic, line 488: I would suggest creating an internal type e.g. `type dockerAssetType string` and defining ContainerInfor and ContainerImageInfo as internal constants.

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+	scanName, err := getScanName(config)
+	if err != nil {
+		return "", provider.FatalError{Err: err}
+	}
+
+	// TODO (adamtagscherer): check if os.TempDir() doesn't create a new unique directory every time it's being called
+	return os.TempDir() + scanName + "_scanconfig.yaml", nil
+}
+
+func getAssetId(config *provider.ScanJobConfig) (string, error) {
```
ramizpolic, line 479: Errors should not be provider-specific

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,585 @@
...
+		}
+		return *containerImageInfo.Name, nil
+	default:
+		return "", provider.FatalError{
+			Err: fmt.Errorf("get asset id not implemented for current object type (%s). Provider=%s", models.Docker, objectType),
+		}
+	}
+}
+
+func (c Client) export(ctx context.Context, config *provider.ScanJobConfig) (io.ReadCloser, error) {
```
ramizpolic, line 543: Error handling

runtime_scan/pkg/provider/docker/client.go
paralta: will change this

runtime_scan/pkg/provider/docker/client.go
paralta: 👍

runtime_scan/pkg/provider/docker/client.go
paralta: agreed, will change this 😄

ramizpolic:
I have added improved concurrency model for docker asset discovery to the branch, along with some examples on how to address error management. The changes can be found on https://github.com/openclarity/vmclarity/tree/feature/docker-cleanup branch

runtime_scan/pkg/provider/docker/container.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to list containers: %w", err)
+	}
+
+	// Results will be written to assets concurrently
+	assetMu := sync.Mutex{}
+	assets := make([]models.AssetType, 0, len(containers))
+
+	// Process each container in an independent processor goroutine
+	errGroup, errGroupCtx := errgroup.WithContext(ctx)
```
ramizpolic, line 29: need to rename to `processGroup, processCtx`

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Create volume if not found
+	resp, err := c.dockerClient.VolumeList(ctx, volume.ListOptions{
+		Filters: filters.NewArgs(filters.Arg("name", scanName)),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to get volumes. Provider=%s: %w", models.Docker, err)
```
ramizpolic, line 133: `Provider=%s` is redundant, we will know the provider issue when the error occurs from the provider interface (e.g. check the `prepareScanVolume` scanner error handling). Same applies to other errors in this method.

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	if len(resp.Volumes) == 1 {
+		fmt.Printf("scan volume already created. Provider=%s", models.Docker)
```
ramizpolic, line 136: Avoid `fmt` package to print logging messages. Also, note that `fmt` package is not thread-safe.
You can create a logger from context via `logger := log.GetLoggerFromContextOrDiscard(ctx)` and use that.

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+	if len(containers) == 0 {
+		return "", fmt.Errorf("scan container not found. Provider=%s: %w", models.Docker, err)
+	}
+	if len(containers) > 1 {
+		return "", fmt.Errorf("found more than one scan container. Provider=%s: %w", models.Docker, err)
+	}
+	return containers[0].ID, nil
+}
+
+func (c *Client) RemoveAssetScan(ctx context.Context, config *provider.ScanJobConfig) error {
```
ramizpolic, line 319: Can we have this public method before all private functions/methods. Usually, I like to follow the style-guide (top to bottom declarations):
1. Public constants and variables
2. Public functions
3. Public types
   - Related public methods first, then private methods
4. Private types
    - Related methods
 5. Private functions

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+				},
+			},
+		}, nil, nil, "")
+	if err != nil {
+		return fmt.Errorf("failed to create helper container. Provider=%s: %w", models.Docker, err)
+	}
+	defer func() {
+		err = c.dockerClient.ContainerRemove(ctx, response.ID, types.ContainerRemoveOptions{})
+		if err != nil {
+			_ = fmt.Errorf("failed to remove helper container. Provider=%s: %w", models.Docker, err)
```
ramizpolic, line 177: You can log this error message using contextual logger, otherwise it has no effect

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+					Source: scanName,
+					Target: "/data",
+				},
+			},
+		}, nil, nil, "")
+	if err != nil {
+		return fmt.Errorf("failed to create helper container. Provider=%s: %w", models.Docker, err)
+	}
+	defer func() {
+		err = c.dockerClient.ContainerRemove(ctx, response.ID, types.ContainerRemoveOptions{})
```
ramizpolic, line 175: If the container gets stuck in an infinite loop, I am not sure that this will remove it. You can specify to forcefully remove it via remove options type.

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to export target. Provider=%s: %w", models.Docker, err)
+	}
+
+	// Create an ephemeral container to populate volume with export output
+	_, err = c.dockerClient.ImagePull(ctx, "alpine", types.ImagePullOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to pull helper image. Provider=%s: %w", models.Docker, err)
+	}
+	response, err := c.dockerClient.ContainerCreate(ctx,
```
ramizpolic, line 158: Naming should be changed, `response` is shadowed by `resp`, and it's also not descriptive

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+	if len(resp.Volumes) == 0 {
+		_, err = c.dockerClient.VolumeCreate(ctx, volume.CreateOptions{
+			Name: scanName,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to create volume. Provider=%s: %w", models.Docker, err)
+		}
+	}
+
+	readCloser, err := c.export(ctx, config)
```
ramizpolic, line 148: Naming is not descriptive

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+func (c *Client) prepareScanVolume(ctx context.Context, config *provider.ScanJobConfig) error {
+	scanName, err := getScanName(config)
+	if err != nil {
+		return err
+	}
+
+	// Create volume if not found
+	resp, err := c.dockerClient.VolumeList(ctx, volume.ListOptions{
```
ramizpolic, line 129: Naming is not descriptive

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to copy data to container. Provider=%s: %w", models.Docker, err)
+	}
+	return nil
+}
+
+func (c *Client) export(ctx context.Context, config *provider.ScanJobConfig) (io.ReadCloser, error) {
+	objectType, err := config.AssetInfo.Discriminator()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get asset object type. Provider=%s: %w", models.Docker, err)
```
ramizpolic, line 190: Provider in the error message is redundant, this will be handled by provider-specific methods

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	switch objectType {
+	case "ContainerInfo":
+		id, err := getAssetId(config)
+		if err != nil {
+			return nil, err
+		}
+		return c.dockerClient.ContainerExport(ctx, id)
+	case "ContainerImageInfo":
```
ramizpolic, line 200: It's nice to put a new line before each new case to signify that is a separate item

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+			&container.Config{
+				Image: name,
+			}, nil, nil, nil, "")
+		if err != nil {
+			return nil, fmt.Errorf("failed to create helper container. Provider=%s: %w", models.Docker, err)
+		}
+		defer func() {
+			err = c.dockerClient.ContainerRemove(ctx, response.ID, types.ContainerRemoveOptions{})
+			if err != nil {
+				_ = fmt.Errorf("failed to remove helper container. Provider=%s: %w", models.Docker, err)
```
ramizpolic, line 216: You can use a contextual logger to log this error message

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+
+	_, err = os.Stat(scanConfigFilePath)
+	if errors.Is(err, os.ErrNotExist) {
+		err = os.WriteFile(scanConfigFilePath, []byte(config.ScannerCLIConfig), 0644)
+		if err != nil {
+			return err
+		}
+	} else if err != nil {
+		return err
+	}
```
ramizpolic, line 239:
```go
// If the error is os.ErrNotExist, try to create a new file and update the error
if errors.Is(err, os.ErrNotExist) {
	err = os.WriteFile(scanConfigFilePath, []byte(config.ScannerCLIConfig), 0644)
}

// If the err was previously os.ErrNotExist, it would have been updated on os.WriteFile
if err != nil {
	return err
}
```

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+		&container.Config{
+			Image: config.ScannerImage,
+			Entrypoint: []string{"sh", "-c",
+				fmt.Sprintf(
+					"/app/vmclarity-cli --config /tmp/%s --server %s --asset-scan-id %s",
+					filepath.Base(scanConfigFilePath),
+					config.VMClarityAddress,
+					config.AssetScanID,
+				),
+			},
```
ramizpolic, line 278: This needs to be broken into a string slice command which calls `/app/vmclarity-cli` directly instead of `sh`

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+func (c *Client) getContainerIdFromContainerName(ctx context.Context, scanName string) (string, error) {
+
+	containers, err := c.dockerClient.ContainerList(ctx, types.ContainerListOptions{
+		All:     true,
+		Filters: filters.NewArgs(filters.Arg("name", scanName)),
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed to get containers. Provider=%s: %w", models.Docker, err)
```
ramizpolic, line 308: Provider in the error message is redundant, this will be handled by provider-specific methods

runtime_scan/pkg/provider/docker/helpers.go
```diff
@@ -0,0 +1,86 @@
...
+	for key, val := range tags {
+		ret = append(ret, models.Tag{
+			Key:   key,
+			Value: val,
+		})
+	}
+	return &ret
+}
+
+func getScanName(config *provider.ScanJobConfig) (string, error) {
```
ramizpolic, line 22: `scanNameFor` is more descriptive than `getScanName` which indicates that some API call is being made

runtime_scan/pkg/provider/docker/helpers.go
```diff
@@ -0,0 +1,86 @@
...
+	id, err := getAssetId(config)
+	if err != nil {
+		return "", fmt.Errorf("failed to get scan name. Provider=%s: %w", models.Docker, err)
+	}
+	return "scan_" + config.ScanID + "_asset_" + strings.Replace(id, ":", "_", -1), nil
+}
+
+func getScanConfigFileName(config *provider.ScanJobConfig) (string, error) {
```
ramizpolic, line 30: `scanConfigFileNameFor` is more descriptive

runtime_scan/pkg/provider/docker/helpers.go
```diff
@@ -0,0 +1,86 @@
...
+	scanName, err := getScanName(config)
+	if err != nil {
+		return "", err
+	}
+
+	// TODO (adamtagscherer): check if os.TempDir() doesn't create a new unique directory every time it's being called
+	return os.TempDir() + scanName + "_scanconfig.yaml", nil
+}
+
+func getAssetId(config *provider.ScanJobConfig) (string, error) {
```
ramizpolic, line 40: `assetIdFor` is more descriptive

runtime_scan/pkg/provider/docker/helpers.go
```diff
@@ -0,0 +1,86 @@
...
+		if err != nil {
+			return "", fmt.Errorf("failed to get asset id. Provider=%s: %w", models.Docker, err)
+		}
+		return *containerImageInfo.Id, nil
+	default:
+		return "", fmt.Errorf("get asset id not implemented for current object type (%s). Provider=%s", models.Docker, objectType)
+	}
+}
+
+func getAssetName(config *provider.ScanJobConfig) (string, error) {
```
ramizpolic, line 64: `assetNameFor` is more descriptive

runtime_scan/pkg/provider/docker/helpers.go
```diff
@@ -0,0 +1,86 @@
...
+			Value: val,
+		})
+	}
+	return &ret
+}
+
+func getScanName(config *provider.ScanJobConfig) (string, error) {
+	id, err := getAssetId(config)
+	if err != nil {
+		return "", fmt.Errorf("failed to get scan name. Provider=%s: %w", models.Docker, err)
```
ramizpolic, line 25: Provider message is redundant

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -264,7 +256,19 @@ func (c *Client) createScanContainer(ctx context.Context, config *provider.ScanJ
...
+	networkResp, err := c.dockerClient.NetworkCreate(
+		ctx,
+		scanName,
+		types.NetworkCreate{
+			CheckDuplicate: true,
+			Driver:         "bridge",
+		},
+	)
+	if err != nil {
+		return "", fmt.Errorf("failed to scan network: %w", err)
```
ramizpolic, line 268: Should be `failed to create network %s: %w`

runtime_scan/pkg/provider/docker/client.go
paralta: marking all comments that refer to `Provider=%s` as resolved, since I addressed this in the latest commit

runtime_scan/pkg/provider/docker/client.go
paralta: thank you! much better :)

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,389 @@
...
+		return provider.FatalErrorf("failed to remove scan network. Provider=%s: %w", models.Docker, err)
+	}
+
+	return nil
+}
+
+func (c *Client) prepareScanVolume(ctx context.Context, config *provider.ScanJobConfig) error {
+	logger := log.GetLoggerFromContextOrDiscard(ctx)
+
+	scanName, err := getScanName(config)
```
fishkerez, line 156: Why not using the AssetScanID (we have it in the config.Metadata)

fishkerez, line 156: scanName->volumeName. It will be less confusing IMO

cli/cmd/root.go
```diff
@@ -108,6 +108,20 @@ var rootCmd = &cobra.Command{
 				return err
 			}
 			setMountPointsForFamiliesInput(mountPoints, config)
+		} else {
+			// Check if volume already mounted
```
fishkerez, line 112: This is too specific for the docker provider implementation. 
If a user wants to scan a path other than "/mnt/snapshot" he will fail

I think you need to pass the "/mnt/snapsho" as part of the families config as an input

cli/cmd/root.go
paralta: I can add this as a constant for the cli and for the provider :)

runtime_scan/pkg/provider/docker/helpers.go
paralta: I was trying to keep the naming consistent with what we have for other providers, e.g. `getInstanceWithID` in AWS

runtime_scan/pkg/orchestrator/config.go
paralta: We reverted this to the previous fallthrough

runtime_scan/pkg/orchestrator/config.go
```diff
@@ -136,10 +136,12 @@ func LoadConfig(backendHost string, backendPort int, baseURL string) (*Config, e
 	switch strings.ToLower(viper.GetString(ProviderKind)) {
 	case strings.ToLower(string(models.Azure)):
 		providerKind = models.Azure
+	case strings.ToLower(string(models.Docker)):
+		providerKind = models.Docker
 	case strings.ToLower(string(models.AWS)):
-		fallthrough
-	default:
 		providerKind = models.AWS
+	default:
```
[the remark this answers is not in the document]
paralta: We reverted this to the previous fallthrough

runtime_scan/pkg/provider/docker/client.go
paralta: solved with https://github.com/openclarity/vmclarity/pull/462#discussion_r1265406115

runtime_scan/pkg/provider/docker/client.go
paralta: agree 👍

cli/cmd/root.go
paralta: looking into these changes now :)

paralta: @fishkerez please have a look at the last commit https://github.com/openclarity/vmclarity/pull/462/commits/61678c56aa60ea6a6c28d5307261f4011f60a035

fishkerez: Looks good

cli/cmd/root.go
paralta: updated with https://github.com/openclarity/vmclarity/pull/462/commits/c99f61c9a6bf68abaf22daa3625e6e2c20fc23e2

runtime_scan/pkg/orchestrator/config.go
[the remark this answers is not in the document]
ramizpolic: @fishkerez was not a fan of using fallthrough, I think that having a specific dedicated default provider is better than fallthrough option (at least, it's more transparent)

runtime_scan/pkg/provider/docker/helpers.go
paralta: func removed

runtime_scan/pkg/provider/docker/helpers.go
paralta: func removed

runtime_scan/pkg/provider/docker/helpers.go
paralta: func removed

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+	if len(networks) > 1 {
+		return "", fmt.Errorf("found more than one scan network: %w", err)
+	}
+	return networks[0].ID, nil
+}
+
+func getScanConfigFileName(config *provider.ScanJobConfig) string {
+	// TODO (adamtagscherer): check if os.TempDir() doesn't create a new unique directory every time it's being called
+	return os.TempDir() + config.AssetScanID + "_scanconfig.yaml"
```
ramizpolic, line 382: Should be `path.Join(os.TempDir(), config.AssetScanID+"_scanconfig.yaml")`

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,394 @@
...
+		return "", fmt.Errorf("scan network not found: %w", err)
+	}
+	if len(networks) > 1 {
+		return "", fmt.Errorf("found more than one scan network: %w", err)
+	}
+	return networks[0].ID, nil
+}
+
+func getScanConfigFileName(config *provider.ScanJobConfig) string {
+	// TODO (adamtagscherer): check if os.TempDir() doesn't create a new unique directory every time it's being called
```
ramizpolic, line 381: Should be `getScanConfigFilePath` rather than FileName.

@adamtagscherer `os.TempDir` does not create any dir, just returns the current os-specific temp dir path.
Note that the temp dir path might not be an absolute path (or a valid/existing path at all), so this might need to be handled too.

ramizpolic: We can keep it like this for now, but it's worth covering this scenario too

sambetts-cisco (changes requested):

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,394 @@
...
+		return nil
+	}
+	if len(volumeResp.Volumes) == 0 {
+		_, err = c.dockerClient.VolumeCreate(ctx, volume.CreateOptions{
+			Name: config.AssetScanID,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to create volume: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
ramizpolic: I don't think you can find more than one identically-named volume in this scope (names are unique to volumes). This should never occur.

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,394 @@
...
+
+	rawContents, err := c.export(ctx, config)
+	if err != nil {
+		return fmt.Errorf("failed to export target: %w", err)
+	}
+
+	// Create an ephemeral container to populate volume with export output
+	pl, err := c.dockerClient.ImagePull(ctx, "alpine", types.ImagePullOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to pull helper image: %w", err)
```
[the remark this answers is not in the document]
ramizpolic: There hasn't been much error handling done; the assumption for now is that every error is fatal. I think in the next iteration we can focus on actually handling the errors properly in terms of the expected workflow.

runtime_scan/pkg/provider/docker/client.go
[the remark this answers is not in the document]
sambetts-cisco: it shouldn't but the API returns a list and there is no guarantee that it len will be exactly 1 or 0, so I think we should guard in the code against the unexpected

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,394 @@
...
+		}, nil, nil, "")
+	if err != nil {
+		return fmt.Errorf("failed to create helper container: %w", err)
+	}
+	defer func() {
+		err = c.dockerClient.ContainerRemove(ctx, containerResp.ID, types.ContainerRemoveOptions{Force: true})
+		if err != nil {
+			logger.Errorf("failed to remove helper container: %s", err.Error())
+		}
+	}()
```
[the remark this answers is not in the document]
ramizpolic: The issue is when the execution chain is broken on CopyToContainer (or any step afterwards, and there will be more introduced), the container will not be cleaned up. We need to ensure that containers are cleaned up.

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,394 @@
...
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create helper container: %w", err)
+		}
+		defer func() {
+			err = c.dockerClient.ContainerRemove(ctx, containerResp.ID, types.ContainerRemoveOptions{Force: true})
+			if err != nil {
+				logger.Errorf("failed to remove helper container: %w", err)
+			}
+		}()
```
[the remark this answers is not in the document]
ramizpolic: Good catch, maybe @paralta has more information on this. If this is the case, then we just need to have an intermediate reader/writer to stream the results into before returning.

runtime_scan/pkg/provider/docker/client.go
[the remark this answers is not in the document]
sambetts-cisco: ah ok yeah that makes sense, perhaps a comment like "defer the container cleanup so that even if CopyToContainer panics the cleanup is done"

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,394 @@
...
+
+func (c *Client) createScanContainer(ctx context.Context, config *provider.ScanJobConfig) (string, error) {
+	containerID, err := c.getContainerIdFromContainerName(ctx, config.AssetScanID)
+	if containerID != "" {
+		return containerID, nil
+	}
+
+	pl, err := c.dockerClient.ImagePull(ctx, config.ScannerImage, types.ImagePullOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to pull scanner image: %w", err)
```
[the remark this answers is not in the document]
ramizpolic: Valid point @sambetts-cisco, but not a focus of this PR. We will handle the actual flow in the next iteration when the E2E tests are in place as well.

runtime_scan/pkg/provider/docker/client.go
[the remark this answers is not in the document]
sambetts-cisco: Whatever we do we must avoid loading all the data into memory, some containers can be multiple GB so that won't be a viable solution

runtime_scan/pkg/provider/docker/image.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to list images: %w", err)
+	}
+
+	// Results will be written to assets concurrently
+	assetMu := sync.Mutex{}
+	assets := make([]models.AssetType, 0, len(images))
+
+	// Process each image in an independent processor goroutine
+	processGroup, processCtx := errgroup.WithContext(ctx)
```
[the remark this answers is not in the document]
ramizpolic: Life was harder back then before errgroup @sambetts-cisco 😺

runtime_scan/pkg/provider/docker/helpers.go
paralta: using get* for consistency 👍

runtime_scan/pkg/provider/docker/client.go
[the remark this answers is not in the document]
paralta: check for `len(volumesResp.Volumes) > 1` just added :)

runtime_scan/pkg/provider/docker/client.go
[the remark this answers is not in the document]
paralta: changes pushed now https://github.com/openclarity/vmclarity/pull/462/commits/3e45d97a6acc24ce010d02e0b7dede4dd66a9f42 :) please have a look

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -204,11 +207,17 @@ func (c *Client) createScanAssetVolume(ctx context.Context, config *provider.Sca
...
 		return "", fmt.Errorf("failed to export asset: %w", err)
 	}
-	defer assetContents.Close()
+	defer func() {
+		err := assetContents.Close()
+		if err != nil {
+			logger.Errorf("failed to close asset contents stream: %s", err.Error())
+		}
+	}()
+	defer exportCleanup()
```
ramizpolic, line 220: No need to have two defers here. Also, there are valid returns with nil cleanup actions.
```go
defer func() {
	err := assetContents.Close()
	if err != nil {
		logger.Errorf("failed to close asset contents stream: %s", err.Error())
	}

	if exportCleanup != nil {
		exportCleanup()
	}
}()
```

ramizpolic (changes requested):

runtime_scan/pkg/provider/docker/client.go
paralta: indeed! thanks, Ramiz!

paralta: pushed 👍

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,452 @@
...
+		cleanup := func() {
+			err := c.dockerClient.ContainerRemove(ctx, containerResp.ID, types.ContainerRemoveOptions{Force: true})
+			if err != nil {
+				logger.Errorf("failed to remove helper container=%s: %v", containerResp.ID, err)
+			}
+		}
+
+		contents, err := c.dockerClient.ContainerExport(ctx, containerResp.ID)
+		if err != nil {
+			return nil, cleanup, fmt.Errorf("failed to export container: %w", err)
```
ramizpolic, line 430: This cleanup action will not be executed due to non-nil error checks where exportAsset is run. Instead, you can execute cleanup action and return nil action before returning an error.

```go
cleanup()
return nil, nil, fmt.Errorf("failed to export container: %w", err)
```
Ideally, cleanup action is only returned when no errors have occurred, otherwise it is handled internally.

ramizpolic: This behaviour should also be documented for `exportAsset` method:
* Returns artifact in the form of a ReadCloser stream, along with a cleanup function that needs to be run once the work with the artifact is done. Cleanup action is only returned on nil error, and it can also be nil to indicate that no cleanup action needs to be performed.

ramizpolic (changes requested):

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+
+	networkID, err := c.getNetworkIDFromName(ctx, config.AssetScanID)
+	if err != nil {
+		return provider.FatalErrorf("failed to get scan network id. Provider=%s: %w", models.Docker, err)
+	}
+	err = c.dockerClient.NetworkRemove(ctx, networkID)
+	if err != nil {
+		return provider.FatalErrorf("failed to remove scan network. Provider=%s: %w", models.Docker, err)
+	}
```
ramizpolic, line 141: This will create race issues since the network is static and is used by all runners, plus it will fail every time there's another container attached to it. I think it's fine to have this network present without having to remove it once created. We can remove this cleanup action.

ramizpolic (changes requested):

runtime_scan/pkg/provider/docker/client.go
paralta: removed, thanks!

ramizpolic (approved):

chrisgacsal (approved):
Awesome work! LGTM! :shipit:

paralta:
@sambetts-cisco any further comments to add to this pull request?

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,456 @@
...
+	"github.com/openclarity/vmclarity/runtime_scan/pkg/provider"
+	"github.com/openclarity/vmclarity/shared/pkg/families"
+	"github.com/openclarity/vmclarity/shared/pkg/log"
+)
+
+var (
+	// mountPointPath defines the location in the container where assets will be mounted.
+	mountPointPath = "/mnt/snapshot"
+	// helperImage defines helper container image that performs init tasks.
+	helperImage = "alpine"
```
sambetts-cisco, line 46: I don't think my comment was addressed about making this configurable and also pinning a version right now we could get any alpine, even one with a nasty malware if someone poisoned upstream

sambetts-cisco (commented):
Awesome work guys, thanks for this! I had one comment that I don't think was addressed, perhaps @ramizpolic  you were going  to handle this in a follow up?

In the future can we avoid resolving conversations that someone else started, I had to go back through all the collapsed conversations to find my reviews to check they the code changes had actually answered all the questions I had. I don't think a conversation should be resolved until the person who left the review is happy with the result.

runtime_scan/pkg/provider/docker/client.go
```diff
@@ -0,0 +1,456 @@
...
+	// networkName defines the user defined bridge network where we attach the scanner container.
+	networkName = "vmclarity"
```
ramizpolic, line 48: Can we have this configurable via ENV Flag? I think it will help @adamtagscherer with E2E testing quite better.

ramizpolic (changes requested):

runtime_scan/pkg/provider/docker/client.go
ramizpolic: fyi Just `networkName` should be configurable via ENV flag

runtime_scan/pkg/provider/docker/client.go
sambetts-cisco: I rather helper image was configurable through an ENV with a default, if for whatever reason dockerhub isn't accessible or a deployer must use a local docker registry because they aren't allowed internet access. We should allow it to be overridden.

runtime_scan/pkg/provider/docker/client.go
ramizpolic, line 48: As a summary, we want to have all these values configurable via ENV var with default values.

ramizpolic (changes requested):

runtime_scan/pkg/provider/docker/client.go
ramizpolic: Sure, can't see why not. I summarised it in a new review item @sambetts-cisco

runtime_scan/pkg/provider/docker/client.go
paralta: @ramizpolic @sambetts-cisco thanks for reviewing again! I will take care of these changes 👍

paralta: We will change helper image and network name, but keep the mount point path as a constant inside the docker client as this path is only used in this client scope.

adamtagscherer:
@ramizpolic @sambetts-cisco I moved the helper image and the network to a config file.

ramizpolic (approved):
Awesome! 🚀

sambetts-cisco (approved):

ramizpolic (approved):

PR #473, opened 2023-07-17
feat: Collect scanner statistics during AssetScan

fishkerez:
## Description

To be able to dynamically estimate time and cost of scans, we need to
gather statistics about the scanned filesystem size and time benchmarks
per family per input.  This information is exported to the AssetScan
where the control plane can use it to estimate the cost of future scans.

## Type of Change

[ ] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)


api/openapi.yaml
```diff
@@ -1626,6 +1628,46 @@ components:
...
+          $ref: '#/components/schemas/AssetScanBenchmark'
+        misconfigurations:
+          $ref: '#/components/schemas/AssetScanBenchmark'
+        exploits:
+          $ref: '#/components/schemas/AssetScanBenchmark'
+
+    AssetScanBenchmark:
+      type: object
+      properties:
+        scannedSize:
```
akpsgit, line 1654: Maybe it's better to put a list of inputs here that for each we'll state:
input type
Input path
Size
start/end time
number of files (optional)

api/openapi.yaml
```diff
@@ -1626,6 +1628,46 @@ components:
       #  - asset
       #  - scan
 
+    AssetScanBenchmarks:
```
akpsgit, line 1631: maybe AssetScanStats?
not all the info inside will be regarding benchmarking but also input info

api/openapi.yaml
```diff
@@ -1626,6 +1628,78 @@ components:
...
+      description: Global statistics for asset scan of all families.
+      properties:
+        size:
+          description: General scanned size in MB.
+          type: integer
+          format: int64
+        scanTime:
+          $ref: '#/components/schemas/AssetScanScanTime'
+
+    AssetScanInputStats:
```
akpsgit, line 1676: AssetScanInputScanStats

akpsgit (dismissed):

api/openapi.yaml
```diff
@@ -1626,6 +1628,78 @@ components:
...
+        exploits:
+          type: array
+          items:
+            $ref: '#/components/schemas/AssetScanInputStats'
+
+    AssetScanGeneralStats:
+      type: object
+      description: Global statistics for asset scan of all families.
+      properties:
+        size:
```
sambetts-cisco, line 1669: I think "size" is the wrong name here, perhaps "totalScannedMb" ?

sambetts-cisco (approved):

api/openapi.yaml
```diff
@@ -1652,6 +1652,8 @@ components:
               type: boolean
             resourceCleanup:
               $ref: '#/components/schemas/ResourceCleanupState'
+            stats:
+              $ref: '#/components/schemas/AssetScanStats'
```
sambetts-cisco, line 1656: I think we also need to add a startTime and endTime to the AssetScan itself. The start time should be set from the the moment the AssetScan moves from Pending to Scheduled and scanEndTime should be set when the AssetScan moves to Done or Failed. That will allow us to calculate the duration the AssetScan really took to scan from the point we start creating resources. We can also add cleanupEndTime, which is the point when the cleanup finishes which will give us the full e2e time for an AssetScan.

Adding these fields is something we need anyway in order to implement a per AssetScan timeout and I'm sure there are other use cases too.

cli/pkg/state/vmclarity.go
```diff
@@ -145,6 +156,48 @@ func (v *VMClarityState) MarkDone(ctx context.Context, errors []error) error {
...
+			totalScanned += *scanStats.Size
+		}
+	}
+	if stats.Malware != nil {
+		for _, scanStats := range *stats.Malware {
+			totalScanned += *scanStats.Size
+		}
+	}
+
+	return &totalScanned
```
sambetts-cisco, line 198: I'm still not 100% sure on this value and its semantics. For example if I have 500mb of data, and I scan for misconfigurations and exploits, totalScannedMb will be 1000mb. This double counting is useful I think for estimation because we really are running over this much data.

I think we need to make sure that we highlight this in the description for the field / in the UI. Just so that people aren't surprised that if they have a VM with a 30GB disk but see TotalScannedMB = 6000MB

fishkerez: I think this value can also be calculated in the backend/UI because it is just aggregate all the results from the assetScan. @akpsgit WDYT?

api/openapi.yaml
fishkerez: Not sure this is part of this PR though. This is not part of the scan benchmarking, but times that took to the orchestrator to run the scan

cli/pkg/state/vmclarity.go
fishkerez: So we can remove it from the CLI reporting, Do we want to also remove it from the API?

sambetts-cisco: I would lean towards removing it from the API for now, as it seems to be redundant, we can always add it back later if we need the aggregate info. I expect the majority of the time we'll probably be using the scanner specific info though.

fishkerez: We can't remove it from the API without a UI support. Maybe we should just leave it for now, it does give more info, and it is already presented nicely in the UI

api/openapi.yaml
sambetts-cisco: Added issue https://github.com/openclarity/vmclarity/issues/509

akpsgit (approved):

sambetts-cisco (approved):

PR #497, opened 2023-07-25
Add IAM policies

ramizpolic:
## Description

Addresses openclarity/openclarity#726 

All changes summarised in https://github.com/openclarity/vmclarity/tree/feature/rbac/pkg/apiserver/iam#readme

### High level architecture
https://github.com/openclarity/vmclarity/tree/feature/rbac/pkg/apiserver/iam/testdata/architecture.png
#### Server-side stack
- **Authenticator** - Interacts with IDP to authenticate request and obtain user identity data. Part of API middleware.
- **RoleSyncer** - Interacts with some kind of store (e.g. database or from JWT token claim) to fetch and sync user role data. Used in API middleware.
- **Authorizer** - Decides if a User can perform a given action on an asset based on provided rules.
Separate component that is used to check user-action permissions.
Used in API middleware for API asset-action (e.g. `api:put`) authorization.

### Test environment
Requires Docker to run. This will create a Zitadel instance with bootstrapped data which you can use
out-of-the-box to enable IAM for VMClarity.

```bash
# Create Zitadel
cd pkg/apiserver/iam/testdata/zitadel
chmod +x ./create-zitadel
RECREATE=true ./create-zitadel

# Get autogenerated file and obtain VMClarity config data.
# Update if required.
cat bootstrap/generated/vmclarity-data.env
```

### Config parameters
```
# IAM global config
IAM_ENABLED=true

# Authentication config
AUTH_OIDC_ISSUER=http://localhost:8080
AUTH_OIDC_CLIENT_ID=app_client_id
AUTH_OIDC_CLIENT_SECRET=app_client_secret

# RoleSynchronization config
ROLESYNCER_JWT_ROLE_CLAIM=roles-claim-key

# Authorization config
AUTHZ_LOCAL_RBAC_RULE_FILEPATH=path-to/rbac_rule_policy_example.csv

# Injection - Client service account
APISERVER_TOKEN=Service Account PAM
```

### TODO (next iterations)
- Add Zitadel component into VMClarity stack. Use Zitadel bootstrapping without using Terraform.
- Add support for injecting auth data into different parts of request (e.g. cookies).
- Handle http response codes properly (currently only 403 is returned on Auth/Z failure)
- Add Authorizer that uses database as Role permission source. Also add bootstrapping for the Role and Rule data.
- Add RoleSyncer that uses database as User Role source to avoid using non-intuitive JWT Claim Role Source.
- Add support for different IAM components (Authenticator and RoleSyncer) to avoid relying only on Zitadel.
- Add support to dynamically add/remove (CRUD) multiple Authenticators in IAM stack.
- Add tests

### More resources
- **Zitadel** - used to create OIDP app and auth roles via JWT token claims - https://zitadel.com/
- **Casbin** - used to validate roles against defined role policies - https://casbin.org/
- **Casbin** Adapters - to extend role rule data synchronization - https://casbin.org/docs/adapters

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[x] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


runtime_scan/pkg/provider/aws/client.go
```diff
@@ -46,9 +46,9 @@ func New(ctx context.Context) (*Client, error) {
 		return nil, fmt.Errorf("invalid configuration. Provider=AWS: %w", err)
 	}
 
-	if err = config.Validate(); err != nil {
-		return nil, fmt.Errorf("failed to validate provider configuration. Provider=AWS: %w", err)
-	}
+	//if err = config.Validate(); err != nil {
+	//	return nil, fmt.Errorf("failed to validate provider configuration. Provider=AWS: %w", err)
+	//}
```
ramizpolic, line 51: This needs to be re-enabled

api/openapi.yaml
```diff
@@ -1675,7 +1687,6 @@ components:
 
     AssetScanTemplate:
       type: object
-      x-go-type-name: "AssetScanTemplate"
```
ramizpolic, line 1678: This should not change anything, but it allows bumping openapi generator version to v.1.13.0

pkg/backend/config/config.go
```diff
@@ -24,39 +24,80 @@ import (
...
+	FakeDataEnvVar   = "FAKE_DATA"
+	LocalDBPath      = "LOCAL_DB_PATH"
 
-	LocalDBPath = "LOCAL_DB_PATH"
+	LogLevel = "LOG_LEVEL"
+)
 
-	FakeDataEnvVar      = "FAKE_DATA"
-	DisableOrchestrator = "DISABLE_ORCHESTRATOR"
+type AuthenticationOIDC struct {
```
fishkerez, line 62: Not sure this is the right place to define these. Maybe under iam/types?

pkg/backend/iam/rolesyncer/jwt.go
```diff
@@ -0,0 +1,75 @@
...
+	return &jwtRoleSyncer{
+		roleClaim: roleClaim,
+	}
+}
+
+func (roleSyncer *jwtRoleSyncer) Type() iam.RoleSyncerType {
+	return RoleSyncerTypeJwt
+}
+
+func (roleSyncer *jwtRoleSyncer) Sync(ctx context.Context, user *iam.User) error {
```
fishkerez, line 41: please add a unit test for this function

pkg/backend/iam/testdata/architecture.drawio
```diff
@@ -0,0 +1,131 @@
+<mxfile host="app.diagrams.net" modified="2023-07-28T12:31:32.181Z" agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36" etag="zbHU9S2qkvYVcnmDOjsn" version="21.6.5" type="device">
```
fishkerez, line 1: do we need this file?

pkg/backend/rest/server.go
```diff
@@ -75,20 +92,24 @@ func createEchoServer(dbHandler databaseTypes.Database, uiSitePath string, uiBac
...
 	apiGroup := e.Group(BaseURL)
 
-	// Use oapi-codegen validation middleware to validate
-	// the API group against the OpenAPI schema.
-	apiGroup.Use(middleware.OapiRequestValidator(swagger))
+	// Use oapi-codegen validation middleware to validate the API group against the
+	// OpenAPI schema along with IAM provider.
+	apiGroup.Use(middleware.OapiRequestValidatorWithOptions(swagger, &middleware.Options{
+		Options: openapi3filter.Options{
+			AuthenticationFunc: iam.OapiFilterForProvider(iamProvider),
```
fishkerez, line 99: Only if "config.IamEnabled" we should set the Authentication middleware

api/openapi.yaml
```diff
@@ -2421,3 +2433,14 @@ components:
...
         type: integer
+
+  securitySchemes:
+    IamPolicy:
+      type: http
+      scheme: bearer
+      bearerFormat: JWT
+
+# Specify global RBAC requirements, must be in form of asset:action, e.g. api:get
+security:
```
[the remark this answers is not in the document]
ramizpolic: Yes, global security requirements will be applied to all operations. Note that this is the first iteration of RBAC, so this served simply as an example.

pkg/backend/rest/server.go
ramizpolic: Yes. However, in this particular configuration, when `config.IamEnabled = false`, there will be no authentication function (actually, there will, but it will do nothing).

pkg/backend/iam/provider/oidc.go
```diff
@@ -0,0 +1,105 @@
...
+	// supported
+	default:
+		return nil, fmt.Errorf("unsupported role syncer type provided: %s", roleSyncerType)
+	}
+
+	// Add custom OIDC options
+	var options []rs.Option
+	if config.TokenURL != "" && config.IntrospectURL != "" {
+		options = append(options, rs.WithStaticEndpoints("tokenUrl", "introspectUrl"))
+	} else if config.TokenURL != "" || config.IntrospectURL != "" {
```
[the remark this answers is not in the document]
ramizpolic: Simple else check would cover both cases when these fields are unset or when at least one is unset. 
This check ensures that static endpoint can be set only when both `TokenURL` and `IntrospectURL` are set. 
If only one of them is set, then we cannot configure the static endpoint (error).

pkg/backend/config/config.go
```diff
@@ -24,39 +24,80 @@ import (
...
+
+	AuthorizationRbacRuleFilePathEnvVar = "AUTHZ_RBAC_RULE_FILE_PATH"
+
 	BackendRestHost       = "BACKEND_REST_HOST"
 	BackendRestDisableTLS = "BACKEND_REST_DISABLE_TLS"
 	BackendRestPort       = "BACKEND_REST_PORT"
 	HealthCheckAddress    = "HEALTH_CHECK_ADDRESS"
 
+	DisableOrchestrator               = "DISABLE_ORCHESTRATOR"
+	OrchestratorAuthBearerTokenEnvVar = "ORCHESTRATOR_AUTH_BEARER_TOKEN"
```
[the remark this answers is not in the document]
ramizpolic: Resolved

pkg/backend/config/config.go
ramizpolic: Resolved

pkg/backend/iam/provider/oidc.go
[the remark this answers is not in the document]
ramizpolic: Moved handling of this case into config validation

pkg/backend/iam/testdata/architecture.drawio
ramizpolic: Resolved

pkg/backend/iam/rolesyncer/jwt.go
ramizpolic: Resolved

pkg/apiserver/iam/provider/rolesyncer/rolesyncer.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"fmt"
+
+	"github.com/openclarity/vmclarity/api/models"
+	"github.com/openclarity/vmclarity/pkg/apiserver/iam"
+	"github.com/openclarity/vmclarity/pkg/apiserver/iam/provider/rolesyncer/jwtrole"
+)
+
+// New creates a new iam.RoleSyncer.
+func New(kind models.IamRoleSyncer) (iam.RoleSyncer, error) {
```
sambetts-cisco, line 27: From having a look around at different identity providers (github, slack etc) and looking at the goth library, it seems like in the majority of cases there isn't a way to customise the scopes/claims inside the token that is returned by the identity provider.

It seems like it would be a better idea to have a users construct in our DB which contains the user's identities and roles something like:

```
{
   "id": "<uuid 4>",
   "displayName": "Blah Blah",
   "identities": [
       {
           "Provider": "Github",
           "UserID": "blahblah@gmail.com"
        }
   ],
   "roles": [
       "foo",
       "bah",
       "baz"
   ],
}
```

this can then be used for all identity providers. This is kind of moving a feature of Zitadel into our code, but unless we want to make Zitadel a hard requirement for OpenClarity then I think we need it.

To support true multi-tenancy and the "SaaS" like nature of things we'd also probably want to make the identity providers configurable from the frontend/API so that you don't need to decide all of that as part of the deployment.

On that note, I don't think it would be a bad idea to consider making Zitadel a required component of the stack. Right now we havn't added anything to do with redirects to external providers and the required callbacks, user management UI, users with multiple identities from different providers, users which belong to different tenants etc. Unless we want to build all that ourselves then it seems like deploying Zitadel in the stack makes sense.

sambetts-cisco (changes requested):

pkg/apiserver/iam/testdata/zitadel/bootstrap/main.tf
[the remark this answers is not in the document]
ramizpolic: Yes, this is possible. In the current context, Terraform is used to bootstrap a Zitadel instance (create required projects, preload it with our redefined roles, service accounts, etc.). 

However, the actual painful point here is handling changes related to IAM (migrations, upgrades...). 
If, for any reason, we want to do this via config file during docker compose, we'd have to redeploy the whole zitadel stack, which is not good (db needs to be redeployed too). 
Otherwise, we'd have to use the Zitadel API to perform these changes, which again is quite cumbersome and limited. Terraform offered the easiest way to preconfigure the Zitadel stack, while also allowing for easy migration/upgrade flow.

I can provide a simple preconfigured docker compose deployment, but note the issues I mentioned above.

pkg/apiserver/iam/injector/envbearer/envbearer.go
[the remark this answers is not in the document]
ramizpolic: This is great! I completely missed these. Thanks for this @sambetts-cisco. I will remove the package fully since it is not relevant anymore.

pkg/apiserver/iam/testdata/zitadel/bootstrap/main.tf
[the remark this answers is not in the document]
sambetts-cisco: right now we don't have any real upgrade path for any of the components so I think it should be ok to statically bootstrap, and then we need to consider longer term how to handle reconfiguration for all of VMClarity.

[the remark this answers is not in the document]
sambetts-cisco: if we just bootstrap with a default global admin with a fixed password, and the service accounts, and then additional users can be added through the U.

We might need to create some init containers for the microservices to auth with Zitadel and generate their inital tokens so that everything can be part of the same stack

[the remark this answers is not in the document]
ramizpolic: We can do that, but I will create a separate issue/PR for this to avoid blowing this PR. This PR introduces IAM principles, the rest of the job is to figure out the deployment of the IAM stack. 
Currently, there is a simple flow for automated deployment and bootstrapping of the IAM stack, albeit relying on Terraform, but it should be straightforward to pick this up and add support for different VMClarity installation methods.

pkg/apiserver/iam/iam.go
```diff
@@ -0,0 +1,94 @@
...
+// OapiFilterForProvider creates an OpenAPI middleware filter function which
+// handles request authentication and authorization using a specific Provider.
+// Logic flow:
+//
+// request -> |Authenticator.Authenticate| -> |RoleSyncer.Sync| -> |Authorizer.CanPerform| -> success
+//
+// Provider will first authenticate the client from request data, synchronize
+// user roles on success, and finally try to authorize the request. If
+// successful, User will be available for fetching in context.
+func OapiFilterForProvider(provider Provider) openapi3filter.AuthenticationFunc {
```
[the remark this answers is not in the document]
fishkerez: +1 for calling Provider IAMProvider. Provider is already reserved for cloud provider in our code

pkg/apiserver/iam/injector/envbearer/envbearer.go
```diff
@@ -0,0 +1,45 @@
...
+		tokenEnv: tokenEnv,
+	}, nil
+}
+
+type bearer struct {
+	tokenEnv string
+}
+
+func (injector *bearer) Inject(_ context.Context, request *http.Request) error {
+	request.Header.Set("Authorization", "Bearer "+os.Getenv(injector.tokenEnv))
```
fishkerez, line 43: As @sambetts-cisco already mentioned, I don'r think we should read the token from an env var on each request, we should probably load it to memory at boot.

pkg/apiserver/iam/types.go
```diff
@@ -0,0 +1,81 @@
...
+	// Inject fetches the authentication data from a source and injects them into
+	// request data (e.g. token from env var into Authorization header).
+	Inject(ctx context.Context, request *http.Request) error
+}
+
+// Authenticator implements server-side authentication methods. It is only
+// intended to be used by server middleware to enable authentication.
+type Authenticator interface {
+	// Authenticate validates and verifies user auth details from request against an
+	// auth provider. Publicly exported User fields should be set, or at least User.ID.
```
fishkerez, line 56: `Publicly exported User fields should be set, or at least User.ID`
Not sure I understand this sentence:

Do you mean it should be set on the context?
 What does "Publicly exported User fields" mean?
 Who should set this?

pkg/apiserver/iam/provider/authenticator/oidc/config.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	if c.ClientSecret == "" {
+		return fmt.Errorf("must specify client secret for Authenticator=OIDC")
+	}
+
+	// Validate endpoints, it must specify either none or both endpoints
+	hasTokenURL := c.TokenURL != ""
+	hasIntrospectURL := c.IntrospectURL != ""
+	if hasTokenURL != hasIntrospectURL {
+		return fmt.Errorf("must specify both token and introspect endpoints for Authenticator=OIDC")
```
fishkerez, line 66: must specify either none or both endpoints...

pkg/uibackend/config.go
```diff
@@ -38,9 +40,10 @@ func LoadConfig() (*Config, error) {
 	viper.SetDefault(ListenAddress, ":8890")
 
 	c := &Config{
-		ListenAddress: viper.GetString(ListenAddress),
-		APIServerHost: viper.GetString(APIServerHost),
-		APIServerPort: viper.GetInt(APIServerPort),
+		ListenAddress:              viper.GetString(ListenAddress),
+		APIServerHost:              viper.GetString(APIServerHost),
+		APIServerPort:              viper.GetInt(APIServerPort),
+		APIServerBearerTokenEnvVar: viper.GetString(APIServerBearerTokenEnvVar),
```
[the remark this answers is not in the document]
ramizpolic: Yea, it wasn't even used, the flag was used instead, I just forgot to remove it from the config.

pkg/apiserver/iam/injector/envbearer/envbearer.go
ramizpolic: Resolved

pkg/apiserver/iam/types.go
ramizpolic: Resolved

pkg/orchestrator/config.go
```diff
@@ -33,9 +33,10 @@ import (
 )
 
 const (
-	APIServerHost       = "APISERVER_HOST"
-	APIServerDisableTLS = "APISERVER_DISABLE_TLS"
-	APIServerPort       = "APISERVER_PORT"
+	APIServerHost              = "APISERVER_HOST"
+	APIServerDisableTLS        = "APISERVER_DISABLE_TLS"
+	APIServerPort              = "APISERVER_PORT"
+	APIServerBearerTokenEnvVar = "APISERVER_BEARER_TOKEN_ENV_VAR"
```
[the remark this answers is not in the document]
ramizpolic: Resolved

api/openapi.yaml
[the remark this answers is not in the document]
ramizpolic: TODO: Add more granular rules for other API paths and other components

pkg/apiserver/iam/provider/authenticator/oidc/oidc.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/openclarity/vmclarity/pkg/apiserver/iam"
+
+	"github.com/zitadel/oidc/pkg/client/rs"
+	"github.com/zitadel/oidc/pkg/oidc"
```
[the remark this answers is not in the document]
ramizpolic: Resolved

pkg/apiserver/iam/testdata/zitadel/bootstrap/main.tf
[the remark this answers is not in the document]
ramizpolic: Another note, many of these items have been summarised in the TODO section.

ramizpolic:
> Hey @ramizpolic, general question: How can a user achieve a token in order to use it in his requests? Is there some serviceAccount/userAccount concept in the current implementation?

If we use Zitadel as a IDP, we can generate both user and service accounts. After creation, it is possible to issue PATs (personal access tokens) which have time duration and can be used as simple bearer tokens in requests.

api/openapi.yaml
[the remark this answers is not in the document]
ramizpolic: Another note, I have added example permissions to `/assets` and `/assets/{assetID}` routes. Adding and changing rules should be trivial now.

pkg/apiserver/rest/server.go
```diff
@@ -68,15 +74,27 @@ func createEchoServer(dbHandler databaseTypes.Database) (*echo.Echo, error) {
...
 
-	apiImpl := &ServerImpl{
-		dbHandler: dbHandler,
+		e.Use(middleware.OapiRequestValidatorWithOptions(swagger, &middleware.Options{
+			Options: openapi3filter.Options{
+				AuthenticationFunc: iam.OapiFilterForProvider(iamProvider),
+			},
+		}))
+	} else {
+		e.Use(middleware.OapiRequestValidator(swagger))
```
fishkerez, line 91: I think we need the validator middleware in anycase (also when iamEnabled)

ramizpolic: This is already the case:
* when IAM is enabled, we use `middleware.OapiRequestValidatorWithOptions` which contains validator with authenticator
* when IAM is disabled, we use `middleware.OapiRequestValidator` which only contains validator (and no authenticator)

ramizpolic: Also, `middleware.OapiRequestValidator(swagger)` is just facade call to `OapiRequestValidatorWithOptions(swagger, nil)`

fishkerez: Ah ok sorry for that

fishkerez:
Thanks @ramizpolic . I think it is worth adding a README in the testdata section on how to set up a full working environment with Zitadel, and to achieve a token

ramizpolic:
> Thanks @ramizpolic . I think it is worth adding a README in the testdata section on how to set up a full working environment with Zitadel, and to achieve a token

There already is a short script and walkthrough (section Test environment in README) for setting up test environment. The script creates an env file which can be used when starting VMClarity.

pkg/apiserver/iam/testdata/architecture.png
Tehsmash, line 1: I'm not sure testdata is the right directory to store this IMO it should be for static data/scripts that aid the unit tests for the apiserver, these kind of docs/diagrams should be in the docs directory, and accessible through the README.

The Zitadel stuff deployment should probably be in a top level directory somewhere link "iamPOC" or something, until it gets integrated as part of the installation flows.

Tehsmash (changes requested):

PR #514, opened 2023-07-31
feat: introduce helm chart for kubernetes installation

### LINKED ISSUE #471: Add Kubernetes release artifacts
Currently, we support AWS and Azure deployments. It would be beneficial to add Kubernetes deployments (preferably a Helm chart) for VMClarity.

ghost:
## Description

This commit adds a helm chart for installing the VMClarity control plane on
Kubernetes. Communications between the control plane services is
performed internal to the cluster. Postgresql is used as the database
which allows the API Server, UI and UI Backend services to be scaled
horizontally.

While the control plane runs on Kubernetes the provider is still
responsible for where the scanner instances run, and how they
communicate back to the control plane.

Closes #471 

## Type of Change

[ ] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


charts/vmclarity/templates/apiserver/deployment.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ printf "%s-apiserver" (include "vmclarity.names.fullname" .) }}
```
FrimIdan, line 4: can be simplified to `{{ include "vmclarity.names.fullname" . }}-apiserver`

charts/vmclarity/templates/apiserver/deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+  namespace: {{ .Release.Namespace | quote }}
```
FrimIdan, line 5: `| quote` is not needed

charts/vmclarity/templates/apiserver/deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+  labels: {{- include "vmclarity.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: apiserver
+spec:
+  selector:
+    matchLabels: {{- include "vmclarity.labels.matchLabels" . | nindent 6 }}
+      app.kubernetes.io/component: apiserver
+  replicas: {{ .Values.apiserver.replicaCount }}
+  template:
+    metadata:
+      labels: {{- include "vmclarity.labels.standard" . | nindent 8 }}
```
FrimIdan, line 15: I think `matchLabels` should be used here, also for all other deployments?

charts/vmclarity/templates/apiserver/deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+        app.kubernetes.io/component: apiserver
+    spec:
+      containers:
+        - name: apiserver
+          image: {{ include "vmclarity.images.image" ( dict "imageRoot" .Values.apiserver.image "global" .Values.global ) }}
+          imagePullPolicy: {{ .Values.apiserver.image.pullPolicy | quote }}
+          args:
+            - run
+            - --log-level
+            - info
```
FrimIdan, line 25: should be configurable, also for all other deployments?

charts/vmclarity/templates/apiserver/deployment.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ printf "%s-apiserver" (include "vmclarity.names.fullname" .) }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "vmclarity.labels.standard" . | nindent 4 }}
```
FrimIdan, line 6: wouldn't it be easier to create an `_helper.tpl` for each component with the full name, the labels, etc..?
I see there are few places we are using the same stuff.

charts/vmclarity/templates/apiserver/service-headless.yaml
```diff
@@ -0,0 +1,21 @@
+{{/*
+Headless Service does not include a ClusterIP so that the DNS record includes
+all Pod addresses. This makes it possible for the API Gateway to load balance
+all replicas instead of relying on kube-proxy to do it.
```
FrimIdan, line 4: how does the API GW is doing the LB?

charts/vmclarity/templates/apiserver/service-headless.yaml
```diff
@@ -0,0 +1,21 @@
...
+*/}}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ printf "%s-apiserver-hl" (include "vmclarity.names.fullname" .) | trunc 63 | trimSuffix "-" }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "vmclarity.labels.standard" . | nindent 4 }}
```
FrimIdan, line 11: I think we must have a `vmclarity-apiserver.labels.standard` for example, here you are missing the `app.kubernetes.io/component: apiserver` for example.

charts/vmclarity/templates/apiserver/service-headless.yaml
```diff
@@ -0,0 +1,21 @@
+{{/*
+Headless Service does not include a ClusterIP so that the DNS record includes
+all Pod addresses. This makes it possible for the API Gateway to load balance
+all replicas instead of relying on kube-proxy to do it.
+*/}}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ printf "%s-apiserver-hl" (include "vmclarity.names.fullname" .) | trunc 63 | trimSuffix "-" }}
```
FrimIdan, line 9: why doing only here the `| trunc 63 | trimSuffix "-"`?

charts/vmclarity/values.yaml
```diff
@@ -0,0 +1,64 @@
...
+    tag: latest
+    pullPolicy: IfNotPresent
+
+gateway:
+  replicas: 1
+
+  image:
+    registry: docker.io
+    repository: library/nginx
+    tag: latest
```
FrimIdan, line 46: not sure we want to use `latest` here.

charts/vmclarity/values.yaml
```diff
@@ -0,0 +1,64 @@
...
+    pullPolicy: IfNotPresent
+
+postgresql:
```
FrimIdan, line 49: can you add also
```
  ## Enable security context
  containerSecurityContext:
    enabled: true
    runAsUser: 1001
    runAsNonRoot: true
```

charts/vmclarity/values.yaml
```diff
@@ -0,0 +1,64 @@
+global:
+  imageRegistry: ""
+
+apiserver:
+  replicas: 1
```
FrimIdan, line 5: What about `podSecurityContext`, `securityContext`, `resources`, `nodeSelector`, `tolerations`, `affinity`?

charts/vmclarity/templates/apiserver/deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+  selector:
+    matchLabels: {{- include "vmclarity.labels.matchLabels" . | nindent 6 }}
+      app.kubernetes.io/component: apiserver
+  replicas: {{ .Values.apiserver.replicaCount }}
+  template:
+    metadata:
+      labels: {{- include "vmclarity.labels.standard" . | nindent 8 }}
+        app.kubernetes.io/component: apiserver
+    spec:
+      containers:
```
FrimIdan, line 18: can you please create a service account for each component and use it in `serviceAccountName`?

FrimIdan, line 18: what about `readinessProbe` and `livenessProbe`?

charts/vmclarity/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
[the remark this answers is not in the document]
sambetts-cisco: LICENSE is one of the standard parts of the helm chart spec, it is shipped with the helm when its packaged and defines that the templates are all Apache licensed.

charts/vmclarity/templates/apiserver/deployment.yaml
sambetts-cisco: a number is a valid namespace name, for example 0123, unless namespace is quoted the kubernetes request fails because yaml converts it to an int instead of a string.

charts/vmclarity/templates/apiserver/deployment.yaml
sambetts-cisco: we want all the labels on the pod, not just the ones used for matching

charts/vmclarity/templates/apiserver/deployment.yaml
sambetts-cisco: +1 I really like this, I'll do that

charts/vmclarity/templates/apiserver/deployment.yaml
FrimIdan: do we?

charts/vmclarity/templates/apiserver/service-headless.yaml
FrimIdan: And why it is better than the k8s service lb algorithm?

charts/vmclarity/templates/apiserver/deployment.yaml
sambetts-cisco: IMO yes, when you do `kubectl describe pod vmclarity-apiserver...` I would expect to see the full set of labels including version etc the same as all the other resources not just the name, instance and component ones which are used for matching a pod to a service/deployment.

charts/vmclarity/templates/apiserver/service-headless.yaml
sambetts-cisco: kube-proxy "loadbalancer" algorithm is fixed and it just picks a random pod from the list of Ready pods for the service. Using a headless service pushes this logic to something which can be smarter about it, for example nginx our current gateway can be configured to load balance round-robin (the default), by least number of connections, or by IP hash. https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/#choosing-a-load-balancing-method

FrimIdan: According to https://kodekloud.com/blog/clusterip-nodeport-loadbalancer/# k8s is also doing round-robin style for pods. But whatever to think is good.

charts/vmclarity/templates/apiserver/service-headless.yaml
sambetts-cisco: this was left around, I've remove it as part of the new helpers

charts/vmclarity/templates/apiserver/deployment.yaml
sambetts-cisco: vmclarity currently has no need to access kubernetes resources, I can make the serviceAccountName configurable but I don't think we need to create a service account per component. The default service account for the namespace is the most locked down service account so that will give us the most security.

FrimIdan: To me it's more clear if every component will have it's own service account created, even if no permissions are needed. I don't like to use the `default` one.
Also, since currently we don't need to access and k8s resource we should set `automountServiceAccountToken: false` on the SA and/or pod so it will be more secure.

charts/vmclarity/templates/apiserver/_helpers.tpl
```diff
@@ -0,0 +1,22 @@
+{{/*
+Base name for the apiserver components
+*/}}
+{{- define "vmclarity.apiserver.name" -}}
+{{ include  "vmclarity.names.fullname" . }}-apiserver
```
FrimIdan, line 5: maybe we need to truncate the full name to `63-len("-apiserver")`
Since we are doing
```
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
```
but then we are adding more suffixes

charts/vmclarity/values.yaml
```diff
@@ -0,0 +1,198 @@
...
+    pullPolicy: IfNotPresent
+
+  podSecurityContext:
+    enabled: true
+    fsGroup: 1001
+
+  containerSecurityContext:
+    enabled: true
+    runAsUser: 1001
+    runAsNonRoot: true
```
FrimIdan, line 142: can we use the following?
```
    capabilities:
      drop:
        - ALL
    runAsNonRoot: true
    runAsGroup: 1001
    runAsUser: 1001
    privileged: false
    allowPrivilegeEscalation: false
    readOnlyRootFilesystem: true
```

charts/vmclarity/values.yaml
FrimIdan: I see all was added except `resources`?

charts/vmclarity/templates/apiserver/deployment.yaml
```diff
@@ -0,0 +1,75 @@
...
+      {{- end }}
+      {{- if .Values.apiserver.podSecurityContext.enabled }}
+      securityContext: {{- omit .Values.apiserver.podSecurityContext "enabled" | toYaml | nindent 8 }}
+      {{- end }}
+      {{- if .Values.apiserver.tolerations }}
+      tolerations: {{- .Values.apiserver.tolerations | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: apiserver
+          image: {{ include "vmclarity.images.image" ( dict "imageRoot" .Values.apiserver.image "global" .Values.global ) }}
```
fishkerez, line 29: can you please explain the use of "dict" here?

charts/vmclarity/templates/apiserver/service.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Service
```
fishkerez, line 2: Not sure why we need both service and headless-service for the api-server. Can you please explain?

charts/vmclarity/templates/gateway/deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+      tolerations: {{- .Values.gateway.tolerations | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: gateway
+          image: {{ include "vmclarity.images.image" ( dict "imageRoot" .Values.gateway.image "global" .Values.global ) }}
+          imagePullPolicy: {{ .Values.gateway.image.pullPolicy | quote }}
+          volumeMounts:
+            - name: gateway-config
+              mountPath: /etc/nginx/nginx.conf
+              subPath: nginx.conf
```
fishkerez, line 36: why do we need the subPath here?

charts/vmclarity/README.md
```diff
@@ -0,0 +1,68 @@
+# vmclarity
```
FrimIdan, line 1: please also add step in the pipeline that validates the charts and also makes sure readme is aligned with the changes

charts/vmclarity/templates/apiserver/deployment.yaml
sambetts-cisco: the `include` function takes two inputs, the name of the template to render, and a dictionary of the data to render the template with.

In this case the "vmclarity.images.image" template expects a dictionary with two top level keys "imageRoot" which is the image info in a standard structure and "global" which is the global part of values that might contain any overrides.

charts/vmclarity/templates/apiserver/service.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: {{ include "vmclarity.apiserver.name" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "vmclarity.apiserver.labels.standard" . | nindent 4 }}
+spec:
+  type: ClusterIP
+  ports:
+    - name: http
+      protocol: TCP
+      port: 8888
```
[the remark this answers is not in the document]
sambetts-cisco: Correct, we have two kubernetes services apiserver and apiserver-hl which both direct to the same port on the apiserver pods.

charts/vmclarity/templates/apiserver/service.yaml
sambetts-cisco: An ordinary service has a ClusterIP address which implements some rudimentary load-balancing capabilities, and exposes the service through the ClusterIP on apiserver.<namespace>.cluster.local. We use this service for internal communication between microservices. 

The headless service has no ClusterIP address, instead when you query the DNS on apiserver-hl.<namespace>.cluster.local, you get back the list of all the pod IP addresses in the response. This is then used by the API gateway to do some smart load balancing for the ingress.

charts/vmclarity/templates/gateway/deployment.yaml
sambetts-cisco: Without the subPath the configmap would be mounted as a folder with files inside it. In this case I only want to override the nginx.conf file because the /etc/nginx directory has other things inside it.

charts/vmclarity/values.yaml
```diff
@@ -0,0 +1,367 @@
...
+apiserver:
+  # -- Number of replicas for the API Server
+  replicas: 1
+
+  image:
+    # -- API Server image registry
+    registry: ghcr.io
+    # -- API Server image repositiory
+    repository: openclarity/vmclarity-apiserver
+    # -- API Server image tag (immutable tags are recommended)
```
FrimIdan, line 13: > immutable tags are recommended

Not sure if `immutable tags` are actually exist :)

Also, don't we want to put here `digest` option since I saw it in the helper file.

charts/vmclarity/templates/apiserver/deployment.yaml
sambetts-cisco: added, please see the latest commits

charts/vmclarity/templates/apiserver/deployment.yaml
```diff
@@ -0,0 +1,75 @@
...
+          livenessProbe:
+            httpGet:
+              path: /healthz/live
+              port: health
+            periodSeconds: 30
+            failureThreshold: 5
+            timeoutSeconds: 10
+          readinessProbe:
+            httpGet:
+              path: /healthz/ready
```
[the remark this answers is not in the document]
FrimIdan: do we need an issue for this?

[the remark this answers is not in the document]
sambetts-cisco: https://github.com/openclarity/vmclarity/issues/547

FrimIdan (commented):
LGTM

FrimIdan (approved):
LGTM, We probably need a task to create a release process for the chart when it will be fully operational.

PR #596, opened 2023-08-24
fix(lint): golangci-lint to test multiple modules

chrisgacsal:
## Description

The `golangci-lint` does not support monorepo type of projects with multiple Go packages which means that currently only  the root package is tested.
This change finds all the Golang packages in the repository and runs `golangci-lint` for each of them.

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[x] Other (please describe)  

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,10 @@ DOCKER_IMAGE ?= $(DOCKER_REGISTRY)/$(BINARY_NAME)
 DOCKER_TAG ?= ${VERSION}
 VMCLARITY_TOOLS_BASE ?=
 
+ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+BIN_DIR := $(ROOT_DIR)/bin
+
 # Dependency versions
-GOLANGCI_VERSION = 1.52.2
 LICENSEI_VERSION = 0.5.0
 
 # HELP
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -131,16 +133,23 @@ clean-ui:
...
+	@mkdir -p $(BIN_DIR)
+
+GOLANGCI_BIN := $(BIN_DIR)/golangci-lint
+GOLANGCI_CONFIG := $(ROOT_DIR)/.golangci.yml
+GOLANGCI_VERSION := 1.52.2
+
+bin/golangci-lint: bin/golangci-lint-$(GOLANGCI_VERSION)
+	@ln -sf golangci-lint-$(GOLANGCI_VERSION) bin/golangci-lint
+
+bin/golangci-lint-$(GOLANGCI_VERSION): $(BIN_DIR)
```
Tehsmash, line 146: For directories please use pipe target dependencies so that it doesn't cause this to re-run if the directory timestamp changes, for example:
```
bin/golangci-lint-$(GOLANGCI_VERSION): | $(BIN_DIR)
```
See Order Only dependencies here: https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html

Makefile
```diff
@@ -131,16 +133,23 @@ clean-ui:
 .PHONY: clean
 clean: clean-ui clean-backend ## Clean all build artifacts
 
-bin/golangci-lint: bin/golangci-lint-${GOLANGCI_VERSION}
-	@ln -sf golangci-lint-${GOLANGCI_VERSION} bin/golangci-lint
-bin/golangci-lint-${GOLANGCI_VERSION}:
-	@mkdir -p bin
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | bash -s -- -b ./bin/ v${GOLANGCI_VERSION}
+$(BIN_DIR):
+	@mkdir -p $(BIN_DIR)
```
Tehsmash, line 137: Need to add this as an order only dependency to all the `bin/vmclarity-...` targets above.

Makefile
```diff
@@ -157,8 +166,7 @@ fix: bin/golangci-lint ## Fix lint violations
 
 bin/licensei: bin/licensei-${LICENSEI_VERSION}
 	@ln -sf licensei-${LICENSEI_VERSION} bin/licensei
-bin/licensei-${LICENSEI_VERSION}:
-	@mkdir -p bin
+bin/licensei-${LICENSEI_VERSION}: $(BIN_DIR)
```
Tehsmash, line 169: ditto order only dependency

Tehsmash (changes requested):

Makefile
chrisgacsal: Wont fix in this PR as those are out of scope for this fix.

Makefile
chrisgacsal: Fixed.

Makefile
chrisgacsal: Fixed.

adamtagscherer (approved):

Issue #586, opened 2023-08-22
Scan Cost and Time Estimation


Tehsmash:
As a user I would like to be able to request that VMClarity estimate how long a scan or asset scan is going to take and how much it will cost to run.

The AssetScan estimation should take into account:
* the required resources on the target cloud account/cluster
* the ScanFamiliesConfiguration 
* the ScannerInstanceConfiguration

The result should include:
* a estimated cost in USD
* a estimated time
* an itemised break down of how the cost was calculated

A Scan estimation should take into account:
* the MaxParallelJobs configuration
* the scope
* the AssetScan configuration

The result should include:
* how many/which assets are going to be scanned
* a per-asset cost/time break down
* a total estimated cost in USD
* a total estimated time

In the UI I expect to be able to run an estimation while creating a ScanConfig.

- [ ] #587 
- [ ] #588
- [ ] #589
- [ ] #590
- [ ] #591
- [ ] #592

PR #603, opened 2023-08-29
Cost estimation infra + aws

fishkerez:
## Description

Resolve https://github.com/openclarity/vmclarity/issues/586

Discussion: https://github.com/openclarity/vmclarity/discussions/465

Adding the ability to post a new scan estimation request which will perform a cost estimation of a scan/assetScan according to the scan configuration (scan scope) and previous asset scans statistics.

The first commit (apiserver) Adds 2 new resources: ScanEstimation and AssetScanEstimation to the API, with corresponding API endpoints.
The corresponding rest controllers and DB tables was added too.

The second commit (orchestrator) adds 2 new reconcile controllers for these new resources.

The third commit (aws cost estimation) adds the implementation of Estimate interface in AWS provider.

## Type of Change

[ ] Bug Fix  
[* ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

- [ *] Existing issues have been referenced (where applicable)
- [* ] All code style checks pass


pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,269 @@
...
+		assetScanEstimationID, ok := assetScanEstimation.GetID()
+		if !ok {
+			logger.Warnf("Skipping due to invalid AssetScanEstimation: ID is nil: %v", assetScanEstimation)
+			continue
+		}
+		scanEstimationID, ok := assetScanEstimation.GetScanEstimationID()
+		if !ok {
+			logger.Warnf("Skipping due to invalid AssetScanEstimation: ScanEstimation.ID is nil: %v", assetScanEstimation)
+			continue
+		}
```
Tehsmash, line 116: AssetScanEstimations should be usable by themselves to estimate a single asset, ScanEstimationID is/should be optional to support this use case and no where in the asset scan estimation watcher should use it outside of informational logging.

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,269 @@
...
+	asset := &models.Asset{
+		AssetInfo:    assetScanEstimation.Asset.AssetInfo,
+		FirstSeen:    assetScanEstimation.Asset.FirstSeen,
+		Id:           &assetScanEstimation.Asset.Id,
+		LastSeen:     assetScanEstimation.Asset.LastSeen,
+		Revision:     assetScanEstimation.Asset.Revision,
+		ScansCount:   assetScanEstimation.Asset.ScansCount,
+		Summary:      assetScanEstimation.Asset.Summary,
+		TerminatedOn: assetScanEstimation.Asset.TerminatedOn,
+	}
```
Tehsmash, line 191: Why are we creating a new models.Asset here when we have assetScanEstimation.Asset $expanded? I think we can just change getLatestAssetScanStats to receive a AssetScanRelationship object instead.

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,269 @@
...
+
+	stats := w.getLatestAssetScanStats(ctx, asset)
+
+	estimation, err := w.provider.Estimate(ctx, stats, asset, assetScanEstimation.AssetScanTemplate)
+	if err != nil {
+		logger.Errorf("Failed to estimate asset scan: %v", err)
+		assetScanEstimation.State.State = utils.PointerTo(models.AssetScanEstimationStateStateFailed)
+		assetScanEstimation.State.Errors = utils.PointerTo(utils.UnwrapErrorStrings(err))
```
Tehsmash, line 199: replace with:
```
assetScanEstimation.State.Message = utils.PointerTo(err.Error())
assetScanEstimation.State.Reason = utils.PointerTo(models.AssetScanEstimationStateReasonUnexpected)
```

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,269 @@
...
+	} else {
+		logger.Infof("Asset scan estimation completed successfully. Estimation=%v", estimation)
+		assetScanEstimation.State.State = utils.PointerTo(models.AssetScanEstimationStateStateDone)
+		assetScanEstimation.Estimation = &models.Estimation{
+			Cost:          estimation.Cost,
+			Size:          estimation.Size,
+			Time:          estimation.Time,
+			CostBreakdown: estimation.CostBreakdown,
+		}
```
Tehsmash, line 208: Need to add:
```
assetScanEstimation.State.Reason = utils.PointerTo(models.AssetScanEstimationStateReasonSuccess)
```

pkg/orchestrator/assetscanestimationwatcher/watcher.go
Tehsmash: We need to check for Fatal errors here, we should only be moving directly to failed for Fatal errors, any other error should be bubbled up to the Reconciler to handle Retryable/unexpected errors which can be retried. Timeout should handle any case when we get stuck in a retry loop.

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,269 @@
...
+
+	now := time.Now()
+	abortTimedOut := now.After(transitionTimeToAbort.Add(w.abortTimeout))
+	if !abortTimedOut {
+		logger.Tracef("AssetScanEstimation in aborted state is not timed out yet. TransitionTime=%s Timeout=%s",
+			transitionTimeToAbort, w.abortTimeout)
+		return nil
+	}
+	logger.Tracef("AssetScanEstimation in aborted state is timed out. TransitionTime=%s Timeout=%s",
+		transitionTimeToAbort, w.abortTimeout)
```
Tehsmash, line 252: We don't need an abort timeout for AssetScanEstimation, we have this in AssetScan because we wait for the CLI Scanner to process aborted. For AssetScanEstimations we just need to process it instantly because its all happening in this flow.

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,386 @@
...
+// Static times from lab tests of family scan duration in seconds per GB.
+var familyScanDurationPerGBMap = map[familiestypes.FamilyType]float64{
+	familiestypes.SBOM:             4.5,
+	familiestypes.Vulnerabilities:  1, // TODO check time with no sbom scan
+	familiestypes.Secrets:          300,
+	familiestypes.Exploits:         0,
+	familiestypes.Rootkits:         0,
+	familiestypes.Misconfiguration: 1,
+	familiestypes.Malware:          360,
+}
```
Tehsmash, line 87: It might be nicer to use a time.Duration here to make this map more readable,:

```
familiestypes.SBOM:             4.5 * time.Second,
```


when we use it you can then do:
```
familyScanDurationPerGBMap[familiestypes.SBOM].Seconds() 
```
to get float64 we need to use in the calculation

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,386 @@
...
+		Cost: utils.PointerTo(float32(jobTotalCost)),
+		CostBreakdown: &[]models.CostBreakdownComponent{
+			{
+				Cost:      float32(sourceSnapshotCost),
+				Operation: string(SourceSnapshot),
+			},
+			{
+				Cost:      float32(destSnapshotCost),
+				Operation: string(DestinationSnapshot),
+			},
```
Tehsmash, line 191: I'm not sure this should be in the breakdown if its not something that is part of the operation for this asset right? Same with dataTransferCost, if we don't need to do dataTransfer it shouldn't be listed even if its "0".

Tehsmash (changes requested):

api/openapi.yaml
```diff
@@ -2407,6 +2795,185 @@ components:
...
+          nullable: true
+        state:
+          $ref: '#/components/schemas/ScanEstimationState'
+        startTime:
+          type: string
+          format: date-time
+        endTime:
+          type: string
+          format: date-time
+        ## TODO why do we need this?
```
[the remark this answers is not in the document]
fishkerez: Thanks for the clarification

api/openapi.yaml
```diff
@@ -2407,6 +2795,185 @@ components:
...
+      type: object
+      properties:
+        id:
+          type: string
+        assetIDs:
+          description: List of asset IDs to be estimated
+          type: array
+          items:
+            type: string
+          nullable: true
```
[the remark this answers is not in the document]
fishkerez: What is the implication of removing nullable? I can still set the assetIds to null, right?

pkg/orchestrator/assetscanestimationwatcher/watcher.go
fishkerez: Took that logic from AssetScan watcher. Should we fix this in there as well? (We have single assetScans as part of the CLI flow, no?)

pkg/orchestrator/assetscanestimationwatcher/watcher.go
fishkerez: We need Asset also for Estimate(), do you think we should change it there as well to receive a AssetScanRelationship object?

api/openapi.yaml
```diff
@@ -1945,6 +2300,39 @@ components:
...
+            - NothingToEstimate
+            - Success
+        lastTransitionTime:
+          type: string
+          format: date-time
+        errors:
+          type: array
+          items:
+            type: string
+          nullable: true
```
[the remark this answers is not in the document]
Tehsmash: This inconsistency is being worked on already there is a PR addressing the issue for AssetScans.

api/openapi.yaml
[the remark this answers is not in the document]
Tehsmash: If we remove nullable it means that if the API server receives:
```
{
   "assetIds": null
}
```
the validator will error because its not a valid request. Only thing you need to be aware of is that when creating a request to the API server you must initialise the array like:
```
assetIds := make([]models.AssetId)
```
and not
```
var assetIds []models.AssetId
```
because the later is "nil" and results it "null" getting sent in the request.

pkg/orchestrator/assetscanestimationwatcher/watcher.go
Tehsmash: It should be removed there too, I think that was missed as part of the "use AssetScans independently" work, I'll open a ticket for it.

Tehsmash: https://github.com/openclarity/vmclarity/issues/607

api/openapi.yaml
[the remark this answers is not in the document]
fishkerez: I that case, every time we Patch, will that patch the assetIds to empty list? or does it knows to ignore it?

pkg/orchestrator/assetscanestimationwatcher/watcher.go
Tehsmash: Hmm thats a great point, I assume we can't cleanly cast between Asset and AssetRelationship? Perhaps AssetRelationship needs a convert helper function `ToAsset()` or something. We can use this library
https://github.com/anchore/go-struct-converter to remove the need to manually map all the fields in the struct.

api/openapi.yaml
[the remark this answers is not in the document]
Tehsmash: AssetIds is a pointer in the struct. If you set the pointer to something then it will result it the PATCH seeing it, as long as you don't set the Assetids pointer to something omitempty will ignore it from the request and it will not get processed. For example:

This will not replace assetIds:
```
ase := models.AssetScanEstimation{
    EndTime: time.Now()
}
backendclient.PatchAssetScanEstimation(ctx, id, ase)
```
json looks like:
```
{
   "EndTime": "123123123"
}
```


This will replace assetIds:
```
assetIds := make([]models.AssetId)
ase := models.AssetScanEstimation{
    AssetIds: &assetIds,
}
backendclient.PatchAssetScanEstimation(ctx, id, ase)
```
json looks like:
```
{
   "assetIds": []
}
```

This is invalid:
```
var assetIds []models.AssetId 
ase := models.AssetScanEstimation{
    AssetIds: &assetIds,
}
backendclient.PatchAssetScanEstimation(ctx, id, ase)
```
json looks like:
```
{
   "assetIds": null
}
```

pkg/orchestrator/provider/aws/scanestimation/estimate.go
fishkerez: We can't use Duration if we want the values here to be float64.

pkg/orchestrator/assetscanestimationwatcher/watcher.go
fishkerez: Yes, we need some general convert func between these two, in order not to manually convert every field. Maybe we can also use json marshal/unmarshal

pkg/orchestrator/provider/aws/scanestimation/estimate.go
Tehsmash: I'm suggesting that we change this map to time.Duration for readability and then use the ".Seconds()" function when we need the seconds in float64 in the calculations.

fishkerez: But in that case, we can't save 4.5 seconds value in the map, right?

api/openapi.yaml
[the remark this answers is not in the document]
fishkerez: So in order to check if assetScanEstimation had any errors, we will check the state to be Failed?

pkg/orchestrator/provider/aws/scanestimation/estimate.go
Tehsmash: time.Durations are an int64 which represents nanoseconds. 4.5 seconds == 4500000000 nanoseconds. You can create a duration from a float like:
```
familiestypes.SBOM: time.Duration(4.5 * float64(time.Second)),
```
the others aren't as complicated:
```
familiestypes.Malware: 6 * time.Minute,
```

api/openapi.yaml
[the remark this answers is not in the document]
Tehsmash: Correct, the state should be Failed if any error occurs during AssetScanEstimation.

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
```diff
@@ -0,0 +1,331 @@
...
+	"me-south-1":     "MES1", // Middle East (Bahrain)
+	"sa-east-1":      "SAE1", // São Paulo
+	"us-gov-west-1":  "UGW1", // AWS GovCloud (US-West)
+	"us-gov-east-1":  "UGE1", // AWS GovCloud (US-East)
+	//"il-central-1": "EUS2", // Tel-Aviv // TODO couldn't find the abbreviation code
+
+}
+
+const (
+	usEast1RegionCode = "us-east-1"
```
Tehsmash, line 75: Can you explain why we're special casing us-east-1, and why in getPricePerUnit we're talking to the us-east-1 region AWS endpoint and not the region where VMClarity is deployed/the client's native region?

fishkerez: > Can you explain why we're special casing us-east-1
From the [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/aws-usage-report-understand.html), US East (N. Virginia) usage type abbreviation will be USE1, or no prefix. But from my tests, only no abbreviation worked, and if I added USE1 it did not find the product file.

> why in getPricePerUnit we're talking to the us-east-1 region AWS endpoint and not the region where VMClarity is deployed/the client's native region?

The products Offer file is general for all region and the endpoint to download it is located in:
`https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/us-east-1/index.json`
See https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/using-ppslong.html

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
```diff
@@ -0,0 +1,331 @@
...
+	if regionCode == usEast1RegionCode {
+		usageType = "EBS:SnapshotUsage"
+	} else {
+		abb, ok := regionCodeToUsageTypeAbbreviation[regionCode]
+		if !ok {
+			return 0, fmt.Errorf("failed to find usage type in regionCodeToUsageTypeAbbreviation map with regionCode %v", regionCode)
+		}
+		// example: USE2-EBS:SnapshotUsage
+		usageType = fmt.Sprintf("%v-EBS:SnapshotUsage", abb)
+	}
```
Tehsmash, line 89: maybe better to make this a helper function, `GetRegionalUsageType(regionCode, usageType) (string, error)` as we're doing it in multiple places.

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
Tehsmash: Would you mind adding this as a comment somewhere, its a pretty odd scenario for an Amazon API endpoint to only be available in one region.

api/openapi.yaml
```diff
@@ -905,6 +905,343 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AssetScanEstimationExists'
+        default:
+          $ref: '#/components/responses/UnknownError'
+      x-codegen-request-body-name: body
+
+  /assetScanEstimations/{assetScanEstimationID}:
+    get:
+      summary: Get an asset scan stimation.
```
akpsgit, line 1131: estimation

api/openapi.yaml
```diff
@@ -1945,6 +2282,31 @@ components:
             type: string
           nullable: true
 
+    AssetScanEstimationState:
+      type: object
+      properties:
+        state:
```
akpsgit, line 2288: Maybe worth adding a description, saying that in order to initiate an estimation, need to create it in pending state

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
```diff
@@ -0,0 +1,331 @@
...
+func (o *PriceFetcherImpl) getPricePerUnit(ctx context.Context, usageType, operation string) (float64, error) {
+	filters := createGetProductsFilters(usageType, "AmazonEC2", operation)
+
+	products, err := o.pricingClient.GetProducts(ctx, &pricing.GetProductsInput{
+		ServiceCode:   utils.PointerTo("AmazonEC2"),
+		Filters:       filters,
+		FormatVersion: utils.PointerTo("aws_v1"),
+	}, func(options *pricing.Options) {
+		options.Region = usEast1RegionCode
+	})
```
Tehsmash, line 187: Do you know if this uses the bulk product API or the product query API. https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/using-pelong.html it looks like it might be the query API which has regional variants for the US, Europe and Asia.

api/openapi.yaml
Tehsmash: I think its worth handling this in the AssetScanEstimation watcher. If it sees an AssetScanEstimation with no state, set it to Pending, PATCH it in the DB and then continue to process it as though it was in Pending.

api/openapi.yaml
```diff
@@ -2407,6 +2769,183 @@ components:
...
+          items:
+            $ref: '#/components/schemas/AssetScanEstimationRelationship'
+        scanTemplate:
+          $ref: '#/components/schemas/ScanTemplate'
+        summary:
+          $ref: '#/components/schemas/ScanEstimationSummary'
+        revision:
+          type: integer
+
+    ## TODO fill out the rest of the fields for the ScanEstimationRelationship.
```
akpsgit, line 2814: what is the impact of not having the rest filled?

api/openapi.yaml
```diff
@@ -2407,6 +2769,183 @@ components:
...
+        revision:
+          type: integer
+        startTime:
+          type: string
+          format: date-time
+        endTime:
+          type: string
+          format: date-time
+
+    ## TODO fill out the rest of the fields for the AssetScanEstimationRelationship.
```
akpsgit, line 2913: same

api/openapi.yaml
```diff
@@ -2407,6 +2769,183 @@ components:
...
+      description: A summary of the AssetScanEstimations under this ScanEstimation
+      properties:
+        jobsLeftToRun:
+          type: integer
+        jobsCompleted:
+          type: integer
+        totalScanTime:
+          description: Total estimated time the scan will take (seconds)
+          type: integer
+        totalScanSize:
```
akpsgit, line 2941: same

pkg/orchestrator/assetscanestimationwatcher/watcher.go
fishkerez: It is an Info log, just for informational purposes

pkg/apiserver/database/gorm/odata.go
```diff
@@ -912,6 +912,125 @@ var schemaMetas = map[string]odatasql.SchemaMeta{
...
+			"id":        odatasql.FieldMeta{FieldType: odatasql.PrimitiveFieldType},
+			"startTime": odatasql.FieldMeta{FieldType: odatasql.PrimitiveFieldType},
+			"endTime":   odatasql.FieldMeta{FieldType: odatasql.PrimitiveFieldType},
+			"assetIDs": odatasql.FieldMeta{
+				FieldType: odatasql.CollectionFieldType,
+				CollectionItemMeta: &odatasql.FieldMeta{
+					FieldType: odatasql.PrimitiveFieldType,
+				},
+			},
+			"state": odatasql.FieldMeta{
```
akpsgit, line 927: this is not primitive

pkg/apiserver/database/gorm/odata.go
```diff
@@ -912,6 +912,125 @@ var schemaMetas = map[string]odatasql.SchemaMeta{
...
+				FieldType: odatasql.PrimitiveFieldType,
+			},
+			"stateMessage": odatasql.FieldMeta{
+				FieldType: odatasql.PrimitiveFieldType,
+			},
+			"stateReason": odatasql.FieldMeta{
+				FieldType: odatasql.PrimitiveFieldType,
+			},
```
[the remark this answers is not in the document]
fishkerez: Did it in the API, forgot to change it here

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,261 @@
...
+		logger.Errorf("Failed to estimate asset scan: %v", err)
+		assetScanEstimation.State.State = utils.PointerTo(models.AssetScanEstimationStateStateFailed)
+		assetScanEstimation.State.StateMessage = utils.PointerTo(fatalError.Error())
+		assetScanEstimation.State.StateReason = utils.PointerTo(models.AssetScanEstimationStateStateReasonUnexpected)
+	case errors.As(err, &retryableError):
+		// nolint:wrapcheck
+		return common.NewRequeueAfterError(retryableError.RetryAfter(), retryableError.Error())
+	case err != nil:
+		logger.Errorf("Failed to estimate asset scan: %v", err)
+		assetScanEstimation.State.State = utils.PointerTo(models.AssetScanEstimationStateStateFailed)
```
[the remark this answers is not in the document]
fishkerez: Do you think we should treat this same as retryableError?

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,261 @@
...
+		return errors.New("invalid AssetScanEstimation: ID is nil")
+	}
+
+	if assetScanEstimation.State == nil || assetScanEstimation.State.State == nil {
+		return errors.New("invalid AssetScanEstimation: State is nil")
+	}
+
+	now := time.Now()
+
+	assetScanEstimation.State.State = utils.PointerTo(models.AssetScanEstimationStateStateDone)
```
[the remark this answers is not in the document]
fishkerez: Took this logic from AssetScan watcher. Need to open an issue to fix this in there?

pkg/orchestrator/assetscanestimationwatcher/watcher.go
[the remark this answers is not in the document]
fishkerez: I saw in AssetScan watcher reconcileScheduled func that we are treating an error as fatal error and just marking it as done. Do we need to fix this in there as well?

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,404 @@
...
+	scannerCost := scannerPerHourCost * ((float64(jobCreationTimeSec + scanDurationSec)) / SecondsInAnHour)
+	scannerRootVolumeCost := scannerRootVolumeMonthlyCost * ((float64(jobCreationTimeSec + scanDurationSec)) / SecondsInAMonth) * float64(scannerRootVolumeSizeGB)
+
+	jobTotalCost := sourceSnapshotCost + volumeFromSnapshotCost + scannerCost + scannerRootVolumeCost + dataTransferCost + destSnapshotCost
+
+	// Create the Estimation object base on the calculated data.
+	costBreakdown := []models.CostBreakdownComponent{
+		{
+			Cost:      float32(destSnapshotCost),
+			Operation: string(DestinationSnapshot),
```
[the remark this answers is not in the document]
fishkerez: Since we always have 2 regions that we are moving data between, I think source and dest are understandable enough. 
But I don't mind changing to what you suggested if you think it is more clear

pkg/orchestrator/assetscanestimationwatcher/helpers.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// getLatestAssetScanStats - for each family, find the latest AssetScan that has AssetScanStats of this family,
+// and add the family stats to the returned aggregated AssetScanStats.
+func (w *Watcher) getLatestAssetScanStats(ctx context.Context, asset *models.Asset) models.AssetScanStats {
+	var stats models.AssetScanStats
+
+	filterTmpl := "asset/id eq '%s' and status/general/state eq 'Done' and status/%s/errors eq null and scanFamiliesConfig/%s/enabled eq true"
+
+	families := []string{"exploits", "sbom", "vulnerabilities", "malware", "misconfigurations", "rootkits", "secrets"}
```
akpsgit, line 35: Need to create a task to do estimation also for the infoFinder

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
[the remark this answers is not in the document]
fishkerez: I agree caching sounds like a great idea.

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,404 @@
...
+			totalScanDuration += scanDuration
+		} else {
+			totalScanDuration += calculateStaticScanDuration(familiestypes.Malware, scanSizeGB)
+		}
+	}
+
+	return totalScanDuration, nil
+}
+
+const MillisecondsInSecond = 1000 * 1000 * 1000
```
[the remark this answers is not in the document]
fishkerez: That makes much more sense :)

api/openapi.yaml
fishkerez: Should we do this for scanEstimation as well?

pkg/orchestrator/assetscanestimationwatcher/watcher.go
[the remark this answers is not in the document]
Tehsmash: IMO yes it should be fixed there too, there is suppose to be three error cases:

* Fatal error - stop now, we know this isn't something a retry will fix

* Retryable/Requeue error - a known case were we need to re-run reconciliation, the raiser can specify how long to wait until its requeue. (IMO we should change the name of this, I don't like "Retryable" its confusing and doesn't fix the use cases its designed for such as waiting for a snapshot to be ready, perhaps RetryAfter or ReconcileAfter @chrisgacsal WDYT?)

* Normal error - an error occurred, it will be logged we'll retry after the next polling interval. We'll keep retrying until the timeout is hit.

pkg/orchestrator/provider/aws/scanestimation/estimate.go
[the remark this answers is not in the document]
Tehsmash: I think its more understandable from the users perspective, they don't know/care about the flow inside the provider and how we move snapshots around, they only care about the resources that were created, and we created two snapshots one in region X and one in region Y.

api/openapi.yaml
Tehsmash: If you do `GET assetScanEstimation/<uuid>?$expand=ScanEstimation` then you will only get the ScanEstimation fields specified here and nothing else.

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
Tehsmash:
```
usageType, err = GetRegionUsageType("us-east-1", "EBS:SnapshotUsage")
if err != nil {
   return 0, err
}
fmt.Println(usageType)

EBS:SnapshotUsage
```

```
usageType, err = GetRegionUsageType("us-east-2", "EBS:SnapshotUsage")
if err != nil {
   return 0, err
}
fmt.Println(usageType)

USE2-EBS:SnapshotUsage
```

```
usageType, err = GetRegionUsageType("us-east-2", fmt.Sprintf("EBS:VolumeUsage.%v", volumeType))
if err != nil {
   return 0, err
}
fmt.Println(usageType)

USE2-EBS:VolumeUsage.spicy
```

api/openapi.yaml
Tehsmash: I think so, eventually we should do it for AssetScan/Scan objects too I think.

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,574 @@
...
+	state, ok := result.GetState()
+	if !ok {
+		return fmt.Errorf("state must not be nil for AssetScan. AssetScanID=%s", *result.Id)
+	}
+
+	s := scanEstimation.Summary
+
+	switch state {
+	case models.AssetScanEstimationStateStatePending:
+		s.JobsLeftToRun = utils.PointerTo(*s.JobsLeftToRun + 1)
```
akpsgit, line 387: do you assume to call this function on every asset scan once? Because if this reconcile function is called several times for the same asset scan estimation and the same scan estimation it will go out of sync.

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
Tehsmash: Looks like the aws-sdk-v2 is using the query APIs not the bulk ones: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/pricing#pkg-overview

Only the US and Asia endpoints are supported though:
> You can use the following endpoints for the Amazon Web Services Price List API:
> https://api.pricing.us-east-1.amazonaws.com/
> https://api.pricing.ap-south-1.amazonaws.com/

Lets add a comment that just highlight that the Pricing API is only available on us-east-1 and ap-south-1, and for now we've chosen to use the us-east-1 endpoint.

pkg/shared/utils/utils_test.go
```diff
@@ -218,3 +218,43 @@ func TestStringKeyMapToArray(t *testing.T) {
...
+		},
+		{
+			name: "not nil",
+			args: args{
+				val: PointerTo(5),
+			},
+			want: 5,
+		},
+		{
+			name: "not nit 0",
```
akpsgit, line 246: nil

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,574 @@
...
+
+	if scanEstimation.AssetScanEstimations == nil {
+		scanEstimation.AssetScanEstimations = &[]models.AssetScanEstimationRelationship{}
+	}
+	*scanEstimation.AssetScanEstimations = append(*scanEstimation.AssetScanEstimations, models.AssetScanEstimationRelationship{Id: ret.Id})
+
+	return nil
+}
+
+func updateScanEstimationSummaryFromAssetScanEstimation(scanEstimation *models.ScanEstimation, result models.AssetScanEstimation) error {
```
akpsgit, line 377: maybe better: result-> assetScanEstimation

pkg/orchestrator/provider/aws/client.go
```diff
@@ -69,6 +73,48 @@ func (c Client) Kind() models.CloudProvider {
 	return models.AWS
 }
 
+func (c *Client) Estimate(ctx context.Context, stats models.AssetScanStats, asset *models.Asset, assetScanTemplate *models.AssetScanTemplate) (*models.Estimation, error) {
```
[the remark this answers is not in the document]
fishkerez: So maybe assetScanStats?

pkg/orchestrator/provider/aws/scanestimation/estimate.go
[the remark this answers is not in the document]
akpsgit: +1
(as far as i know there is 1000 :P)

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
```diff
@@ -0,0 +1,331 @@
...
+			return 0, fmt.Errorf("failed to find usage type in regionCodeToUsageTypeAbbreviation map with regionCode %v", regionCode)
+		}
+		// example: USE2-EBS:SnapshotUsage
+		usageType = fmt.Sprintf("%v-EBS:SnapshotUsage", abb)
+	}
+
+	// TODO do we want to return estimated price in case of an error base on current pricePerUnit?
+	price, err := o.getPricePerUnit(ctx, usageType, "")
+	if err != nil {
+		return 0.05, nil
```
akpsgit, line 94: can you please add a comment regarding where 0.05 was obtained?

fishkerez: Sure. Basically I tested all the regions, and for some reason the only region that I failed to find the snapshot price was Ireland. So this is the standard snapshot price in Ireland:
https://aws.amazon.com/ebs/pricing/

api/openapi.yaml
```diff
@@ -1945,6 +2300,39 @@ components:
...
+          description: Human-readable message indicating details about the last state transition.
+          type: string
+        stateReason:
+          description: Machine-readable, UpperCamelCase text indicating the reason for the condition's last transition.
+          type: string
+          enum:
+            - Aborted
+            - TimedOut
+            - OneOrMoreAssetFailedToEstimate
+            - DiscoveryFailed
```
FrimIdan: `DiscoveryFailed` is related here?

pkg/orchestrator/provider/aws/scanestimation/estimate.go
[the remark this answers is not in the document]
akpsgit: @Tehsmash, o.k to resolve?

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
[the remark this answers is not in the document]
akpsgit: created https://github.com/openclarity/vmclarity/issues/656

pkg/orchestrator/assetscanestimationwatcher/helpers.go
akpsgit: https://github.com/openclarity/vmclarity/issues/657

api/openapi.yaml
akpsgit: not relevant and not used. can be removed

pkg/apiserver/database/gorm/odata.go
```diff
@@ -912,6 +912,127 @@ var schemaMetas = map[string]odatasql.SchemaMeta{
...
+	},
+	"ScanEstimationState": {
+		Fields: odatasql.Schema{
+			"state": odatasql.FieldMeta{FieldType: odatasql.PrimitiveFieldType},
+			"stateMessage": odatasql.FieldMeta{
+				FieldType: odatasql.PrimitiveFieldType,
+			},
+			"stateReason": odatasql.FieldMeta{
+				FieldType: odatasql.PrimitiveFieldType,
+			},
```
Tehsmash, line 1033: Missing last transition time here

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,281 @@
...
+		logger.Errorf("Failed to estimate asset scan: %v", err)
+		assetScanEstimation.State.State = utils.PointerTo(models.AssetScanEstimationStateStateFailed)
+		assetScanEstimation.State.StateMessage = utils.PointerTo(fatalError.Error())
+		assetScanEstimation.State.StateReason = utils.PointerTo(models.AssetScanEstimationStateStateReasonUnexpected)
+	case errors.As(err, &retryableError):
+		// nolint:wrapcheck
+		return common.NewRequeueAfterError(retryableError.RetryAfter(), retryableError.Error())
+	case err != nil:
+		logger.Errorf("Failed to estimate asset scan. Retrying after %v: %v", w.pollPeriod, err)
+		return common.NewRequeueAfterError(w.pollPeriod, err.Error())
```
Tehsmash, line 224: This should just be a normal error, and the Reconciler code will log it otherwise we'll be holding this in memory unnecessarily.

Tehsmash (changes requested):
Two minor issues, and the code is in conflict. I'll address these issues and push a new revision in the morning, and then I think we can merge.

fishkerez:
Thanks @Tehsmash please notice that I have rebased and pushed, so please do pull --rebase before pushing

api/models/assetrelationship.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func (a *AssetRelationship) ToAsset() (*Asset, error) {
+	if a == nil {
+		return nil, fmt.Errorf("asset relationship is nil")
+	}
+
+	B, err := json.Marshal(a)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal: %v", err)
```
chrisgacsal, line 30:
```suggestion
		return nil, fmt.Errorf("failed to marshal: %w", err)
```

api/models/assetrelationship.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	var asset Asset
+	if err := json.Unmarshal(B, &asset); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal: %v", err)
```
chrisgacsal, line 35:
```suggestion
		return nil, fmt.Errorf("failed to unmarshal: %w", err)
```

pkg/apiserver/rest/asset_scan_estimation_controller.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/openclarity/vmclarity/api/models"
+	"github.com/openclarity/vmclarity/pkg/apiserver/common"
+	databaseTypes "github.com/openclarity/vmclarity/pkg/apiserver/database/types"
+	"github.com/openclarity/vmclarity/pkg/shared/utils"
+)
+
+func (s *ServerImpl) GetAssetScanEstimations(ctx echo.Context, params models.GetAssetScanEstimationsParams) error {
+	dbAssetScanEstimations, err := s.dbHandler.AssetScanEstimationsTable().GetAssetScanEstimations(params)
+	if err != nil {
+		return sendError(ctx, http.StatusInternalServerError, fmt.Sprintf("failed to get asset scan estimations results from db: %v", err))
```
chrisgacsal, line 35: wdyt about using this instead?

```suggestion
		return sendError(ctx, http.StatusInternalServerError, fmt.Errorf("failed to get asset scan estimations results from db: %w", err).Error())
```

But honestly the `sendError` should accept `error` instead of `string` as 3rd parameter. But this is definitely unrelated to this PR.

pkg/cli/presenter/vmclarity.go
```diff
@@ -88,7 +88,7 @@ func (v *VMClarityPresenter) ExportSbomResult(ctx context.Context, res families.
 		assetScan.Stats = &models.AssetScanStats{}
 	}
 
-	errs := []string{}
+	var errs []string
```
chrisgacsal, line 91: Would be nice to have a separate PR for this fix.

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,282 @@
...
+		return errors.New("invalid AssetScanEstimation: ID is nil")
+	}
+
+	if assetScanEstimation.Asset == nil || assetScanEstimation.Asset.AssetInfo == nil {
+		return errors.New("invalid AssetScanEstimation: Asset or AssetInfo is nil")
+	}
+
+	asset, err := assetScanEstimation.Asset.ToAsset()
+	if err != nil {
+		return fmt.Errorf("failed to convert assetRelationship to asset: %v", err)
```
chrisgacsal, line 206:
```suggestion
		return fmt.Errorf("failed to convert assetRelationship to asset: %w", err)
```

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,282 @@
...
+	case errors.As(err, &fatalError):
+		logger.Errorf("Failed to estimate asset scan: %v", err)
+		assetScanEstimation.State.State = utils.PointerTo(models.AssetScanEstimationStateStateFailed)
+		assetScanEstimation.State.StateMessage = utils.PointerTo(fatalError.Error())
+		assetScanEstimation.State.StateReason = utils.PointerTo(models.AssetScanEstimationStateStateReasonUnexpected)
+	case errors.As(err, &retryableError):
+		// nolint:wrapcheck
+		return common.NewRequeueAfterError(retryableError.RetryAfter(), retryableError.Error())
+	case err != nil:
+		return fmt.Errorf("failed to estimate asset scan. Retrying after %v: %v", w.pollPeriod, err)
```
chrisgacsal:
```suggestion
		return fmt.Errorf("failed to estimate asset scan. Retrying after %v: %w", w.pollPeriod, err)
```

pkg/orchestrator/provider/aws/client.go
```diff
@@ -69,6 +73,49 @@ func (c Client) Kind() models.CloudProvider {
 	return models.AWS
 }
 
+func (c *Client) Estimate(ctx context.Context, assetScanStats models.AssetScanStats, asset *models.Asset, assetScanTemplate *models.AssetScanTemplate) (*models.Estimation, error) {
+	var err error
+	const jobCreationTimeConst = 2
+
+	vminfo, err := asset.AssetInfo.AsVMInfo()
+	if err != nil {
+		return nil, fmt.Errorf("failed to use asset info as vminfo: %v", err)
```
chrisgacsal, line 82:
```suggestion
		return nil, fmt.Errorf("failed to convert AssetInfo into VMInfo: %w", err)
```

pkg/orchestrator/provider/aws/client.go
```diff
@@ -69,6 +73,49 @@ func (c Client) Kind() models.CloudProvider {
...
+	}
+
+	location, err := NewLocation(vminfo.Location)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse location %v: %v", vminfo.Location, err)
```
chrisgacsal, line 87:
```suggestion
		return nil, fmt.Errorf("failed to parse location %v: %w", vminfo.Location, err)
```

pkg/orchestrator/provider/aws/client.go
```diff
@@ -69,6 +73,49 @@ func (c Client) Kind() models.CloudProvider {
...
+		ScannerInstanceType:     ec2types.InstanceType(scannerInstanceType),
+		JobCreationTimeSec:      int64(jobCreationTimeSec),
+		ScannerRootVolumeSizeGB: int64(scannerRootVolumeSizeGB),
+		Stats:                   assetScanStats,
+		Asset:                   asset,
+		AssetScanTemplate:       assetScanTemplate,
+	}
+	ret, err := c.scanEstimator.EstimateAssetScan(ctx, params)
+	if err != nil {
+		return nil, fmt.Errorf("failed to estimate asset scan: %v", err)
```
chrisgacsal, line 113:
```suggestion
		return nil, fmt.Errorf("failed to estimate asset scan: %w", err)
```

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,406 @@
...
+// We are not taking into account Reserved Instances (RIs) or Saving Plans (SPs) since we don't know the exact OnDemand configuration in order to launch them.
+// In the future, we can let the user choose to use RI's or SP's as the scanner instances.
+
+// nolint:cyclop
+func (s *ScanEstimator) EstimateAssetScan(ctx context.Context, params EstimateAssetScanParams) (*models.Estimation, error) {
+	var sourceSnapshotMonthlyCost float64
+	var err error
+
+	if params.AssetScanTemplate == nil || params.AssetScanTemplate.ScanFamiliesConfig == nil {
+		return nil, fmt.Errorf("scan families config was not provided")
```
chrisgacsal, line 106:
```suggestion
		return nil, errors.New("scan families config was not provided")
```

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+	familiesConfig := params.AssetScanTemplate.ScanFamiliesConfig
+
+	// Get scan size and scan duration using previous stats and asset info.
+	scanSizeMB, err := getScanSize(params.Stats, params.Asset)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get scan size: %v", err)
```
chrisgacsal, line 113:
```suggestion
		return nil, fmt.Errorf("failed to get scan size: %w", err)
```

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,406 @@
...
+	jobCreationTimeSec := params.JobCreationTimeSec
+	scannerInstanceType := params.ScannerInstanceType
+	scannerRootVolumeSizeGB := params.ScannerRootVolumeSizeGB
+	scannerVolumeType := params.ScannerVolumeType
+
+	// Get relevant current prices from AWS price list API
+	// Fetch the dest snapshot monthly cost.
+	destSnapshotMonthlyCost, err := s.priceFetcher.GetSnapshotMonthlyCostPerGB(ctx, destRegion)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get dest snapshot monthly cost: %v", err)
```
chrisgacsal, line 135:
```suggestion
		return nil, fmt.Errorf("failed to get monthly cost for destination snapshot : %w", err)
```

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	// Fetch the scanner job monthly cost.
+	scannerPerHourCost, err := s.priceFetcher.GetInstancePerHourCost(ctx, destRegion, scannerInstanceType, marketOption)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get scanner per hour cost: %v", err)
```
chrisgacsal, line 141:
```suggestion
		return nil, fmt.Errorf("failed to get scanner per hour cost: %w", err)
```

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	// Fetch the scanner job volume monthly cost.
+	scannerRootVolumeMonthlyCost, err := s.priceFetcher.GetVolumeMonthlyCostPerGB(ctx, destRegion, scannerVolumeType)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get volume monthly cost per GB: %v", err)
```
chrisgacsal, line 147:
```suggestion
		return nil, fmt.Errorf("failed to get volume monthly cost per GB: %w", err)
```

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	// Fetch the monthly cost of the volume that was created from the snapshot.
+	volumeFromSnapshotMonthlyCost, err := s.priceFetcher.GetVolumeMonthlyCostPerGB(ctx, destRegion, fromSnapshotVolumeType)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get volume monthly cost per GB: %v", err)
```
chrisgacsal, line 153:
```suggestion
		return nil, fmt.Errorf("failed to get volume monthly cost per GB: %w", err)
```

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	dataTransferCost := 0.0
+	sourceSnapshotCost := 0.0
+	if sourceRegion != destRegion {
+		// if the scanner is in a different region then the scanned asset, we have another snapshot created in the
+		// source region.
+		sourceSnapshotMonthlyCost, err = s.priceFetcher.GetSnapshotMonthlyCostPerGB(ctx, sourceRegion)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get source snapshot monthly cost: %v", err)
```
chrisgacsal, line 163:
```suggestion
			return nil, fmt.Errorf("failed to get source snapshot monthly cost: %w", err)
```

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -0,0 +1,406 @@
...
+		}
+
+		sourceSnapshotCost = sourceSnapshotMonthlyCost * ((float64(jobCreationTimeSec + scanDurationSec)) / SecondsInAMonth) * scanSizeGB
+
+		// Fetch the data transfer cost per GB (if source and dest regions are the same, this will be 0).
+		dataTransferCostPerGB, err := s.priceFetcher.GetDataTransferCostPerGB(sourceRegion, destRegion)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get data transfer cost per GB: %v", err)
```
chrisgacsal, line 171:
```suggestion
			return nil, fmt.Errorf("failed to get data transfer cost per GB: %w", err)
```

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
```diff
@@ -0,0 +1,343 @@
...
+				Values: []string{"Linux/UNIX"},
+			},
+		},
+		InstanceTypes: []ec2types.InstanceType{instanceType},
+		StartTime:     utils.PointerTo(timeNow.Add(-12 * time.Hour)), // last 12 hours should be enough as we only want the latest.
+	}, func(options *ec2.Options) {
+		options.Region = regionCode
+	})
+	if err != nil {
+		return 0, fmt.Errorf("failed to describe spot price history: %v", err)
```
chrisgacsal, line 167:
```suggestion
		return 0, fmt.Errorf("failed to describe spot price history: %w", err)
```

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
```diff
@@ -0,0 +1,343 @@
...
+	if len(ret.SpotPriceHistory) == 0 {
+		return 0, fmt.Errorf("failed to find spot instances history")
+	}
+
+	// take the latest price (most updated)
+	spotPrice := ret.SpotPriceHistory[0].SpotPrice
+
+	price, err := strconv.ParseFloat(*spotPrice, 64)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse float %v: %v", *spotPrice, err)
```
chrisgacsal, line 179:
```suggestion
		return 0, fmt.Errorf("failed to parse float %v: %w", *spotPrice, err)
```

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
```diff
@@ -0,0 +1,343 @@
...
+	products, err := o.pricingClient.GetProducts(ctx, &pricing.GetProductsInput{
+		ServiceCode:   utils.PointerTo("AmazonEC2"),
+		Filters:       filters,
+		FormatVersion: utils.PointerTo("aws_v1"),
+	}, func(options *pricing.Options) {
+		// the Pricing API is only available on us-east-1 and ap-south-1. for now, we've chosen to use the us-east-1 endpoint.
+		options.Region = usEast1RegionCode
+	})
+	if err != nil {
+		return 0, fmt.Errorf("failed to get products. usageType=%v: %v", usageType, err)
```
chrisgacsal, line 197:
```suggestion
		return 0, fmt.Errorf("failed to get products. usageType=%v: %w", usageType, err)
```

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+	if len(products.PriceList) != 1 {
+		return 0, fmt.Errorf("excpecting exactly one product in price list, got %v", len(products.PriceList))
+	}
+
+	priceStr, err := getPricePerUnitFromJSONPriceList(products.PriceList[0])
+	if err != nil {
+		return 0, fmt.Errorf("failed to get pricePerUnit from json price list: %v", err)
```
chrisgacsal, line 205:
```suggestion
		return 0, fmt.Errorf("failed to get pricePerUnit from json price list: %w", err)
```

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	price, err := strconv.ParseFloat(priceStr, 64)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse float %v: %v", priceStr, err)
```
chrisgacsal, line 210:
```suggestion
		return 0, fmt.Errorf("failed to parse float %v: %w", priceStr, err)
```

pkg/orchestrator/provider/aws/scanestimation/pricefetcher.go
```diff
@@ -0,0 +1,343 @@
...
+  "version": "20230817222016",
+  "publicationDate": "2023-08-17T22:20:16Z"
+}
+More details on: https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/procedures.html
+*/
+func getPricePerUnitFromJSONPriceList(jsonPriceList string) (string, error) {
+	var productMap map[string]any
+	err := json.Unmarshal([]byte(jsonPriceList), &productMap)
+	if err != nil {
+		return "", fmt.Errorf("failed to unmarshal. jsonPriceList=%s:  %v", jsonPriceList, err)
```
chrisgacsal, line 293:
```suggestion
		return "", fmt.Errorf("failed to unmarshal. jsonPriceList=%s:  %w", jsonPriceList, err)
```

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,596 @@
...
+	filter := fmt.Sprintf("state/state ne '%s' and state/state ne '%s'", models.ScanEstimationStateStateDone, models.ScanEstimationStateStateFailed)
+	selector := "id"
+	params := models.GetScanEstimationsParams{
+		Filter: &filter,
+		Select: &selector,
+		Count:  utils.PointerTo(true),
+	}
+	scanEstimations, err := w.backend.GetScanEstimations(ctx, params)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get running sScanEstimations: %v", err)
```
chrisgacsal, line 92:
```suggestion
		return nil, fmt.Errorf("failed to get running sScanEstimations: %w", err)
```

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -0,0 +1,596 @@
...
+	var wg sync.WaitGroup
+	for _, id := range *scanEstimation.AssetIDs {
+		wg.Add(1)
+		assetID := id
+		go func() {
+			defer wg.Done()
+
+			err := w.createAssetScanEstimationForAsset(ctx, scanEstimation, assetID)
+			if err != nil {
+				logger.WithField("AssetID", assetID).Errorf("Failed to create AssetScanEstimation: %v", err)
```
chrisgacsal, line 320:
```suggestion
				logger.WithField("AssetID", assetID).Errorf("Failed to create AssetScanEstimation: %w", err)
```

pkg/shared/utils/utils.go
```diff
@@ -71,6 +71,20 @@ func Int32PointerValOrEmpty(val *int32) int32 {
 	return *val
 }
 
+func Float32PointerValOrEmpty(val *float32) float32 {
```
chrisgacsal, line 74: I'd prefer to use generic here if possible. Something like this: https://go.dev/play/p/6p596MSAHhD?v=goprev

chrisgacsal (changes requested):

pkg/cli/presenter/vmclarity.go
Tehsmash: this fix shouldn't be in this PR, it was pulled out into a separate PR and so far has been rejected because of the nullable issues. I'll remove this.

api/models/assetrelationship.go
Tehsmash: I wonder if there is a linter which can catch these things, we should be wrapping everywhere now except in specific instances

pkg/apiserver/rest/asset_scan_estimation_controller.go
Tehsmash: Something I've had in my mind for a long time is switching our code generator to "strict mode". https://github.com/deepmap/oapi-codegen/tree/master#strict-server-generation

In strict mode the rest controllers have to return a specific response object. For example:

```
func (p *PetStore) FindPetByID(ctx context.Context, request FindPetByIDRequestObject) (FindPetByIDResponseObject, error) {
```

```
return FindPetByIDdefaultJSONResponse{StatusCode: http.StatusNotFound, Body: Error{Code: http.StatusNotFound, Message: fmt.Sprintf("Could not find pet with ID %d", request.Id)}}, nil
```

In this way we wouldn't have to have nasty functions like sendError etc the generated code and generated structs would handle that for us.

We should open a ticket for this IMO if there isn't one already.

api/models/assetrelationship.go
Tehsmash: Looks like its this one:
https://golangci-lint.run/usage/linters/#errorlint

The check is disabled by default but we can enable it using the "errorf" config option.

chrisgacsal: We definitely should in order to make sure we do wrap errors properly.

chrisgacsal (approved):
LGTM! Thanks for the fixes.

Tehsmash (approved):

Issue #608, opened 2023-08-30
Incorrect use of errors.Join in aws/client.go


Tehsmash:
https://github.com/openclarity/vmclarity/blob/efac30af61d52e1deda4b0bd6264825fca6a9066/pkg/orchestrator/provider/aws/client.go#L400-L409

errors.Join should be called with the complete list of errors to wrap them correctly. In should not be called in a loop otherwise you end up with a nested structure like:
```
JoinError{Errs: []error{JoinError{Errs: []error{JoinError{}}}}}
```
instead of 
```
JoinError{Errs: []err{err1, err2, err3}}
```

PR #609, opened 2023-08-30
Fixing issue for Incorrect use of errors.Join in aws/client.go #608

praveen0raj:
## Description

Changing data type of "errs" from channel to slice 
By using slice , there is no need for "for loop" for iterating channel, we can use slice directly in errors.Join(errs...) function

## Type of Change

[x] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/pkg/orchestrator/provider/aws/client.go
+++ b/pkg/orchestrator/provider/aws/client.go
@@ -314,18 +314,8 @@ func (c *Client) RunAssetScan(ctx context.Context, config *provider.ScanJobConfi
 	//                    in parallel.
 
 	// Create scanner instance
-	var errs []error
-	errsChan := make(chan error)
-
-	var errWg sync.WaitGroup
-	errWg.Add(1)
-	go func() {
-		defer errWg.Done()
-		for e := range errsChan {
-			errs = append(errs, e)
-		}
-	}()
-
+	numOfGoroutines := 2
+	errs := make(chan error, numOfGoroutines)
 	var scannnerInstance *Instance
 	var wg sync.WaitGroup
 	wg.Add(1)
@@ -337,20 +327,20 @@ func (c *Client) RunAssetScan(ctx context.Context, config *provider.ScanJobConfi
 		var err error
 		scannnerInstance, err = c.createInstance(ctx, c.config.ScannerRegion, config)
 		if err != nil {
-			errsChan <- WrapError(fmt.Errorf("failed to create scanner VM instance: %w", err))
+			errs <- WrapError(fmt.Errorf("failed to create scanner VM instance: %w", err))
 			return
 		}
 
 		ready, err := scannnerInstance.IsReady(ctx)
 		if err != nil {
-			errsChan <- WrapError(fmt.Errorf("failed to get scanner VM instance state: %w", err))
+			errs <- WrapError(fmt.Errorf("failed to get scanner VM instance state: %w", err))
 			return
 		}
 		logger.WithFields(logrus.Fields{
 			"ScannerInstanceID": scannnerInstance.ID,
 		}).Debugf("Scanner instance is ready: %t", ready)
 		if !ready {
-			errsChan <- RetryableError{
+			errs <- RetryableError{
 				Err:   errors.New("scanner instance is not ready"),
 				After: InstanceReadynessAfter,
 			}
@@ -370,7 +360,7 @@ func (c *Client) RunAssetScan(ctx context.Context, config *provider.ScanJobConfi
 
 		assetVMLocation, err := NewLocation(vmInfo.Location)
 		if err != nil {
-			errsChan <- FatalError{
+			errs <- FatalError{
 				Err: fmt.Errorf("failed to parse Location for asset VM instance: %w", err),
 			}
 			return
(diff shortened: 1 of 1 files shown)


pkg/orchestrator/provider/aws/client.go
```diff
@@ -256,7 +256,7 @@ func (c *Client) RunAssetScan(ctx context.Context, config *provider.ScanJobConfi
 
 	// Create scanner instance
 	numOfGoroutines := 2
-	errs := make(chan error, numOfGoroutines)
+	errs := make([]error, numOfGoroutines)
```
FrimIdan, line 259: You shouldn't write into a slice by multiple goroutines.
You can keep it as an unbufferd channel and do something like
```
var errs []error
errorsChan := make(chan error)
go func() {
	for e := range errorsChan {
		errs = append(errs, e)
	}
}()
.
.
.

close(errorsChan)
err = errors.Join(errs...)
```

praveen0raj: Yes @FrimIdan . Thanks for reminding .

I think we should use buffered channel as we already know how many goroutines we need in advance. 

**Sample:**

```
.
.
numOfGoroutines := 3
var errs []error
errsChan := make(chan error, numOfGoroutines)

var wg sync.WaitGroup

wg.Add(1)
go func() {
defer wg.Done()
.
.
errsChan <- WrapError(fmt.Errorf("failed to delete scanner volume: %w", err))
.
.
}()


wg.Add(1)
go func() {
defer wg.Done()
.
.
errsChan <- WrapError(fmt.Errorf("failed to delete scanner volume snapshot: %w", err))
.
.
}()

wg.Add(1)
go func() {
defer wg.Done()
.
.
errsChan <- fmt.Errorf("failed to delete asset volume snapshot: %w", err).
.
.
}()

wg.Wait()
close(errsChan)
for e := range errsChan {
        if e != nil {
	        // nolint:typecheck
	        errs = append(errs, e)
        }
}
if err != nil {
        errs = append(errs, err)
}
err = errors.Join(errs...)

```

Let me know if above changes are fine or anything else is needed.

Thanks

Tehsmash: Using a buffered channel assumes that every go routine is going to error and only error once, hopefully none of the go routines will error so allocating a buffered channel allocates memory we don't need.

I think its much cleaner/readable to have a routine draining the error channel into the slice, and then after everything is done close the channels which end the routines, and then you can just join the slice straight away instead of having another loop to run after everything else is completed.

praveen0raj: **For the first part:**
It seems to me, we can not drain all errors of unbufferred channel at one go.
As per documentaion, If the channel is unbuffered, the sender blocks until the receiver has received the value. 

I have tried to work with sample program below.
It works with buffered channel but not working fine with unbuffered channel (errsChan := make(chan error)). 

**For the second part:** 
yes it's a better approach for a separate gorutine for draining error channel.
I have mentioned below, please check.

```
func main() {

	var errs []error
	errsChan := make(chan error, 4)

	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		defer wg.Done()
		errsChan <- errors.New("error 1")
		errsChan <- errors.New("error 2")
	}()

	wg.Add(1)
	go func() {
		defer wg.Done()
		errsChan <- errors.New("error 3")
	}()

	wg.Add(1)
	go func() {
		defer wg.Done()
		errsChan <- errors.New("error 4")
	}()
	wg.Wait()
	close(errsChan)

	wg.Add(1)
	go func() {
		defer wg.Done()
		for e := range errsChan {
			errs = append(errs, e)
		}
	}()
	wg.Wait()

	fmt.Print(errs)
}
```

Tehsmash: Ah I see race condition, basically we need to make sure that the draining routine is done after we close the channel so that there is no race between the `fmt.Print(errs)` and the drain. This code works https://go.dev/play/p/UEUxO-gWIz6 please take a look.

praveen0raj: Great . Thanks a lot for knowledge sharing . :+1: 

I have pushed changes again.

chrisgacsal (commented):
I am not sure about the proposed fix as it seems to be more complex that it needs to be.

An alternative approach which is much simpler and does have the same result.

```
diff --git a/pkg/orchestrator/provider/aws/client.go b/pkg/orchestrator/provider/aws/client.go
index 2cd2df63..9cd19d11 100644
--- a/pkg/orchestrator/provider/aws/client.go
+++ b/pkg/orchestrator/provider/aws/client.go
@@ -398,12 +398,14 @@ func (c *Client) RunAssetScan(ctx context.Context, config *provider.ScanJobConfi
 	close(errs)
 
 	// NOTE: make sure to drain results channel
+	listOfErrors := make([]error, 0)
 	for e := range errs {
 		if e != nil {
-			// nolint:typecheck
-			err = errors.Join(err, e)
+			listOfErrors = append(listOfErrors, e)
 		}
 	}
+
+	err = errors.Join(listOfErrors...)
 	if err != nil {
 		return err
 	} 
```

praveen0raj:
> I think I agree with @chrisgacsal, we've kind of exceeded the scope of the ticket with this change. I think we should leave the go routines as they were and just fix the errors.Join. The go routines can be changed in a follow up PR / other refactor PR. @praveen0raj are you ok to revert the go routine changes and just leave the fix for errors.Join?

Yes I too agree. 

I have reverted changes and I have included changes as suggested by @chrisgacsal.
Thanks

Tehsmash (approved):

PR #613, opened 2023-08-31
fix: family errors export

fishkerez:
## Description

In some of the families, when we export results from the CLI, we are not exporting the status/errors as null, but as an empty list. This breaks our ability to do the query: "status/family/errors eq null" for an assetScan.

## Type of Change

[ *] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

- [* ] I have verified this change is not present in other open pull requests

### DIFF
--- a/pkg/cli/presenter/vmclarity.go
+++ b/pkg/cli/presenter/vmclarity.go
@@ -87,7 +87,7 @@ func (v *VMClarityPresenter) ExportSbomResult(ctx context.Context, res families.
 		assetScan.Stats = &models.AssetScanStats{}
 	}
 
-	errs := []string{}
+	var errs []string
 
 	if res.Err != nil {
 		errs = append(errs, res.Err.Error())
@@ -135,7 +135,7 @@ func (v *VMClarityPresenter) ExportVulResult(ctx context.Context, res families.F
 		assetScan.Stats = &models.AssetScanStats{}
 	}
 
-	errs := []string{}
+	var errs []string
 
 	if res.Err != nil {
 		errs = append(errs, res.Err.Error())
@@ -182,7 +182,7 @@ func (v *VMClarityPresenter) ExportSecretsResult(ctx context.Context, res famili
 		assetScan.Stats = &models.AssetScanStats{}
 	}
 
-	errs := []string{}
+	var errs []string
 
 	if res.Err != nil {
 		errs = append(errs, res.Err.Error())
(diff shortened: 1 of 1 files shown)


chrisgacsal (commented):
@fishkerez Thanks for fixing this. Could you please leave a comment in the code for each `err` definition about this issue. So others won't revert this fix as normally ppl would initialize slices and some linters also mark this as an error.

The proper fix would be to introduce the `Nullable` type @Tehsmash has been working on to fix the problem where we need to differentiate between nil and empty values.

Tehsmash (changes requested):
I think exporting an empty list is the correct behaviour, IMO its the query that should be fixed so that it looks for null or an empty list. If we send null it leads to weird behaviour in the API/DB.

We check for null or empty in a few other queries, I don't know if `eq []` is supported in odata, I think it should be `length(errors) eq 0` so we need to add support for the length function in odatasql.

fishkerez:
> I think exporting an empty list is the correct behaviour, IMO its the query that should be fixed so that it looks for null or an empty list.

Currently we do not have a uniform behavior, in some families we send null and in some empty list, so we need to fix this anyway. 

>  If we send null it leads to weird behaviour in the API/DB.

If we don't want to allow errors to be null we need to remove the nullable from the API. In case of arrays, maybe we really don't need the nullable.

> We check for null or empty in a few other queries, I don't know if eq [] is supported in odata, I think it should be length(errors) eq 0 so we need to add support for the length function in odatasql.

Can you reference me how to check for empty list in the query? If it is not supported than how can we do that?

I think we need to write some clear guidelines on this subject of empty vs null to avoid weird behaviours in the future. DYT?

Tehsmash:
> > If we send null it leads to weird behaviour in the API/DB.
> 
> If we don't want to allow errors to be null we need to remove the nullable from the API. In case of arrays, maybe we really don't need the nullable.

I agree we need to remove nullable from the API, its not working as expected and causes bugs. I have a PR open that tries to fix the behaviour https://github.com/openclarity/vmclarity/pull/444

> Can you reference me how to check for empty list in the query? If it is not supported than how can we do that?

Looking at the odata documentation it should be `$filter=length(state/general/errors) eq 0`. We need to add the support for the length function in odatasql.

> I think we need to write some clear guidelines on this subject of empty vs null to avoid weird behaviours in the future. DYT?

100%, in POST/PUT null shouldn't be support for any fields. It has a different meaning for PATCH, in PATCH it means "delete/unset the data in that field" which is different from storing "null".

Tehsmash:
I've opened https://github.com/openclarity/vmclarity/pull/665 which adds support for the length() function so then we don't need to rely on null anymore. We still need to address the nullable issue in PATCH which we'll continue in #444

PR #621, opened 2023-09-01
feat: Lay the foundations for the kubernetes provider

### LINKED ISSUE #561: Kubernetes Provider Infrastructure
Based on the design #527 we should create a new Kubernetes provider which understands how to discover and scan Kubernetes assets (Containers and Container Images).

The first part of this work is to create the framework for the Kubernetes provider so that the features can be implemented in parallel:
* A new kubernetes provider struct should be created which implements the Discoverer and Scanner interfaces:
  * the RunAssetScan and RemoveAssetScan functions should be left empty/return a NotImplemented error were required.
  * a Kubernetes client should be initialised
* A configuration type should be added which is prepared to load environment variables with the prefix VMCLARITY_KUBERNETES
   * Some initial configurations should be added such as scanner job namespace

Tehsmash:
## Description

This commit:

* Adds the provider struct and the empty functions for discovering and
  scanning.

* Creates a kubernetes client for accessing the kubernetes API

* Adds a configuration loader which looks for environment variables
  prefixed with VMCLARITY_KUBERNETES

* Adds the Kubernetes provider type to the api spec

* Updates uibackend to handle the kubernetes provider type

* Adds location field to the container asset type which can be used by
  the provider to keep track of where a container was found

* Support kubernetes provider configuration in helm-chart

Closes #561

## Type of Change

[ ] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


pkg/orchestrator/provider/kubernetes/provider.go
```diff
@@ -0,0 +1,87 @@
...
+			return nil, fmt.Errorf("unable to load client configuration: %w", err)
+		}
+	} else {
+		cc, err := clientcmd.LoadFromFile(config.KubeConfig)
+		if err != nil {
+			return nil, fmt.Errorf("unable to load client configuration: %w", err)
+		}
+		clientConfig, err = clientcmd.NewNonInteractiveClientConfig(*cc, "", &clientcmd.ConfigOverrides{}, nil).ClientConfig()
+		if err != nil {
+			return nil, fmt.Errorf("unable to load client configuration: %w", err)
```
paralta, line 58: Maybe change this error message to something different than the two previous messages

paralta (dismissed):
Just a small suggestion, otherwise LGTM :D

pkg/orchestrator/provider/kubernetes/provider.go
```diff
@@ -0,0 +1,87 @@
...
+	clientset kubernetes.Interface
+	config    *Config
+}
+
+var _ provider.Provider = &Provider{}
+
+func New(ctx context.Context) (provider.Provider, error) {
+	config, err := NewConfig()
+	if err != nil {
+		return nil, fmt.Errorf("invalid configuration. Provider=%s: %w", models.Kubernetes, err)
```
FrimIdan, line 40: Do we really need to add `Provider=%s` here? and if here why not in the other error prints

pkg/orchestrator/provider/kubernetes/provider.go
FrimIdan: +1

Tehsmash: Updated please check the latest revision.

pkg/uibackend/rest/dashboard_riskiest_assets.go
```diff
@@ -212,13 +212,31 @@ func getAssetInfo(asset *backendmodels.AssetType) (*models.AssetInfo, error) {
...
+	return &models.AssetInfo{
+		Name: info.ContainerName,
+		Type: utils.PointerTo(models.Container),
+	}, nil
+}
+
+func containerImageInfoToAssetInfo(info backendmodels.ContainerImageInfo) (*models.AssetInfo, error) {
+	return &models.AssetInfo{
+		Name: info.Name,
+		Type: utils.PointerTo(models.Container),
```
FrimIdan, line 234: `models.ContainerImage`?
Also, what about `location` here and in `containerInfoToAssetInfo`?

pkg/orchestrator/provider/kubernetes/provider.go
Tehsmash: Good point, probably not, I copied this structure from one of the other providers

pkg/uibackend/rest/dashboard_riskiest_assets.go
Tehsmash: I forgot to go back and update these, great catch! I've pushed a new revision.

pkg/uibackend/rest/dashboard_riskiest_assets.go
```diff
@@ -212,13 +213,34 @@ func getAssetInfo(asset *backendmodels.AssetType) (*models.AssetInfo, error) {
 	switch info := discriminator.(type) {
 	case backendmodels.VMInfo:
 		return vmInfoToAssetInfo(info)
+	case backendmodels.ContainerInfo:
+		return containerInfoToAssetInfo(info)
+	case backendmodels.ContainerImageInfo:
+		return containerImageInfoToAssetInfo(info)
```
FrimIdan, line 219: need to update the unit test to cover the new cases

pkg/uibackend/rest/dashboard_riskiest_assets_test.go
```diff
@@ -270,6 +294,28 @@ func createVMInfo(t *testing.T, instanceID, location string) *backendmodels.Asse
...
+	})
+	assert.NilError(t, err)
+	return &info
+}
+
+func createContainerImageInfo(t *testing.T) *backendmodels.AssetType {
+	t.Helper()
+	info := backendmodels.AssetType{}
+	err := info.FromContainerImageInfo(backendmodels.ContainerImageInfo{
+		Name: utils.PointerTo("ghcr.io/openclarity/vmclarity-orchestrator"),
```
FrimIdan, line 313: what about the image tag/digest? how it is represented?

Tehsmash: ContainerImageInfo has an ID field which will contain the digest. Right now we don't have a field that contains the tags. This is all UI related and I'm not sure we're 100% sure how we want to represent this stuff in the UI so this is mostly placeholder.

I need to sync with @paralta, @ramizpolic and @adamtagscherer and update the Container and ContainerInfo to include details in a generic way that works for both Docker, ContainerD and Kubernetes.

pkg/orchestrator/provider/kubernetes/provider.go
Tehsmash: updated in the latest revision

FrimIdan (approved):

Issue #445, opened 2023-07-05
Paginate Assets in DiscoverAssets


ghost:
Thats a great point, the AWS DiscoverAssets flow paginates the AWS call but it still results in all the models.AssetInfo's in memory. We should probably change that into a channel so the pagination is passed through and each Asset is handled one at a time.

_Originally posted by @sambetts-cisco in https://github.com/openclarity/vmclarity/pull/438#discussion_r1250962691_

PR #626, opened 2023-09-05
Buffered Asset Discovery

fishkerez:
## Description

Resolve https://github.com/openclarity/vmclarity/issues/445

Modify DiscoverAssets interface to return a channel of assets instead of
a list. This lays the foundation to prevent us from keeping big chunks
of assets in memory.

In a future PR, all the providers should be updated to paginate fetching
assets from their downstream APIs and feed the channel.
 
## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/pkg/orchestrator/provider/aws/client.go
+++ b/pkg/orchestrator/provider/aws/client.go
@@ -70,34 +70,46 @@ func (c Client) Kind() models.CloudProvider {
 }
 
 // nolint:cyclop
-func (c *Client) DiscoverAssets(ctx context.Context) ([]models.AssetType, error) {
-	regions, err := c.ListAllRegions(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get regions: %w", err)
-	}
+func (c *Client) DiscoverAssets(ctx context.Context) provider.AssetDiscoverer {
+	assetDiscoverer := provider.NewSimpleAssetDiscoverer()
 
-	logger := log.GetLoggerFromContextOrDiscard(ctx)
+	go func() {
+		defer close(assetDiscoverer.OutputChan)
 
-	assets := make([]models.AssetType, 0)
-	for _, region := range regions {
-		instances, err := c.GetInstances(ctx, []ec2types.Filter{}, region.Name)
+		regions, err := c.ListAllRegions(ctx)
 		if err != nil {
-			logger.Warnf("Failed to get instances. region=%v: %v", region, err)
-			continue
+			assetDiscoverer.Error = fmt.Errorf("failed to get regions: %w", err)
+			return
 		}
 
-		for _, instance := range instances {
-			asset, err := getVMInfoFromInstance(instance)
+		logger := log.GetLoggerFromContextOrDiscard(ctx)
+
+		for _, region := range regions {
+			instances, err := c.GetInstances(ctx, []ec2types.Filter{}, region.Name)
 			if err != nil {
-				return nil, FatalError{
-					Err: fmt.Errorf("failed convert EC2 Instance to AssetType: %w", err),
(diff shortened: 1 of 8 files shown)


pkg/orchestrator/provider/aws/client.go
```diff
@@ -70,34 +70,44 @@ func (c Client) Kind() models.CloudProvider {
...
+
+			for _, instance := range instances {
+				asset, err := getVMInfoFromInstance(instance)
+				if err != nil {
+					assetDiscoverer.Error = FatalError{
+						Err: fmt.Errorf("failed convert EC2 Instance to AssetType: %w", err),
+					}
+					return
 				}
+				output <- asset
```
Tehsmash, line 105: I think we need to make sure that we handle being canceled properly otherwise we could block here forever, for example:
```
select {
   case output <- asset:
   case <-ctx.Done():
       assetDiscoverer.Error = ctx.Err()
       return
}
```

Tehsmash (changes requested):
One minor tweak for context management, but otherwise it looks great! We should open follow up tickets for each provider to review/refactor the DiscoverAssets function to make sure they are paginating properly, add config options to determine how many items should be included in a page etc.

pkg/orchestrator/provider/aws/client.go
```diff
@@ -70,34 +70,44 @@ func (c Client) Kind() models.CloudProvider {
...
 
 // nolint:cyclop
-func (c *Client) DiscoverAssets(ctx context.Context) ([]models.AssetType, error) {
-	regions, err := c.ListAllRegions(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get regions: %w", err)
+func (c *Client) DiscoverAssets(ctx context.Context) provider.AssetDiscoverer {
+	output := make(chan models.AssetType)
+	assetDiscoverer := provider.AssetDiscovererImpl{
+		OutputChan: output,
```
FrimIdan, line 76: Maybe better to create a `CreateAssetDiscovererImpl` func for the creation of the `AssetDiscoverer`?

You can also create it once outside the `DiscoverAssets` func and pass it as an arg to that function, no? (to avoid doing it for each provider implementation)

You might need to add a `SetErr` func to the interface.

Tehsmash: I don't think it should be created outside of the DiscoverAssets function, it should be down to the provider to decide how to implement the AssetDiscoverer interface. This is just a default naive implementation IMO.

I agree with having a `func NewNaiveAssetDiscoverer() NaiveAssetDiscoverer` function to remove the duplication around the channel creation.

FrimIdan: make sense

pkg/orchestrator/provider/aws/client.go
```diff
@@ -70,34 +70,47 @@ func (c Client) Kind() models.CloudProvider {
 }
 
 // nolint:cyclop
-func (c *Client) DiscoverAssets(ctx context.Context) ([]models.AssetType, error) {
-	regions, err := c.ListAllRegions(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get regions: %w", err)
-	}
+func (c *Client) DiscoverAssets(ctx context.Context) provider.AssetDiscoverer {
+	output := make(chan models.AssetType)
```
fishkerez, line 74: I wonder if we want to make this channel buffered with a max size of items. Theoretically, we can have a situation where all the assets are inserted into the channel and the memory is still blown

Tehsmash (approved):

Issue #576, opened 2023-08-18
YARA Malware scanner job


Tehsmash:
Similar to ClamAV integration, need to:

* Add the YARA CLI to the base image
* Add it into the Malware family
* Add to scan families config
* Generate the YARA rule list from the directory
* YARA output parser
* Extend the Reporter and the Exported to contain:
   * Rule name
   * Add the rule name to the malware table and drilldown

PR #638, opened 2023-09-06
feat: yara support

pbalogh-sa:
## Description

This PR adds Yara support for VMClarity. 
closes: #576 
Merge after `yara-rule-server` was released: https://github.com/openclarity/yara-rule-server/pull/1

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


pkg/shared/families/malware/yara/yara.go
```diff
@@ -0,0 +1,168 @@
...
+
+	"github.com/openclarity/vmclarity/pkg/shared/families/malware/common"
+	"github.com/openclarity/vmclarity/pkg/shared/families/malware/yara/config"
+	"github.com/openclarity/vmclarity/pkg/shared/families/malware/yara/util"
+	sharedutils "github.com/openclarity/vmclarity/pkg/shared/utils"
+)
+
+const (
+	ScannerName               = "yara"
+	TemporaryYaraRuleFilePath = "/tmp/yara-rules.yarac"
```
FrimIdan, line 39: Shouldn't it be configurable?

pkg/shared/families/malware/yara/yara.go
```diff
@@ -0,0 +1,168 @@
...
+		args := []string{"-C", rulesPath, "-r", "-w", "-m", userInput}
+
+		s.logger.Infof("Running yara...")
+		// Execute the yara command
+		// nolint:gosec
+		yaraCommand := exec.Command(s.config.YaraBinaryPath, args...)
+		out, err := sharedutils.RunCommand(yaraCommand)
+		if err != nil {
+			s.sendResults(retResults, fmt.Errorf("failed to run yara command: %s", err.Error()))
+		}
```
FrimIdan, line 80: missing a return?

pkg/shared/families/malware/yara/yara.go
```diff
@@ -0,0 +1,168 @@
...
+			s.sendResults(retResults, nil)
+			return
+		}
+
+		s.logger.Debugf("Yara binary path: %s", s.config.YaraBinaryPath)
+		s.logger.Debugf("Yara rules URL: %s", s.config.YaraRuleURL)
+		rulesPath, err := s.getRuleFilePath()
+		if err != nil {
+			retResults.Error = fmt.Errorf("failed to get Yara rules file path: %v", err)
+		}
```
FrimIdan, line 67: missing a `sendResults` and `return`?

pkg/shared/families/malware/yara/yara.go
```diff
@@ -0,0 +1,168 @@
...
+	case "http", "https":
+		return s.downloadRules()
+	default:
+		return "", errors.New("unsupported Yara rules URL")
+	}
+}
+
+func (s *Scanner) downloadRules() (string, error) {
+	s.logger.Infof("Downloading Yara rules...")
+	out, err := os.Create(TemporaryYaraRuleFilePath)
```
FrimIdan, line 111: what if already exist?

pkg/shared/families/malware/yara/config/config.go
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+type Config struct {
+	YaraBinaryPath string `yaml:"yara_binary_path" mapstructure:"yara_binary_path"`
```
Tehsmash, line 19: no need to prefix these with yara, we're already inside of the yara.Config{} struct.

pkg/shared/families/malware/yara/util/yara_util.go
```diff
@@ -0,0 +1,126 @@
...
+	words[yaraRuleNamePosition] = strings.TrimSpace(fields[0])
+	fields = strings.Split(fields[1], "]")
+	if len(fields) != 2 { // nolint:gomnd
+		return nil
+	}
+	words[yaraRuleMetadataPosition] = fields[0]
+	words[detectedMalwarePathPosition] = strings.TrimSpace(fields[1])
+
+	return words
+}
```
Tehsmash, line 78: It might be clearer to use strings.Cut for example:
```
ruleName, remainder, found := strings.Cut(line, "[")
if !found {
    return nil
}

metadata, path, found := strings.Cut(remainder, "]")
if !found {
   return nil
}

return []string{strings.TrimSpace(ruleName), strings.TrimSpace(metadata), strings.TrimSpace(path)}
```

Perhaps we also want to return a struct here instead of a list:

```
type ScanResultLine struct {
    RuleName string
    Metadata string
    Path string
}
```

You could then parse the metadata into KV as part of this function and return:

```
type ScanResultLine struct {
    RuleName string
    Metadata map[string]string
    Path string
}
```

pkg/shared/families/malware/yara/util/yara_util.go
```diff
@@ -0,0 +1,126 @@
...
+
+func parseYaraRuleMetadata(metadata string) map[string]string {
+	m := make(map[string]string)
+	if metadata == "" {
+		return m
+	}
+	entries := strings.Split(metadata, ",")
+	for _, e := range entries {
+		parts := strings.Split(e, "=")
```
Tehsmash, line 118: ditto about using `strings.Cut` here:
```
key, value, found := strings.Cut(e, "=")
...
```

pkg/shared/families/malware/yara/util/yara_util_test.go
```diff
@@ -0,0 +1,179 @@
...
+				words: []string{},
+			},
+			want: nil,
+		},
+		{
+			name: "words are exist",
+			args: args{
+				words: []string{
+					"vmdetect",
+					"author=\"nex\",description=\"Possibly employs anti-virtualization techniques\"",
```
Tehsmash, line 46: you can use "raw" strings via backticks to avoid all the escaping in these cases:
```
`author="nex",description="..."`
```

pkg/shared/families/malware/yara/yara.go
Tehsmash: I think we should use the golang Mktmp etc functions for temporary directories

pkg/shared/families/malware/yara/yara.go
Tehsmash: We need to handle the scenario where the yara scanner called multiple-times (multiple inputs), but we probably only want to download the file once. Perhaps this should be downloaded as part of the scanners' initialisation instead of part of its "Run(..)" or wrapped in a "sync.Once(...)" so that it happens once per CLI run.

api/openapi.yaml
```diff
@@ -2118,6 +2118,8 @@ components:
           type: string
         malwareType:
           $ref: '#/components/schemas/MalwareType'
+        ruleName:
+          type: string
```
Tehsmash, line 2122: The name of this field is still bothering me, I don't like how specific it seems to be to yara and how much of a singleton it makes this object.

Remember if two malware scanners identify the same malware in the same file it should ideally be a single entry in the result not two separate ones otherwise all we're doing is smashing all the results together.

I think we need a structure like:

```
matches:
   type: array
   items:
      type: object
   properties:
      scanner:
         type: string
      signature:
         type: string
```

then each finding in the malware report is for a specific Malware and Path.

For example:

```
malwares: [
   {
      "malwareName": "trojan bad bad",
      "path": "/tmp/averybadfile",
      "matches": [
          {
             "scanner": "yara",
             "signature": "thetrojanbadbad_rule1"
          },
          {
              "scanner": "clamav",
              "signature": "unknown"
          },
      ]
   }
]
```

For a match which doesn't know the malwareName, then we can just use "unknown". For example:

```
   {
      "malwareName": "unknown",
      "path": "/tmp/averybadfile",
      "matches": [
          {
             "scanner": "yara",
             "signature": "somerule_thatdoesnt_list_themalware"
          },
      ]
   }
```

If there are multiple matches for the same path but we don't know the malware they can be merged together like:

```
   {
      "malwareName": "unknown",
      "path": "/tmp/averybadfile",
      "matches": [
          {
             "scanner": "yara",
             "signature": "somerule_thatdoesnt_list_themalware"
          },
          {
             "scanner": "yara",
             "signature": "another_rule_with_no_malware"
          },
      ]
   }
```

Tehsmash (changes requested):

Makefile
```diff
@@ -64,8 +64,7 @@ docker-cli: ## Build CLI Docker image
 		--build-arg BUILD_TIMESTAMP=$(shell date -u +"%Y-%m-%dT%H:%M:%SZ") \
 		--build-arg COMMIT_HASH=$(shell git rev-parse HEAD) \
 		${VMCLARITY_TOOLS_CLI_DOCKER_ARG} \
-		-t ${DOCKER_IMAGE}-cli:${DOCKER_TAG} \
-		.
+		-t ${DOCKER_IMAGE}-cli:${DOCKER_TAG}
```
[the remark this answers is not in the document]
pbalogh-sa: mistake

pkg/shared/families/malware/yara/config/config.go
```diff
@@ -0,0 +1,21 @@
...
+	YaraRuleURL    string `yaml:"yara_rule_url" mapstructure:"yara_rule_url"`
```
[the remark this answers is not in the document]
pbalogh-sa: I've updated the replaced YaraRuleURL to CompiledRuleURL. 
I think we can extend later the config with the `RuleSources` as a follow-up task.

pkg/shared/families/malware/yara/util/yara_util.go
```diff
@@ -0,0 +1,126 @@
...
+	scanError                   = "error scanning"
+	yaraRuleNamePosition        = 0
+	yaraRuleMetadataPosition    = 1
+	detectedMalwarePathPosition = 2
+	malwareNameString           = "name" // The key defined in the metadata of yara rule
+	malwareTypeString           = "type" // The key defined in the metadata of yara rule
+	yaraOutputFiledNum          = 3      // Number of yara output fields. <yara rule name> [<coma,separated,meta data>] <path>
+)
+
+func ParseYaraScanOutput(yaraOutput string, logger *logrus.Entry) []common.DetectedMalware {
```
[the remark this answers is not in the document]
pbalogh-sa: You are right. I think we need to update clam output parsing as well. So it can be a follow-up task for both of the yara and clam. We have to create a new function that has two args (command, parser_func) that runs the command and parse output line-by-line. WDYT?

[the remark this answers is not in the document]
Tehsmash: Sure, lets make sure we open a ticket to track it.

pkg/shared/families/malware/yara/util/yara_util.go
```diff
@@ -0,0 +1,126 @@
...
+
+const (
+	// A yara detected malware line should be in the following format: <yara rule name> [<coma,separated,meta data>] <path>.
+	scanError                   = "error scanning"
+	yaraRuleNamePosition        = 0
+	yaraRuleMetadataPosition    = 1
+	detectedMalwarePathPosition = 2
+	malwareNameString           = "name" // The key defined in the metadata of yara rule
+	malwareTypeString           = "type" // The key defined in the metadata of yara rule
+	yaraOutputFiledNum          = 3      // Number of yara output fields. <yara rule name> [<coma,separated,meta data>] <path>
```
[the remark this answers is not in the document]
pbalogh-sa: removed

api/openapi.yaml
pbalogh-sa: I've opened an issue: https://github.com/openclarity/vmclarity/issues/677

pkg/shared/families/malware/yara/yara.go
Tehsmash: @pbalogh-sa it doesn't look like the downloadRules function will handle multiple runs of the yara scanner, so if we have multiple inputs (multiple volumes for example) then it'll download the rules multiple times still. Are you going to address this in a follow up?

pkg/shared/families/malware/yara/util/yara_util.go
```diff
@@ -0,0 +1,126 @@
...
+	logger.Debugf("Parsing yara output: %s", yaraOutput)
+	lines := strings.Split(yaraOutput, "\n")
+
+	malwareInfoList := make([]common.DetectedMalware, 0)
+	for i := range lines {
+		logger.Debugf("line %d: %s", i, lines[i])
+		if strings.Contains(lines[i], scanError) {
+			logger.Errorf("Yara scan failed: %s", lines[i])
+			continue
+		}
```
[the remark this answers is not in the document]
Tehsmash: @pbalogh-sa wdyt?

pkg/orchestrator/assetscanprocessor/malware.go
```diff
@@ -33,7 +33,7 @@ func (asp *AssetScanProcessor) getExistingMalwareFindingsForScan(ctx context.Con
 	existingFilter := fmt.Sprintf("findingInfo/objectType eq 'Malware' and foundBy/id eq '%s'", *assetScan.Id)
 	existingFindings, err := asp.client.GetFindings(ctx, models.GetFindingsParams{
 		Filter: &existingFilter,
-		Select: utils.PointerTo("id,findingInfo/malwareName,findingInfo/malwareType,findingInfo/path"),
+		Select: utils.PointerTo("id,findingInfo/malwareName,findingInfo/malwareType,findingInfo/path,findingIfo/ruleName"),
```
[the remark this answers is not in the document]
Tehsmash:
```suggestion
		Select: utils.PointerTo("id,findingInfo/malwareName,findingInfo/malwareType,findingInfo/path,findingInfo/ruleName"),
```

pkg/shared/families/malware/yara/yara.go
Tehsmash: On re-review I the called for downloadRules has been moved into the New function 👍

Tehsmash (changes requested):

pkg/shared/families/malware/yara/util/yara_util.go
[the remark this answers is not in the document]
pbalogh-sa: We need to research it because there are no findings lines if nothing matches a rule, but permission errors can occur during opening files. So It is hard to find the right solution.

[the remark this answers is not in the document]
Tehsmash: ah I see, yeah thats tough... looks like we can force yara to output the files it scans using `--scan-list` so then we could compare it to the errors, but we'd need to fix the output streaming first to avoid blowing up the memory.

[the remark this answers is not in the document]
Tehsmash: Maybe worth opening another ticket for this to investigate

pkg/orchestrator/assetscanprocessor/malware.go
```diff
@@ -33,7 +33,7 @@ func (asp *AssetScanProcessor) getExistingMalwareFindingsForScan(ctx context.Con
 	existingFilter := fmt.Sprintf("findingInfo/objectType eq 'Malware' and foundBy/id eq '%s'", *assetScan.Id)
 	existingFindings, err := asp.client.GetFindings(ctx, models.GetFindingsParams{
 		Filter: &existingFilter,
-		Select: utils.PointerTo("id,findingInfo/malwareName,findingInfo/malwareType,findingInfo/path"),
+		Select: utils.PointerTo("id,findingInfo/malwareName,findingInfo/malwareType,findingInfo/path,findingInfo/ruleName"),
```
Tehsmash, line 36: We don't need this here *unless* GenerateMalwareKey is using it, but that function hasn't been updated.

installation/azure/vmclarity-install.sh
```diff
@@ -284,11 +321,14 @@ services:
       - source: swagger_config
         target: /usr/share/nginx/html/swagger-config.json
 
+
```
Tehsmash, line 324: nit: extra whitespace

pkg/shared/utils/utils.go
```diff
@@ -57,6 +57,14 @@ func PointerTo[T any](value T) *T {
 	return &value
 }
 
+func PointerToStringOrNil(s string) *string {
+	if s == "" {
+		return nil
+	}
+
+	return &s
+}
```
Tehsmash, line 66: I wonder if we can make a generic here which checks for the zero type. Something like:

```
func OmitEmpty[T comparable](t T) *T {
	var empty T
	if t == empty {
		return nil
	}
	return &t
}
```
https://go.dev/play/p/UW1rdYWe_V2

This works for all comparable types, but doesn't work for things like []string unfortunately.

ui/src/layout/Findings/Malware/MalwareTable.js
```diff
@@ -17,7 +17,14 @@ const MalwareTable = () => {
...
-            width: 300
+            width: 200
+        },
+        {
+            Header: "Rule name",
+            id: "ruleName",
+            sortIds: ["findingInfo.ruleName"],
+            accessor: "findingInfo.ruleName",
+            width: 200
         },
```
Tehsmash, line 28: I think it would probably be better to have this after Malware Name before Path, the rule is serving a similar purpose to Rule Name at the moment. (This will probably change if we redo the model anyway though so we can leave it if we like)

Tehsmash (changes requested):

akpsgit (dismissed):
I request changes

pkg/orchestrator/assetscanprocessor/malware.go
```diff
@@ -83,6 +83,7 @@ func (asp *AssetScanProcessor) reconcileResultMalwareToFindings(ctx context.Cont
 				MalwareName: item.MalwareName,
 				MalwareType: item.MalwareType,
 				Path:        item.Path,
+				RuleName:    item.RuleName,
```
Tehsmash, line 86: If we don't update getExistingMalwareFindingsForScan + GenerateMalwareKey, then Malware's are considered unique if MalwareName, MalwareType and Path are the same (which will be true if we restructure things to include the matches) but right now this means that if two rules match on the same file, then we through away any existing matches. Are we ok with that?

pbalogh-sa: I am adding the `ruleName` to the `MalwareKey`.

pkg/shared/findingkey/malware.go
```diff
@@ -19,22 +19,25 @@ import (
...
 type MalwareKey struct {
 	MalwareName string
 	MalwareType string
 	Path        string
+	RuleName    string
 }
 
 func (k MalwareKey) String() string {
-	return fmt.Sprintf("%s.%s.%s", k.MalwareName, k.MalwareType, k.Path)
+	return fmt.Sprintf("%s.%s.%s.%s", k.MalwareName, k.MalwareType, k.Path, k.MalwareName)
```
Tehsmash, line 33: Should be RuleName right?

fishkerez (dismissed):

pkg/shared/findingkey/malware.go
```diff
@@ -19,22 +19,25 @@ import (
...
 
 func GenerateMalwareKey(info models.MalwareFindingInfo) MalwareKey {
 	return MalwareKey{
-		MalwareName: *info.MalwareName,
-		MalwareType: *info.MalwareType,
-		Path:        *info.Path,
+		MalwareName: utils.StringPointerValOrEmpty(info.MalwareName),
+		MalwareType: utils.StringPointerValOrEmpty(info.MalwareType),
+		Path:        utils.StringPointerValOrEmpty(info.Path),
+		RuleName:    utils.StringPointerValOrEmpty(info.RuleName),
```
Tehsmash, line 41: You'll need to add findingInfo/ruleName back to the $select in the function in assetscanprocessor otherwise this will always be empty.

Tehsmash (changes requested):

installation/docker/yara-rule-server.yaml
```diff
@@ -0,0 +1,9 @@
+enable_json_log: true
+rule_update_schedule: "0 0 * * *"
+rule_sources:
+  - name: "base"
+    url: "https://github.com/Yara-Rules/rules/archive/refs/heads/master.zip"
+    exclude_regex: ".*index.*.yar|.*/utils/.*|.*/deprecated/.*|.*index_.*|.*MALW_AZORULT.yar"
+  - name: "magic"
+    url: "https://github.com/securitymagic/yara/archive/refs/heads/main.zip"
+    exclude_regex: ".*index.*.yar"
```
chrisgacsal, line 9: Missing `\n`.

pkg/shared/families/malware/yara/config/config.go
```diff
@@ -0,0 +1,21 @@
+// Copyright © 2023 Cisco Systems, Inc. and its affiliates.
```
chrisgacsal, line 1: Why we need to put the `config.go` under its own subdirectory? I'd put it under the `yara` directory instead and avoid subdirectories for packages with single files. `Config` could live in the `yara` pacakge so importing it would be more straightforward instead of using aliases.

I prefer this:

```go
import "github.com/openclarity/vmclarity/pkg/shared/families/malware/yara/"

...

type ScannersConfig struct {
        ...
	Yara yara.Config `yaml:"yara" mapstructure:"yara"`
}

```

Instead of this: https://github.com/openclarity/vmclarity/pull/638/files#diff-6efd9504dcd8edbb0bd19fbc5274207a9216b1bb81d5b7ad9f7e87409f3ce152R18-R21

pkg/shared/findingkey/malware.go
```diff
@@ -19,22 +19,25 @@ import (
...
-	return fmt.Sprintf("%s.%s.%s", k.MalwareName, k.MalwareType, k.Path)
+	return fmt.Sprintf("%s.%s.%s.%s", k.MalwareName, k.MalwareType, k.Path, k.RuleName)
 }
 
 func GenerateMalwareKey(info models.MalwareFindingInfo) MalwareKey {
 	return MalwareKey{
-		MalwareName: *info.MalwareName,
-		MalwareType: *info.MalwareType,
-		Path:        *info.Path,
+		MalwareName: utils.StringPointerValOrEmpty(info.MalwareName),
```
chrisgacsal, line 38: Please use `utils.ValueOrZero` instead of `utils.StringPointerValOrEmpty`. Latter is obsoleted by the former.

```suggestion
		MalwareName: utils.ValueOrZero(info.MalwareName),
```

chrisgacsal (changes requested):

pkg/shared/families/malware/yara/config/config.go
pbalogh-sa: It causes some import cycle because we are using them in the common.Config that is used by both yara and clam.
We need to do some refactoring to achieve this.

Tehsmash (approved):

chrisgacsal (approved):

PR #665, opened 2023-09-08
Add support for length() function in ODataSQL

Tehsmash:
## Description

The length function is used to get the length of a collection field as
part of a filter query. For example `$filter=length(assetIds) eq 5`.

This new filter is used to address an issue with the asset statistics query
as it needs to check to see if a list of errors is unset or empty.

## Type of Change

[X] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/pkg/apiserver/database/gorm/odata.go
+++ b/pkg/apiserver/database/gorm/odata.go
@@ -49,12 +49,12 @@ var schemaMetas = map[string]odatasql.SchemaMeta{
 	},
 	"ScanTemplate": {
 		Fields: odatasql.Schema{
-			"scope": odatasql.FieldMeta{FieldType: odatasql.PrimitiveFieldType},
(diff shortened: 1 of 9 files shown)


chrisgacsal (approved):
LGTM!

FrimIdan:
We really need to think of moving the ODataSQL code to a different library repo

akpsgit (approved):

Tehsmash:
> 

I would be down for that.

I was actually thinking that we could rearrange the code too so that the DB driver was underneith the ODATA layer... effectively creating an ODATA native equivalent to gorm. Then all the tables/indexes etc would be managed by the odata layer. Even differences between like mongo/gorm would be underneith that layer so you could interact with all of them the same way using the odata query lanaguage.

PR #669, opened 2023-09-10
feat: extrapolate estimated scan time from scan size

fishkerez:
## Description

In scan estimation, use extrapolation to find static scan time from scan size. 

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[x] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/pkg/orchestrator/provider/aws/scanestimation/estimate.go
+++ b/pkg/orchestrator/provider/aws/scanestimation/estimate.go
@@ -19,8 +19,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math"
-	"time"
 
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
@@ -28,6 +26,7 @@ import (
 	kubeclarityUtils "github.com/openclarity/kubeclarity/shared/pkg/utils"
 
 	"github.com/openclarity/vmclarity/api/models"
+	"github.com/openclarity/vmclarity/pkg/orchestrator/provider/common"
 	familiestypes "github.com/openclarity/vmclarity/pkg/shared/families/types"
 	"github.com/openclarity/vmclarity/pkg/shared/utils"
 )
@@ -78,17 +77,22 @@ const (
 	DataTransfer       recipeResource = "DataTransfer"
 )
 
-// Static times from lab tests of family scan duration in seconds per GB.
+// scanSizesGB represents the memory sizes on the machines that the tests were taken on.
+var scanSizesGB = []float64{0.01, 2.5, 8.1}
+
+// FamilyScanDurationsMap Calculate the logarithmic fit of each family base on static measurements of family scan duration in seconds per scanSizesGB value.
 // The tests were made on a t2.large instance with a gp2 volume.
+// The times correspond to the scan size values in scanSizesGB.
+// TODO add infoFinder family stats.
 // nolint:gomnd
-var familyScanDurationPerGBMap = map[familiestypes.FamilyType]time.Duration{
-	familiestypes.SBOM:             time.Duration(4.5 * float64(time.Second)),
-	familiestypes.Vulnerabilities:  1 * time.Second, // TODO check time with no sbom scan
-	familiestypes.Secrets:          300 * time.Second,
-	familiestypes.Exploits:         0,
-	familiestypes.Rootkits:         0,
-	familiestypes.Misconfiguration: 1 * time.Second,
-	familiestypes.Malware:          360 * time.Second,
+var FamilyScanDurationsMap = map[familiestypes.FamilyType]*common.LogarithmicFormula{
+	familiestypes.SBOM:             common.MustLogarithmicFit(scanSizesGB, []float64{0.01, 11, 37}),
+	familiestypes.Vulnerabilities:  common.MustLogarithmicFit(scanSizesGB, []float64{0.01, 1, 11}), // TODO check time with no sbom scan
+	familiestypes.Secrets:          common.MustLogarithmicFit(scanSizesGB, []float64{0.01, 720, 1320}),
+	familiestypes.Exploits:         common.MustLogarithmicFit(scanSizesGB, []float64{0, 0, 0}),
+	familiestypes.Rootkits:         common.MustLogarithmicFit(scanSizesGB, []float64{0, 0, 0}),
+	familiestypes.Misconfiguration: common.MustLogarithmicFit(scanSizesGB, []float64{0.01, 4, 5}),
+	familiestypes.Malware:          common.MustLogarithmicFit(scanSizesGB, []float64{0.01, 840, 2460}),
 }
 
 // Reserved Instances are not physical instances, but rather a billing discount that is applied to the running On-Demand Instances in your account.
@@ -304,17 +308,6 @@ func findMatchingStatsForInputTypeRootFS(stats *[]models.AssetScanInputScanStats
 	return models.AssetScanInputScanStats{}, false
 }
 
-const constantOfProportionality = 2.5
-
-func calculateStaticScanDuration(familyType familiestypes.FamilyType, scanSizeGB float64) int64 {
-	if scanSizeGB < 1 {
-		// The log of a value less than 1 is negative.
-		return int64(familyScanDurationPerGBMap[familyType].Seconds() * scanSizeGB)
-	}
-
-	return int64(familyScanDurationPerGBMap[familiestypes.SBOM].Seconds() * constantOfProportionality * math.Log(scanSizeGB))
-}
-
 // nolint:cyclop
 func getScanDuration(stats models.AssetScanStats, familiesConfig *models.ScanFamiliesConfig, scanSizeMB int64) int64 {
 	var totalScanDuration int64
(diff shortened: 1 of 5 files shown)


pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -78,17 +78,84 @@ const (
...
+	lnXData := make([]float64, n)
+
+	for i := 0; i < n; i++ {
+		if xData[i] <= 0 {
+			return 0, 0, fmt.Errorf("input data contains non-positive x values: %v", xData)
+		}
+		lnXData[i] = math.Log(xData[i])
+	}
+
+	// Calculate the coefficients of the linear regression model y = a * ln(x) + b
```
Tehsmash, line 146: Above you call this a logarithmic model, and here you call it a linear regression model. Which is it?

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -78,17 +78,84 @@ const (
...
+	familiestypes.Secrets:          {0.01, 720, 1320},
+	familiestypes.Exploits:         {0, 0, 0},
+	familiestypes.Rootkits:         {0, 0, 0},
+	familiestypes.Misconfiguration: {0.01, 4, 5},
+	familiestypes.Malware:          {0.01, 840, 2460},
+}
+
+// familyCoefficients populated during init in order to extrapolate the scan duration from scan size per family.
+// The extrapolation formula is y = a * ln(x) + b.
+var familyCoefficients = make(map[familiestypes.FamilyType][]float64, len(families))
```
Tehsmash, line 103: If each coefficient is a known set of two float64s, I would much rather we use a struct here.

```
// struct representing a model for a specific family
type familyCofficient struct {
   a float64
   b float64
}
```

we could even add the functions on the struct:

```
func (fc *familyCofficient) calculateDuration(sizeInGb float64) (float64) {
     return fc.a * sizeInGb + fc.b
}
```

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -304,15 +371,8 @@ func findMatchingStatsForInputTypeRootFS(stats *[]models.AssetScanInputScanStats
...
-const constantOfProportionality = 2.5
-
 func calculateStaticScanDuration(familyType familiestypes.FamilyType, scanSizeGB float64) int64 {
-	if scanSizeGB < 1 {
-		// The log of a value less than 1 is negative.
-		return int64(familyScanDurationPerGBMap[familyType].Seconds() * scanSizeGB)
-	}
-
-	return int64(familyScanDurationPerGBMap[familiestypes.SBOM].Seconds() * constantOfProportionality * math.Log(scanSizeGB))
+	return int64(familyCoefficients[familyType][0]*scanSizeGB + familyCoefficients[familyType][1])
```
Tehsmash, line 375: Shouldn't we be taking the natural log of scanSizeGB here?

Tehsmash (changes requested):

pkg/orchestrator/provider/aws/scanestimation/estimate.go
akpsgit: logarithmic regression :)

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -78,17 +78,84 @@ const (
...
+	// Calculate the coefficients of the linear regression model y = a * ln(x) + b
+	var sumLnX, sumY, sumLnX2, sumLnXY float64
+	for i := 0; i < n; i++ {
+		sumLnX += lnXData[i]
+		sumY += yData[i]
+		sumLnX2 += lnXData[i] * lnXData[i]
+		sumLnXY += lnXData[i] * yData[i]
+	}
+
+	a := ((float64(n))*sumLnXY - sumLnX*sumY) / ((float64(n))*sumLnX2 - sumLnX*sumLnX)
```
akpsgit, line 155: I think that worth adding a non zero check for ((float64(n))*sumLnX2 - sumLnX*sumLnX) to be on the safe side

pkg/orchestrator/provider/aws/scanestimation/estimate.go
akpsgit: agree

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -78,17 +78,84 @@ const (
...
+
+func init() {
+	if err := initFamilyScanDurationCoefficients(); err != nil {
+		logrus.Errorf("Failed to init family scan duration coefficients: %v", err)
+	}
+}
+
+func initFamilyScanDurationCoefficients() error {
+	var err error
+	var xCoefficient, yCoefficient float64
```
akpsgit, line 113: they are not coefficients, i think better xCoefficient, yCoefficient - > a, b

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -78,17 +78,84 @@ const (
...
+// The extrapolation formula is y = a * ln(x) + b.
+var familyCoefficients = make(map[familiestypes.FamilyType][]float64, len(families))
+
+func init() {
+	if err := initFamilyScanDurationCoefficients(); err != nil {
+		logrus.Errorf("Failed to init family scan duration coefficients: %v", err)
+	}
+}
+
+func initFamilyScanDurationCoefficients() error {
```
akpsgit, line 111: I think that it's better to call it initFamilyScanDurationEstimators/Fits/Curves/etc

pkg/orchestrator/provider/aws/scanestimation/estimate.go
```diff
@@ -78,17 +78,84 @@ const (
...
+	var err error
+	var xCoefficient, yCoefficient float64
+	// xData represents the used memory size in GB.
+	xData := []float64{0.01, 2.5, 8.1}
```
akpsgit, line 115: I don't think that we should keep the x data in one place. I think that it's better to keep the x and y data together even if it's duplicated. It will also allow us not to couple the measurements of the families

pkg/orchestrator/provider/aws/scanestimation/estimate.go
fishkerez: Good catch thank you

akpsgit: Right, according to the example:
```
	xToExtrapolate := 22.5
	lnXToExtrapolate := math.Log(xToExtrapolate)
	y := a*lnXToExtrapolate + b

	fmt.Printf("Y=%.2f\n", y)
```

pkg/orchestrator/provider/common/utils.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"fmt"
+	"math"
+
+	"github.com/sirupsen/logrus"
+
+	familiestypes "github.com/openclarity/vmclarity/pkg/shared/families/types"
+)
+
+// scanSizesGB represents the memory sizes on the machines that the tests were taken on.
```
akpsgit, line 27: I think that this section (up to type LogarithmicFormula struct {) should be in the aws package

pkg/orchestrator/provider/common/utils.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// LogarithmicFormula represents the formula y = a * ln(x) + b.
+type LogarithmicFormula struct {
+	a float64
+	b float64
+}
+
+// Extract returns receive an x value and returns the y value of the formula.
+func (lf *LogarithmicFormula) Extract(x float64) float64 {
```
akpsgit, line 52: maybe Evaluate()?

pkg/orchestrator/provider/common/utils.go
```diff
@@ -0,0 +1,109 @@
...
+	return lf.a*math.Log(x) + lf.b
+}
+
+func initFamilyScanDurationEstimators(scanSizes, scanTimes []float64) *LogarithmicFormula {
```
akpsgit, line 56: doesn't really matter what we evaluate here, I think it should be:
func initFamilyScanDurationEstimators(xData, yData []float64) *LogarithmicFormula {

pkg/orchestrator/provider/common/utils.go
```diff
@@ -0,0 +1,109 @@
...
+	return ret
+}
+
+func LogarithmicFit(xData, yData []float64) (*LogarithmicFormula, error) {
+	var err error
+	var a, b float64
+
+	a, b, err = logarithmicFit(xData, yData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to calculate least squares fitting: %w", err)
```
akpsgit, line 70: failed to calculate a logarithmic fit

pkg/orchestrator/provider/common/utils.go
akpsgit: The idea was to create 2 functions, one to be used as a lib, and another specific to aws to be used during init. If you want to move it to aws then it's o.k. but it's a private function in the utils section that is used by AWS logic. it doesn't make sense

pkg/orchestrator/provider/common/utils.go
```diff
@@ -0,0 +1,89 @@
...
+
+// LogarithmicFormula represents the formula y = a * ln(x) + b.
+type LogarithmicFormula struct {
+	a float64
+	b float64
+}
+
+// Evaluate receive an x value and returns the y value of the formula.
+func (lf *LogarithmicFormula) Evaluate(x float64) float64 {
+	return lf.a*math.Log(x) + lf.b
```
Tehsmash, line 33: pretty surprised this isn't a linter error.
```suggestion
	return lf.a * math.Log(x) + lf.b
```

Tehsmash (approved):

akpsgit (approved):

PR #672, opened 2023-09-11
scan estimation time with parallel scanners

fishkerez:
## Description

Take number of parallel scanners into consideration of scan time estimation

## Type of Change

[ ] Bug Fix  
[* ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -512,6 +514,24 @@ func (w *Watcher) reconcileInProgress(ctx context.Context, scanEstimation *model
...
+		// find the number of parallel scanners that needs to run for this scan.
+		var actualParallelScanners int
+		if maxParallelScanners <= numberOfJobs {
+			actualParallelScanners = maxParallelScanners
+		} else {
+			actualParallelScanners = numberOfJobs
+		}
+
+		// Note: This is a rough estimation, as we don't know which jobs will be running in parallel.
+		*scanEstimation.Summary.TotalScanTime = *scanEstimation.Summary.TotalScanTime / actualParallelScanners
```
Tehsmash, line 531: I think we said that the best way to calculate this was to calculate the mean time taken by the scanners to adjust for scanners that take a very long time and a very short time and how they are distributed between the batches. Then multiply the mean scanner time by number of batches to get the TotalScanTime.

TotalScanTime  = (SUM(Scanner Times) / NumberOfAssets) * (NumberOfAssets / MaxParallelScanners)

Tehsmash (changes requested):

pkg/orchestrator/scanestimationwatcher/watcher.go
akpsgit: isn't it the same? :)

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -512,6 +514,24 @@ func (w *Watcher) reconcileInProgress(ctx context.Context, scanEstimation *model
 	return nil
 }
 
+func updateTotalScanTimeWithParallelScans(scanEstimation *models.ScanEstimation) {
+	if scanEstimation.ScanTemplate.MaxParallelScanners != nil {
+		maxParallelScanners := *scanEstimation.ScanTemplate.MaxParallelScanners
+		numberOfJobs := *scanEstimation.Summary.JobsCompleted
+
+		// find the number of parallel scanners that needs to run for this scan.
```
akpsgit, line 522: maybe just 
```
*scanEstimation.Summary.TotalScanTime = *scanEstimation.Summary.TotalScanTime / Min(*scanEstimation.ScanTemplate.MaxParallelScanners, *scanEstimation.Summary.JobsCompleted) 
```

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -512,6 +514,24 @@ func (w *Watcher) reconcileInProgress(ctx context.Context, scanEstimation *model
 	return nil
 }
 
+func updateTotalScanTimeWithParallelScans(scanEstimation *models.ScanEstimation) {
+	if scanEstimation.ScanTemplate.MaxParallelScanners != nil {
```
chrisgacsal, line 518:
```suggestion
	 if scanEstimation == nil || scanEstimation.ScanTemplate == nil {
            return
        }
	
	maxParallelScanners := utils.ValueOrZero(scanEstimation.ScanTemplate.MaxParallelScanners)
	
	if maxParallelScanners > 1 {
```

chrisgacsal (changes requested):

pkg/orchestrator/scanestimationwatcher/watcher.go
Tehsmash: I don't think these are the same at all because the current calculation isn't taking into account how the parallel scanners interact.

Take the scenario when you have maxParallelScanners set to 3, and you are scanning 10 items. To simplify we'll assume that all tasks take the same time, 3 minutes, (in my calculation I use the mean task time instead).

In my calculation, we'll need to run 4 batches (3, 3, 3, 1) of tasks one after the other to complete all 10 items. So the total estimated time taken is 4 batches * task time (3 minutes) = 12 minutes

The current calculation in this PR would use the total time (10 * 3  = 30 minutes) / 3 parallel scanners = 10 minutes, which is under estimating the time.

fishkerez: @Tehsmash I see what you mean, only problem is that in the example you gave, I don't think it does this:

> TotalScanTime = (SUM(Scanner Times) / NumberOfAssets) * (NumberOfAssets / MaxParallelScanners)

It is the same as writing  `TotalScanTime = SUM(Scanner Times)* MaxParallelScanners` no? 
Since you divide and then multiply by NumberOfAsset.

I think what you suggest is that we calculate the mean job time by `SUM(Scanner Times) / NumberOfAssets`
and then we calculate the number of batches by:

`if numberOfAsset < maxParallelScanners: numBatches=maxParallelScanners
else numBatches=numberOfAssets/maxParallelScanners + 1 (or 0)`

Tehsmash: This is not the same as we're not multiplying by NumberOfAssets, we're multiplying by the number of batches.
Number of batches should be:

math.Ceil(NumberOfAssets/MaxParallelScanners) so:
  - math.Ceil(10 / 3) = 4
  - math.Ceil(10 / 1) = 10 
  - math.Ceil(5 / 5) = 1
 
Final calculation should be something like:

```
meanScannerTime := sum(assetScanTimes) * maxParallelScanners
numberOfBatches := math.Ceil(numberOfAssets / maxParallelScanners)
totalScanTime = meanScannerTime * numberOfBatches
```

Tehsmash: Jobs do come in an out of the pool, but if you normalise the job times by using the mean you can remove this factor. In this case all the jobs in each batch take exactly the same time and there for you get good estimate between the worst case scenario: the 4 longest jobs run back to back and the best case scenario all the longest jobs run concurrently.

Tehsmash: if we want to use facts to decide, then we should look at the real times in the lab and compare to see which is a more accurate estimation.

Tehsmash: Looks like running through some experiments that this algorithm underestimates, and my algorithm overestimates depending on how the mean works out. We need to do some more research here. I think we can merge this as is for now.

One thought we had for calculating the TotalTime is to use a packing algorithm to calculate the best case scenario, i.e. given all the scanner times what is the least amount of time they could take to run, then calculate the worst case scenario, then take the average of the two.

akpsgit (approved):

akpsgit (approved):

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -512,6 +515,23 @@ func (w *Watcher) reconcileInProgress(ctx context.Context, scanEstimation *model
 	return nil
 }
 
+func updateTotalScanTimeWithParallelScans(scanEstimation *models.ScanEstimation) {
```
chrisgacsal, line 518: Should this function actually return an `error` if the `scanEstimation` is `nil` or one of its fields like `ScanTemplate` or `Summary`? And let the caller to decide how to handle the error?

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -512,6 +515,23 @@ func (w *Watcher) reconcileInProgress(ctx context.Context, scanEstimation *model
...
+	if scanEstimation == nil || scanEstimation.ScanTemplate == nil {
+		return
+	}
+
+	maxParallelScanners := utils.ValueOrZero(scanEstimation.ScanTemplate.MaxParallelScanners)
+
+	if maxParallelScanners > 1 {
+		numberOfJobs := *scanEstimation.Summary.JobsCompleted
```
chrisgacsal, line 526:
```suggestion
                var numberOfJobs int
                
                if scanEstimation.Summary != nil {
                    numberOfJobs = utils.ValueOrZero(scanEstimation.Summary.JobsCompleted)
                }
```

or 

```suggestion
                if scanEstimation.Summary == nil {
                    return
                }
                
                numberOfJobs := utils.ValueOrZero(scanEstimation.Summary.JobsCompleted)
```

chrisgacsal (changes requested):

pkg/orchestrator/scanestimationwatcher/watcher.go
fishkerez: I was not checking this field for nil since this is an internal function that being called once, and in the context it is being called, scanEstimation.Summary is already set and cannot be nil. 
Do you think we still need to check for nil in that case?

pkg/orchestrator/scanestimationwatcher/watcher.go
fishkerez: I can add an error return here in that case.

Tehsmash (approved):
LGTM, there are linter errors to be fixed though

akpsgit (approved):

chrisgacsal (approved):
LGTM! Thanks for the update.

PR #688, opened 2023-09-13
add ttl to scanEstimation and assetScanEstimation

fishkerez:
## Description

Following https://github.com/openclarity/vmclarity/pull/603 needed to clean up scanEstimation and assetScanEstimation resources after a period of times.
This PR adds ttl to these resources and clean them up from the DB after this time. 
The default ttl for both resources (if not set) is 2 hours

## Type of Change

[ ] Bug Fix  
[* ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

- [ *] I have read the [contributing guidelines](https://github.com/openclarity/vmclarity/blob/main/CONTRIBUTING.md)
- [* ] Existing issues have been referenced (where applicable)


api/openapi.yaml
```diff
@@ -2888,6 +2905,9 @@ components:
           $ref: '#/components/schemas/ScanEstimationSummary'
         revision:
           type: integer
+        ttl:
+          description: The duration in seconds this resource should last until it is deleted.
```
chrisgacsal, line 2909: What do you think about overriding the field name by adding the `x-go-name: TTL` directive?

```suggestion
          description: The duration in seconds this resource should last until it is deleted.
          x-go-name: TTL
```

chrisgacsal: It makes the field named in the generated code `TTL` and not `Ttl` where the later is not the way how you would name a exported field for a struct in Go.

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -78,8 +78,8 @@ func (w *Watcher) GetAssetScanEstimations(ctx context.Context) ([]AssetScanEstim
 	logger := log.GetLoggerFromContextOrDiscard(ctx)
 	logger.Debugf("Fetching AssetScanEstimations which need to be reconciled")
 
-	filter := fmt.Sprintf("state/state ne '%s' and state/state ne '%s'",
-		models.AssetScanEstimationStateStateDone, models.AssetScanEstimationStateStateFailed)
+	filter := fmt.Sprintf("(state/state ne '%s' and state/state ne '%s') or (deleteAfter eq null or '%s' > deleteAfter)",
```
FrimIdan, line 81: When comparing time, you should use `gt` and not `>`.
Please look at the following for example
```
(lastSeen eq null or lastSeen lt %s)", discoveryTime.Format(time.RFC3339)))
```

api/openapi.yaml
chrisgacsal: Thanks for adopting.

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -237,9 +284,20 @@ func (w *Watcher) reconcilePending(ctx context.Context, assetScanEstimation *mod
 
 	assetScanEstimation.State.LastTransitionTime = utils.PointerTo(time.Now())
 
+	// Set default ttl if not set.
+	if assetScanEstimation.TTLSecondsAfterFinished == nil {
+		assetScanEstimation.TTLSecondsAfterFinished = utils.PointerTo(DefaultAssetScanEstimationTTLSeconds)
+	}
+	endTime := time.Now()
```
FrimIdan, line 291: can you explain why the `endTime` and `startTime` is now?

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -237,9 +284,20 @@ func (w *Watcher) reconcilePending(ctx context.Context, assetScanEstimation *mod
 
 	assetScanEstimation.State.LastTransitionTime = utils.PointerTo(time.Now())
```
FrimIdan, line 285: multiple call to `time.Now()` in the same flow, probably can be called only once.

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -80,7 +80,8 @@ func (w *Watcher) GetScanEstimations(ctx context.Context) ([]ScanEstimationRecon
 	logger := log.GetLoggerFromContextOrDiscard(ctx)
 	logger.Debugf("Fetching running ScanEstimations")
 
-	filter := fmt.Sprintf("state/state ne '%s' and state/state ne '%s'", models.ScanEstimationStateStateDone, models.ScanEstimationStateStateFailed)
+	filter := fmt.Sprintf("(state/state ne '%s' and state/state ne '%s') or (deleteAfter eq null or '%s' > deleteAfter)",
```
FrimIdan, line 83: same

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -160,13 +160,60 @@ func (w *Watcher) Reconcile(ctx context.Context, event AssetScanEstimationReconc
...
 	return nil
 }
 
+func (w *Watcher) reconcileDone(ctx context.Context, assetScanEstimation *models.AssetScanEstimation) error {
+	if assetScanEstimation.EndTime == nil {
+		// TODO error?
+		return nil
+	}
+	if assetScanEstimation.TtlSecondsAfterFinished == nil {
+		// TODO error?
```
FrimIdan, line 178: why not just use the default?

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -160,13 +160,60 @@ func (w *Watcher) Reconcile(ctx context.Context, event AssetScanEstimationReconc
...
+		}
 	default:
 	}
 
 	return nil
 }
 
+func (w *Watcher) reconcileDone(ctx context.Context, assetScanEstimation *models.AssetScanEstimation) error {
+	if assetScanEstimation.EndTime == nil {
+		// TODO error?
```
FrimIdan, line 174: why not just use `now()`?

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -248,10 +293,16 @@ func (w *Watcher) reconcilePending(ctx context.Context, scanEstimation *models.S
...
+	// Set default ttl if not set.
+	if scanEstimation.TTLSecondsAfterFinished == nil {
+		scanEstimation.TTLSecondsAfterFinished = utils.PointerTo(DefaultScanEstimationTTLSeconds)
+	}
+
 	scanEstimationPatch := &models.ScanEstimation{
-		StartTime: utils.PointerTo(time.Now()),
-		AssetIDs:  scanEstimation.AssetIDs,
-		State:     scanEstimation.State,
+		StartTime:               utils.PointerTo(time.Now()),
```
FrimIdan, line 302: you are missing the `endTime` and the `DeleteAfter` here?

pkg/orchestrator/assetscanestimationwatcher/config.go
```diff
@@ -23,9 +23,10 @@ import (
 )
 
 const (
-	DefaultPollInterval     = 5 * time.Second
-	DefaultReconcileTimeout = 15 * time.Second
-	DefaultAbortTimeout     = 1 * time.Minute
+	DefaultPollInterval                  = 5 * time.Second
+	DefaultReconcileTimeout              = 15 * time.Second
+	DefaultAbortTimeout                  = 1 * time.Minute
+	DefaultAssetScanEstimationTTLSeconds = 7200 // 2 hours
```
chrisgacsal, line 29:
```suggestion
	DefaultAssetScanEstimationTTLSeconds = 7200 * time.Second // 2 hours
```

pkg/orchestrator/assetscanestimationwatcher/watcher.go
fishkerez: Thanks, wasn't sure about it, wanted to check this logic but the VMClarity images was broken

chrisgacsal (changes requested):

pkg/orchestrator/assetscanestimationwatcher/config.go
chrisgacsal: Ahh, nevermind. It'd require a bit more change.

pkg/orchestrator/assetscanestimationwatcher/watcher.go
fishkerez: It is basically because the action of Estimate is really fast, but you are right, I will move the startTime to before calling Estimate

pkg/orchestrator/assetscanestimationwatcher/watcher.go
fishkerez: Now this times are not related :) they are being called in different phases in the code.

pkg/orchestrator/assetscanestimationwatcher/watcher.go
fishkerez: I can, but this might hide something not good that happened, since it should have been set before

pkg/orchestrator/scanestimationwatcher/watcher.go
fishkerez: No, This is only pending reconcile, the endTime and deleteAfter are set when the scan is done (see reconcileInProgress)

pkg/orchestrator/assetscanestimationwatcher/watcher.go
FrimIdan: not sure what is better, if you just return a `nil` the object can stay in reconciling loop forever.
You can print an error and set the defaults, this way you will not hide anything

Tehsmash: Ideally the reconcile loop should fix things like this if it notices it, probably outside of any specific state processing. I.e. 
```
reconcile() {

if assetScanEstimation.TtlSecondsAfterFinished == nil {
   set default
}

set any other missing defaults

patch assetScanEstimation

switch assetScanEstimation.State {
 case pending:
    reconcilePending(...)
 case in progress:
    ...
 case done:
   ...
}

}
```

pkg/orchestrator/assetscanestimationwatcher/config.go
Tehsmash: could have been `2 * time.Hour` but nevermind...

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -78,8 +78,8 @@ func (w *Watcher) GetAssetScanEstimations(ctx context.Context) ([]AssetScanEstim
 	logger := log.GetLoggerFromContextOrDiscard(ctx)
 	logger.Debugf("Fetching AssetScanEstimations which need to be reconciled")
 
-	filter := fmt.Sprintf("state/state ne '%s' and state/state ne '%s'",
-		models.AssetScanEstimationStateStateDone, models.AssetScanEstimationStateStateFailed)
+	filter := fmt.Sprintf("(state/state ne '%s' and state/state ne '%s') or (deleteAfter eq null or '%s' > deleteAfter)",
+		models.AssetScanEstimationStateStateDone, models.AssetScanEstimationStateStateFailed, time.Now().String())
```
Tehsmash, line 82: The time here needs to rendered into RFC3339 e.g. `discoveryTime.Format(time.RFC3339)`

Tehsmash, line 82: The time here needs to rendered into RFC3339 e.g. discoveryTime.Format(time.RFC3339)

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -237,9 +284,20 @@ func (w *Watcher) reconcilePending(ctx context.Context, assetScanEstimation *mod
 
 	assetScanEstimation.State.LastTransitionTime = utils.PointerTo(time.Now())
 
+	// Set default ttl if not set.
+	if assetScanEstimation.TTLSecondsAfterFinished == nil {
+		assetScanEstimation.TTLSecondsAfterFinished = utils.PointerTo(DefaultAssetScanEstimationTTLSeconds)
+	}
```
Tehsmash, line 290: I think we should probably rename reconcileNoState to reconcileDefaults and always run it before the rest of the reconciling logic. This and the state default can be processed there.

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -80,7 +80,8 @@ func (w *Watcher) GetScanEstimations(ctx context.Context) ([]ScanEstimationRecon
 	logger := log.GetLoggerFromContextOrDiscard(ctx)
 	logger.Debugf("Fetching running ScanEstimations")
 
-	filter := fmt.Sprintf("state/state ne '%s' and state/state ne '%s'", models.ScanEstimationStateStateDone, models.ScanEstimationStateStateFailed)
+	filter := fmt.Sprintf("(state/state ne '%s' and state/state ne '%s') or (deleteAfter eq null or '%s' > deleteAfter)",
+		models.ScanEstimationStateStateDone, models.ScanEstimationStateStateFailed, time.Now().String())
```
Tehsmash, line 84: RFC3339

pkg/orchestrator/scanestimationwatcher/watcher.go
```diff
@@ -169,15 +170,59 @@ func (w *Watcher) Reconcile(ctx context.Context, event ScanEstimationReconcileEv
...
 	return nil
 }
 
+func (w *Watcher) reconcileDone(ctx context.Context, scanEstimation *models.ScanEstimation) error {
+	if scanEstimation.EndTime == nil {
+		return nil
+	}
+	if scanEstimation.TTLSecondsAfterFinished == nil {
+		return nil
+	}
```
Tehsmash, line 189: same as AssetScanEstimations we should reconcile defaults for these before this point, and if they are not set it should be an error

Tehsmash (changes requested):

pkg/orchestrator/assetscanestimationwatcher/watcher.go
```diff
@@ -160,13 +160,58 @@ func (w *Watcher) Reconcile(ctx context.Context, event AssetScanEstimationReconc
...
 	return nil
 }
 
+func (w *Watcher) reconcileDone(ctx context.Context, assetScanEstimation *models.AssetScanEstimation) error {
+	if assetScanEstimation.EndTime == nil {
+		assetScanEstimation.EndTime = utils.PointerTo(time.Now())
+	}
+	if assetScanEstimation.TTLSecondsAfterFinished == nil {
+		assetScanEstimation.TTLSecondsAfterFinished = utils.PointerTo(DefaultAssetScanEstimationTTLSeconds)
+	}
```
Tehsmash, line 178: Shouldn't we patch these back to the API if they are missing?

Tehsmash (approved):

PR #689, opened 2023-09-13
feat(scanresult): resource cleanup status refactor

adamtagscherer:
## Description

Updates the `ResourceCleanupState` for the new object, and also enforces the proper state transitions when the API is called.

## Type of Change

[x] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[x] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


pkg/apiserver/database/gorm/asset_scan.go
```diff
@@ -112,6 +112,13 @@ func (s *AssetScansTableHandler) CreateAssetScan(assetScan models.AssetScan) (mo
 		}
 	}
 
+	// Check the user wants to create a resource cleanup with other status than pending
```
fishkerez, line 115: I think you need to check for nil values here:
`if assetScan.ResourceCleanup == nil || assetScan.ResourceCleanup.State == nil {...}`

Beside, if ResourceCleanup is not set by the user, we need to set it to Pending

pkg/apiserver/rest/asset_scan_controller.go
```diff
@@ -83,14 +83,20 @@ func (s *ServerImpl) PatchAssetScansAssetScanID(ctx echo.Context, assetScanID mo
...
+	existingAssetScan, err := s.dbHandler.AssetScansTable().GetAssetScan(assetScanID, models.GetAssetScansAssetScanIDParams{})
 	if err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
 			return sendError(ctx, http.StatusNotFound, fmt.Sprintf("asset scan was not found. assetScanID=%v: %v", assetScanID, err))
 		}
 		return sendError(ctx, http.StatusInternalServerError, fmt.Sprintf("failed to get asset scan. assetScanID=%v: %v", assetScanID, err))
 	}
 
+	// check for valid resource cleanup state transition
+	err = existingAssetScan.ResourceCleanup.UpdateState(*assetScan.ResourceCleanup.State)
```
fishkerez, line 95: Need to check 
`if assetScan.ResourceCleanup != nil && assetScan.ResourceCleanup.State != nil`

This is a patch operation and the ResourceCleanup field will not always be set

pkg/orchestrator/assetscanwatcher/watcher.go
```diff
@@ -195,8 +195,8 @@ func (w *Watcher) reconcilePending(ctx context.Context, assetScan *models.AssetS
 	if assetScan.Scan != nil && assetScan.Scan.Id != "" && assetScan.Scan.MaxParallelScanners != nil {
 		// Check whether we have reached the maximum number of running scans
 		// TODO(chrisgacsal): the number of concurrent scans needs to be part of the provider config and handled there
-		filter := fmt.Sprintf("scan/id eq '%s' and status/general/state ne '%s' and status/general/state ne '%s' and resourceCleanup eq '%s'",
-			assetScan.Scan.Id, models.AssetScanStateStateDone, models.AssetScanStateStatePending, models.ResourceCleanupStatePending)
+		filter := fmt.Sprintf("scan/id eq '%s' and status/general/state ne '%s' and status/general/state ne '%s' and resourceCleanup/state eq '%s'",
+			assetScan.Scan.Id, models.AssetScanStateStateDone, models.AssetScanStateStatePending, models.ResourceCleanupStatusStatePending)
```
fishkerez, line 199: Can you please explain why only resources with cleanup state pending are considered to be running?

pkg/orchestrator/assetscanwatcher/watcher.go
```diff
@@ -330,10 +330,18 @@ func (w *Watcher) cleanupResources(ctx context.Context, assetScan *models.AssetS
 
 	switch w.scannerConfig.DeleteJobPolicy {
 	case DeleteJobPolicyNever:
-		assetScan.ResourceCleanup = utils.PointerTo(models.ResourceCleanupStateSkipped)
+		assetScan.ResourceCleanup = &models.ResourceCleanupStatus{
```
fishkerez, line 333: Worth having a function here IMO
```
func createStatus(LastTransitionTime time.Time, message, reason string, state models.ResourceCleanupStatusState) *models.ResourceCleanupStatus {
	return &models.ResourceCleanupStatus{
		LastTransitionTime: utils.PointerTo(LastTransitionTime),
		Message:            utils.PointerTo(message),
		Reason:             utils.PointerTo(reason),
		State:              utils.PointerTo(state),
	}
}

```

pkg/apiserver/rest/asset_scan_controller.go
chrisgacsal: I'd probably add a getter to the `AssetScan` the same way as we have for other objects to get their fields safely.

```suggestion
	func (s *AssetScan) ResourceCleanup() (*ResourceCleanup, bool) {
	    if s.ResourceCleanup == nil || s.ResourceCleanup.State == nil {
	        nil, false
	    }
	    
	    return s.ResourceCleanup, true
	}
```

And use this method for checking whether the `ResourceCleanup` object is valid or not.

```suggestion
        if _, ok := existingAssetScan.ResourceCleanup(); !ok {
            return ...
        }
        
	err = existingAssetScan.ResourceCleanup.UpdateState(*assetScan.ResourceCleanup.State)
```

pkg/apiserver/rest/asset_scan_controller.go
```diff
@@ -132,14 +138,20 @@ func (s *ServerImpl) PutAssetScansAssetScanID(ctx echo.Context, assetScanID mode
...
```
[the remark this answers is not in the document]
chrisgacsal: I do not necessary see why this would be a problem unless we do not use the  `existingAssetScan` to update the object in the database.

However I agree that could be confusing, so probably the best to create a temporary variable string the `existingAssetScan.ResourceCleanup` and performing the status update on this object so this way it does not modify the `existingAssetScan` object itself avoiding possible future bugs when someone wants to reuse the `existingAssetScan` without realizing it has been modified for validation purposes.

pkg/orchestrator/assetscanwatcher/watcher.go
chrisgacsal: The reason for this is we want to make sure that resource cleanup also taken into consideration when determining the number of allowed assets being scanned to prevent exceeding the resources certain providers have.

This is a result of a long discussion with @Tehsmash when resource cleanup was introduced.

pkg/orchestrator/assetscanwatcher/watcher.go
chrisgacsal: Agreed., but I would name the contructor as `NewResourceCleanupStatus(...)` to make it aligned with the naming convention Golang has.

chrisgacsal (changes requested):

chrisgacsal (approved):
LGTM! :shipit:

chrisgacsal (approved):

pkg/apiserver/rest/asset_scan_controller.go
```diff
@@ -83,14 +91,23 @@ func (s *ServerImpl) PatchAssetScansAssetScanID(ctx echo.Context, assetScanID mo
...
+	existingAssetScan, err := s.dbHandler.AssetScansTable().GetAssetScan(assetScanID, models.GetAssetScansAssetScanIDParams{})
 	if err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
 			return sendError(ctx, http.StatusNotFound, fmt.Sprintf("asset scan was not found. assetScanID=%v: %v", assetScanID, err))
 		}
 		return sendError(ctx, http.StatusInternalServerError, fmt.Sprintf("failed to get asset scan. assetScanID=%v: %v", assetScanID, err))
 	}
 
+	// check for valid resource cleanup state transition
+	_, ok := assetScan.GetResourceCleanupStatus()
```
chrisgacsal, line 103:
```suggestion
	currentResourceCleanupStatus, ok := assetScan.GetResourceCleanupStatus()
	...
			err = currentResourceCleanupStatus.UpdateState(*assetScan.ResourceCleanupStatus.State)
```

Just to avoid updating the `ResourceCleanupStatus` for the current `AssetScan` since this step is only for validation the requested state transition.

pkg/apiserver/rest/asset_scan_controller.go
```diff
@@ -132,14 +149,23 @@ func (s *ServerImpl) PutAssetScansAssetScanID(ctx echo.Context, assetScanID mode
...
+	if ok {
+		err = existingAssetScan.ResourceCleanupStatus.UpdateState(*assetScan.ResourceCleanupStatus.State)
```
chrisgacsal, line 163: Same as above. ^

chrisgacsal (changes requested):

api/models/status.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+type Defaulter interface {
+	Default()
+}
+
+func NewResourceCleanupStatus(state ResourceCleanupStatusState, m string, s ...string) *ResourceCleanupStatus {
+	lastTransitionTime := time.Now()
+	var reason string
+	if len(s) == 1 {
```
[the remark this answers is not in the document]
adamtagscherer: To make it optional, because most of the time there is no error message when creatin a `ResourceCleanupStatus` entity and we would just pass `nil`s here.

pkg/orchestrator/assetscanwatcher/watcher.go
```diff
@@ -367,21 +373,32 @@ func (w *Watcher) cleanupResources(ctx context.Context, assetScan *models.AssetS
 		var retryableError provider.RetryableError
 		switch {
 		case errors.As(err, &fatalError):
-			assetScan.ResourceCleanup = utils.PointerTo(models.ResourceCleanupStateFailed)
+			assetScan.ResourceCleanupStatus = models.NewResourceCleanupStatus(
+				models.ResourceCleanupStatusStateFailed,
+				"Resource cleanup failed.",
+				fatalError.Error(),
```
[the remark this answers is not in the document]
adamtagscherer: By machine readable we mean something that was generated by an error and not handwritten, and here this happens.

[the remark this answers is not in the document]
Tehsmash: By "machine readable" we mean something that is machine parsable as a simple string like the "reason" field from Kubernetes conditions. https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions

For example it could be something like "UnexpectedError" or "ResourceCleanupFailed" or "OneOrMoreCleanupStepsFailed".

These are machine parsable fixed strings so that it doesn't matter what human readable text is in the "message" field. The client code reading this field can do:
```
if err.Reason == "UnexpectedError" {
 ...
}
```

api/models/status.go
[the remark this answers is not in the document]
Tehsmash: I think Reason should always have a value, there is always a reason why it moved from the previous state to this state even  if its just "CompletedCleanup".

api/models/status.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return fmt.Errorf("invalid transition: %s > %s", *r.State, toState)
+}
+
+func (r *ResourceCleanupStatus) Default() {
+	now := time.Now()
+	state := ResourceCleanupStatusStatePending
+
+	if r == nil {
+		*r = *NewResourceCleanupStatus(state, "Default ResourceCleanupStatus created.")
```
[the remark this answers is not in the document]
Tehsmash: This is change the location the pointer is pointing to, its not dereferencing it, this isn't too uncommon in the golang world for these kind of usecases.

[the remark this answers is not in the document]
fishkerez: A simple unit test:
```
func TestResourceCleanupStatus_Default(t *testing.T) {
	var r *ResourceCleanupStatus
	r.Default()
}

```

Will crash, so I am not sure what do you mean

fishkerez:
LGTM!
General question: The PR description states: `Creates the base status API object in the openapi spec.`
But in the PR we are creating a specific status object for resourceCleanup. 
I think it is a good idea to make the status object more general since we use it a lot in all the objects.

api/openapi.yaml
```diff
@@ -2047,13 +2047,54 @@ components:
...
+        reason:
+          description: |
+            Machine readable reason for state transition.
+
+            | State   | Reason          | Description                                                |
+            | ------- | --------------- | ---------------------------------------------------------- |
+            | Pending | AssetScanCreate | AssetScan created                                          |
+            | Skipped | DeletePolicy    | Resource cleanup has been skipped due to Delete Job Policy |
+            | Failed  | ProviderError   | Failed due to Provider error                               |
+            | Failed  | InternalError   | Failed due to internal error                               |
```
Tehsmash, line 2082: do we have a cleanup timeout? I can't remember :/

api/openapi.yaml
```diff
@@ -2047,13 +2047,54 @@ components:
...
-        - Pending
-        - Skipped
-        - Failed
-        - Done
+    ResourceCleanupStatus:
+      type: object
+      required:
+        - state
+        - reason
+        - lastTransitionTime
```
Tehsmash, line 2055: If you make these required then it is no longer possible to PATCH just one of these fields, I assume that we should always be PATCHing all of them when making a change?

pkg/apiserver/rest/asset_scan_controller.go
```diff
@@ -45,6 +45,14 @@ func (s *ServerImpl) PostAssetScans(ctx echo.Context) error {
 		return sendError(ctx, http.StatusBadRequest, fmt.Sprintf("failed to bind request: %v", err))
 	}
 
+	_, ok := assetScan.GetResourceCleanupStatus()
+	if !ok {
+		assetScan.ResourceCleanupStatus = models.NewResourceCleanupStatusWithDefaults()
+	}
```
Tehsmash, line 51: A pattern we've been following for the other objects is using the reconciler to set missing defaults instead of putting logic in the rest controllers, in this case the reconciler would detect ResourceCleanupStatus == nil, and then set it to Pending prior to processing anything else, the same logic can/should apply to any other defaults.
If somehow the ResourceCleanupStatus gets cleared, the reconciler should fix it.

If we're going to move this logic into the rest controllers then we needs to be consistent with handling all defaults in the rest controllers not just the states.

installation/docker/image_override.env
```diff
@@ -1,8 +1,7 @@
 # This is an example of how to override the default set of VMClarity container
 # images used by docker compose.
-APIServerContainerImage=docker.io/tehsmash/vmclarity-apiserver:swaggerui
-OrchestratorContainerImage=docker.io/tehsmash/vmclarity-orchestrator:swaggerui
-UIContainerImage=docker.io/tehsmash/vmclarity-ui:swaggerui
-UIBackendContainerImage=docker.io/tehsmash/vmclarity-ui-backend:swaggerui
-ScannerContainerImage=docker.io/tehsmash/vmclarity-cli:swaggerui
-YaraRuleServerContainerImage=docker.io/poke/yara-rule-server:test
```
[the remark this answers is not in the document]
adamtagscherer: my fault, I committed it unintentionally

pkg/apiserver/rest/asset_scan_controller.go
```diff
@@ -83,14 +91,26 @@ func (s *ServerImpl) PatchAssetScansAssetScanID(ctx echo.Context, assetScanID mo
...
+	if resourceCleanupStatus, ok := assetScan.GetResourceCleanupStatus(); ok {
+		existingResourceCleanupStatus, ok := existingAssetScan.GetResourceCleanupStatus()
+		if !ok {
+			return sendError(ctx, http.StatusInternalServerError, fmt.Sprintf("failed to retrieve ResourceCleanupStatus for existing asset scan. assetScanID=%v", existingAssetScan.Id))
+		}
+		err = existingResourceCleanupStatus.IsValidTransition(resourceCleanupStatus)
+		if err != nil {
+			return sendError(ctx, http.StatusBadRequest, err.Error())
+		}
+	}
```
Tehsmash, line 112: If we're moving to handling stuff in the rest controllers then shouldn't we handle the LastTransitionTime in the rest controller too? i.e. detect that a status change has been PATCH/PUTed and then update the time stamp to prevent the orchestrator or other API clients from a) having to manage it, b) potentially faking the data.

api/openapi.yaml
chrisgacsal: Yeah, that was our thinking here, to make sure that the whole `Status` object is updated when PATCHing. I am not sure if it make sense to update only one filed but not the others.

pkg/apiserver/rest/asset_scan_controller.go
chrisgacsal: We briefly touched this problem with @adamtagscherer. And the best would be to make the API server to be responsible for updating the `LastTransitionTime` field for `Status` objects. But we thought that it would be better to do the refactor for all the `Status` objects first and than have a separate PR for making the API server setting the `LastTransitionTime`.

pkg/apiserver/rest/asset_scan_controller.go
chrisgacsal: Probably the best would be to move the defaulting to `Orchestrator` and the `Scanner CLI` (CICD mode) for now and let the `API Server` to return a HTTP400 if the `ResourceCleanupStatus` field is missing from the request.

api/openapi.yaml
chrisgacsal: There is no timeout for resource cleanup. We might add a timeout to the reconcile logic at some point, but I do not think it is critical for now.

Tehsmash: 👍 cool, if we had it then TimedOut would be another reason to have here for failed. If we don't have a timeout then we're good.

pkg/apiserver/rest/asset_scan_controller.go
Tehsmash: I would prefer that it wasn't a 400 error, i.e. the object gets created with just the fields that the user specifies populated, and then the orchestrator backfills any missing information as part of the first steps in the reconcile loop. The object will simply remain unchanged if the orchestrator isn't running.

pkg/apiserver/rest/asset_scan_controller.go
Tehsmash: I think this makes sense, I was also thinking that we might like to add UpdatedAt/CreatedAt fields to all objects in the API.

pkg/apiserver/rest/asset_scan_controller.go
chrisgacsal: @Tehsmash What about the `CICD` mode? There is no `Orchestrator` involved in that case if I am not mistaken.

Tehsmash: thats a good point, in the CICD case if ResourceCleanupStatus is required (400 error), then the client should be POSTing something like `State: "Skipped", Reason:"NotRelevant"`.

Alternatively if we make it optional to we can leave it empty because its not relevant to that mode, and it'll never get populated because the orchestrator will never look at it.

api/models/resourcecleanupstatus.go
```diff
@@ -0,0 +1,98 @@
...
+		Message:            m,
+		LastTransitionTime: time.Now(),
+	}
+}
+
+func NewResourceCleanupStatusWithDefaults() *ResourceCleanupStatus {
+	return NewResourceCleanupStatus(ResourceCleanupStatusStatePending, ResourceCleanupStatusReasonAssetScanCreated, nil)
+}
+
+func (rs *ResourceCleanupStatus) Equals(r *ResourceCleanupStatus) bool {
```
akpsgit, line 45: Can you please add unit tests for the following functions?

pkg/apiserver/rest/asset_scan_controller.go
Tehsmash: @chrisgacsal wdyt?

chrisgacsal: > thats a good point, in the CICD case if ResourceCleanupStatus is required (400 error), then the client should be POSTing something like State: "Skipped", Reason:"NotRelevant".

Agreed. Probably I would use the `State: "Skipped", Reason:"NotApplicable"` instead. I vote for this approach.

> Alternatively if we make it optional to we can leave it empty because its not relevant to that mode, and it'll never get populated because the orchestrator will never look at it.

I prefer the former better.

@Tehsmash ^

Tehsmash: Sounds good, lets go with that.

pkg/apiserver/rest/asset_scan_controller.go
adamtagscherer: I created an issue for this: https://github.com/openclarity/vmclarity/issues/720

pkg/apiserver/rest/asset_scan_controller.go
adamtagscherer: I removed the defaulting from the AssetScan resource POST endpoint, and added to the CLI.

api/models/resourcecleanupstatus.go
adamtagscherer: I wouldn't do that for the following 2 reasons:
- we have now `e2e` tests in place which tests this function as well, and I'd feel an unit test for this would be redundant
- there are currently no unit tests written in the `api` package and I'm not sure if we want to introduce it now with all it's dependencies; I think it'd need a discussion whether we want to go down that path

api/models/resourcecleanupstatus.go
```diff
@@ -0,0 +1,98 @@
...
+			if transition == r.State {
+				return nil
+			}
+		}
+	}
+
+	return fmt.Errorf("invalid transition: from=%s to=%s", rs.State, r.State)
+}
+
+func (rs *ResourceCleanupStatus) isValidReasonTransition(r *ResourceCleanupStatus) error {
```
chrisgacsal, line 72:
```suggestion
func (rs *ResourceCleanupStatus) isValidReason(r *ResourceCleanupStatus) error {
```

The name for this method is misleading. There is no transition for `reason`. There are valid `reason`(s) for certain `state`(s).

pkg/apiserver/rest/asset_scan_controller.go
```diff
@@ -45,6 +45,14 @@ func (s *ServerImpl) PostAssetScans(ctx echo.Context) error {
 		return sendError(ctx, http.StatusBadRequest, fmt.Sprintf("failed to bind request: %v", err))
 	}
 
+	status, ok := assetScan.GetResourceCleanupStatus()
```
chrisgacsal, line 48:
```suggestion

	status, ok := assetScan.GetResourceCleanupStatus()
	switch {
	case !ok:
		return sendError(ctx, http.StatusBadRequest, "invalid request: resource cleanup status is missing")
	case status.State == models.ResourceCleanupStatusStatePending:
		fallthrough
	case status.State == models.ResourceCleanupStatusStateSkipped:
		break
	default:
		return sendError(ctx, http.StatusBadRequest, fmt.Sprintf("invalid request: initial state for resource cleanup status is invalid: %s", status.State))
	}
```

Alternative approach for this switch-case which allows to handle the `Pending` and `Skipped` cases differently. (Which is not necessary for now.)

api/models/resourcecleanupstatus.go
```diff
@@ -0,0 +1,98 @@
...
+		ResourceCleanupStatusStateDone,
+	},
+}
+
+var resourceCleanupStatusReasonMapping = map[ResourceCleanupStatusState][]ResourceCleanupStatusReason{
+	ResourceCleanupStatusStatePending: {
+		ResourceCleanupStatusReasonAssetScanCreated,
+	},
+	ResourceCleanupStatusStateSkipped: {
+		ResourceCleanupStatusReasonDeletePolicy,
```
chrisgacsal, line 21:
```suggestion
		ResourceCleanupStatusReasonDeletePolicy,
		ResourceCleanupStatusReasonNotApplicable,
```

chrisgacsal (changes requested):

api/models/resourcecleanupstatus.go
```diff
@@ -0,0 +1,98 @@
+package models
+
+import (
+	"fmt"
+	"time"
+)
+
+var resourceCleanupStatusStateTransitions = map[ResourceCleanupStatusState][]ResourceCleanupStatusState{
```
chrisgacsal, line 8:
```suggestion
var resourceCleanupStatusStateTransitions = map[ResourceCleanupStatusState][]ResourceCleanupStatusState{
	ResourceCleanupStatusStatePending: {
		ResourceCleanupStatusStateSkipped,
		ResourceCleanupStatusStateFailed,
		ResourceCleanupStatusStateDone,
	},
	ResourceCleanupStatusStateSkipped: {},
}
```

chrisgacsal (changes requested):

chrisgacsal (dismissed):

chrisgacsal (approved):

PR #696, opened 2023-09-15
feat(scanners): Add support to all existing scanners for containers

Tehsmash:
## Description

This commit adds support to all existing scanners for all families for
the image, docker-archive, oci-archive, and oci-directory input types.

In the scenario where a scanner does not support containers natively a
new library containerrootfs has been added which converts the container
into a rootfs folder.

The containerrootfs library supports a cached mode which indentifies
containers based on the source and cache the result so that multiple
scanners (even concurrent ones) attempting to convert the same container
to a rootfs only perform it once. All these cached folders are cleaned
up by a function called once all the families have completed.

## Type of Change

[ ] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


go.mod
```diff
@@ -485,11 +491,11 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-// github.com/openclarity/kubeclarity v0.0.0-20230503074245-0199b41a7b8f (v2.17.1)
+// github.com/openclarity/kubeclarity v0.0.0-20230914140301-d308a3187537 (v2.17.1)
```
FrimIdan, line 494: please also update `(v2.17.1)`

pkg/shared/families/malware/clam/clam.go
```diff
@@ -80,8 +83,23 @@ func (s *Scanner) Run(sourceType utils.SourceType, userInput string) error {
...
+			source := kubeclarityUtils.CreateSource(sourceType, userInput, false)
+			rootfs, err := containerRootfs.ToCachedTempDirectory(context.TODO(), source)
+			defer rootfs.Cleanup()
+			if err != nil {
+				s.sendResults(retResults, fmt.Errorf("failed to expand container to rootfs directory: %w", err))
+				return
+			}
+			rootFsPath = rootfs.Dir()
+		}
+
```
FrimIdan, line 100: looks like this code can be put in a function used by all families.

pkg/shared/families/malware/clam/clam.go
```diff
@@ -112,7 +130,7 @@ func (s *Scanner) Run(sourceType utils.SourceType, userInput string) error {
 }
 
 func New(c job_manager.IsConfig, logger *log.Entry, resultChan chan job_manager.Result) job_manager.Job {
-	conf := c.(*common.ScannersConfig) // nolint:forcetypeassert
+	conf := c.(common.ScannersConfig) // nolint:forcetypeassert
```
FrimIdan, line 133: Why the change here is needed? All other families are casting to a pointer.

pkg/shared/families/malware/config.go
```diff
@@ -25,5 +25,5 @@ type Config struct {
 	ScannersList    []string               `yaml:"scanners_list" mapstructure:"scanners_list"`
 	StripInputPaths bool                   `yaml:"strip_input_paths" mapstructure:"strip_input_paths"`
 	Inputs          []types.Input          `yaml:"inputs" mapstructure:"inputs"`
-	ScannersConfig  *common.ScannersConfig `yaml:"scanners_config" mapstructure:"scanners_config"`
+	ScannersConfig  common.ScannersConfig  `yaml:"scanners_config" mapstructure:"scanners_config"`
```
FrimIdan, line 28: again, why?

pkg/shared/families/rootkits/chkrootkit/chkrootkit.go
```diff
@@ -134,7 +134,7 @@ func (s *Scanner) isValidInputType(sourceType utils.SourceType) bool {
 	switch sourceType {
 	case utils.DIR, utils.ROOTFS:
 		return true
-	case utils.FILE, utils.IMAGE, utils.SBOM:
+	case utils.FILE, utils.IMAGE, utils.SBOM, utils.DOCKERARCHIVE, utils.OCIARCHIVE, utils.OCIDIR:
```
FrimIdan, line 137: we don't want to add the same support also here?

pkg/shared/utils/containerRootfs/fs.go
```diff
@@ -0,0 +1,166 @@
...
+			if numBytes >= perFileReadLimit || errors.Is(err, io.EOF) {
+				return fmt.Errorf("zip read limit hit (potential decompression bomb attack)")
+			}
+			if err != nil {
+				return fmt.Errorf("unable to copy file: %w", err)
+			}
+
+			err = output.Close()
+			if err != nil {
+				// TODO log the error
```
FrimIdan, line 60: why not to log it now?

pkg/shared/utils/containerRootfs/fs.go
```diff
@@ -0,0 +1,166 @@
...
+
+	tmpdir, err := os.MkdirTemp("", "")
+	if err != nil {
+		return tdrf, fmt.Errorf("unable to create temp directory: %w", err)
+	}
+	tdrf.dir = tmpdir
+
+	err = ToDirectory(ctx, src, tmpdir)
+	if err != nil {
+		return tdrf, fmt.Errorf("unable output squashed image to directory: %w", err)
```
FrimIdan, line 89: unable **to** output..?

pkg/shared/utils/containerRootfs/fs.go
```diff
@@ -0,0 +1,166 @@
...
+	var errs []error
+
+	if tdrf.dir != "" {
+		err := os.RemoveAll(tdrf.dir)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("unable to remove temp directory: %w", err))
+		}
+	}
+
+	return errors.Join(errs...)
```
FrimIdan, line 113: why we need `errs` and `Join` if we have only single error here?

pkg/shared/utils/containerRootfs/fs.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (tdrf *TempDirRootfs) Dir() string {
+	return tdrf.dir
+}
+
+func (tdrf *TempDirRootfs) Cleanup() error {
+	var errs []error
+
+	if tdrf.dir != "" {
```
FrimIdan, line 106: I prefer 
```
if tdrf.dir == "" {
  return nil
}

if err := os.RemoveAll....
```

pkg/shared/utils/containerRootfs/fs.go
```diff
@@ -0,0 +1,166 @@
...
+func (crfs *cachedRootfs) Dir() string {
+	return crfs.rootfs.Dir()
+}
+
+func (crfs *cachedRootfs) Cleanup() error {
+	return nil
+}
+
+func ToCachedTempDirectory(ctx context.Context, src string) (Rootfs, error) {
+	cacheMutex.Lock()
```
FrimIdan, line 134: why not to use `sync.Once`?

pkg/shared/utils/containerRootfs/fs.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	cachedEntry := &cachedRootfs{
+		complete: make(chan struct{}),
+		err:      nil,
+	}
+	cache[src] = cachedEntry
+	cacheMutex.Unlock()
+
+	rootfs, err := ToTempDirectory(ctx, src)
+	cachedEntry.err = err
```
FrimIdan, line 149: If one call to `ToCachedTempDirectory` failed for this `src` there is no way to try it again?

pkg/shared/utils/containerRootfs/fs.go
```diff
@@ -0,0 +1,166 @@
...
+
+func CleanupAllCached() error {
+	var errs []error
+	cacheMutex.Lock()
+	defer cacheMutex.Unlock()
+	for _, entry := range cache {
+		if entry.rootfs == nil {
+			continue
+		}
+		errs = append(errs, entry.rootfs.Cleanup())
```
FrimIdan, line 163: what about removing the entry from the cache?

pkg/shared/families/malware/clam/clam.go
Tehsmash: I need to root cause this some more, but I was always getting a nil pointer here which was causing a panic at L137 because it was trying to access the "Clam" field. I'll do some more digging

pkg/shared/families/rootkits/chkrootkit/chkrootkit.go
Tehsmash: missing this one, it should be added here, thanks for catching!

pkg/shared/utils/containerRootfs/fs.go
Tehsmash: Originally I had multiple things getting cleaned up here, but now its just the one thing. I'll refactor this.

pkg/shared/utils/containerRootfs/fs.go
Tehsmash: I'm not sure sync.Once will work here, when we have two callers asking to convert the same input concurrently, we need to be able to check if the conversion is already in progress for that particular src, and if it is we need to wait, if it is not then we need to start it.

Even if we used sync.Once we need to store a "completed" channel somewhere, and we'd also need to store the instance of sync.Once somewhere so we'd still need a lock for that.

pkg/shared/utils/containerRootfs/fs.go
Tehsmash: not right now as it keeps the in progress wait case and the complete wait case the same so this code is simpler. I can certainly update it so that if we receive a new call to ToCachedTempDirectory after it has completed with error that it tries again. WDYT?

pkg/shared/utils/containerRootfs/fs.go
Tehsmash: oops missed the delete here 👍

pkg/shared/utils/containerRootfs/fs.go
FrimIdan: `sync.Once` will make sure that function only called once, right?
So, if the function of conversion only run once and the path of the dir is kept on the struct (and maybe the error), other callers will just get the path (or the error) as a return.

something like this?
```
var (
	once sync.Once
)

type Rootfs interface {
	Dir() string
	Cleanup() error
}

type TempDirRootfs struct {
	dir string
}

type cachedTempDir struct {
	tdrf  *TempDirRootfs
	error error
}

func (c *cachedTempDir) ToCachedTempDirectory(ctx context.Context, src string) (Rootfs, error) {
	once.Do(func() {
		rootfs, err := ToTempDirectory(ctx, src)
		c.tdrf = rootfs
		c.error = err
	})
	return c.tdrf, c.error
}
```

FrimIdan: But I see that I missed the handling of the logic for different sources, so maybe `sync.Once` is not the case here.

Tehsmash: ToCachedTempDirectory can be called with multiple srcs i.e.

ToCachedTempDirectory(ctx, "nginx.tar")
ToCachedTempDirectory(ctx, "grype-server.tar")

In your implementation, only the first src will ever be unpacked. Also if we have two scanners calling it:

ToCachedTempDirectory(ctx, "nginx.tar")
ToCachedTempDirectory(ctx, "nginx.tar")

the second caller will get an partially unpacked directory because ToTempDirectory is still running.

FrimIdan: > the second caller will get an partially unpacked directory because ToTempDirectory is still running.

not sure why is that, but nm I think I get why sync.Once is not a good option here

Tehsmash: sync.Once() only guarantees that the passed func is called exactly once, following callers aren't blocked if that call is still running.

```
func Do() {
    once.Do(func() { time.Sleep(infinity) })
}

go func() {
    Do() <--- sleeps forever
}()

go func() {
   Do() <--- returns instantly
}()
```

FrimIdan: I'm not sure that is correct, based on https://pkg.go.dev/sync#Once.Do 
> Because no call to Do returns until the one call to f returns, if f causes Do to be called, it will deadlock.

https://go.dev/play/p/jqU5WQoyORm

Tehsmash: huh... well I've misread that part of the docs so many times, sync.Once is much more useful in that case!

FrimIdan: 😄 
maybe we can hold `once sync.Once` per `src` - something like 
```
type TempDirRootfs struct {
	dir string
    once sync.Once
}

var cache map[string]*TempDirRootfs
```

pkg/shared/families/malware/clam/clam.go
Tehsmash: I reverted this and it no longer gave me a nil pointer issue... I think this might be caused by not having the malware section in the families.yaml, we need to handle defaults a bit better.

pkg/shared/families/malware/config.go
Tehsmash: reverted

pkg/shared/utils/containerrootfs/totempdirectory.go
```diff
@@ -0,0 +1,168 @@
...
+	tmpdir, err := os.MkdirTemp("", "")
+	if err != nil {
+		return tdrf, fmt.Errorf("unable to create temp directory: %w", err)
+	}
+	tdrf.dir = tmpdir
+
+	err = ToDirectory(ctx, src, tmpdir)
+	if err != nil {
+		return tdrf, fmt.Errorf("unable to output squashed image to directory: %w", err)
+	}
```
FrimIdan, line 71: If we failed here, shouldn't we delete the tempdir?

pkg/shared/utils/containerrootfs/totempdirectory.go
```diff
@@ -0,0 +1,168 @@
...
+		entry.Wait()
+		return entry, entry.err
+	}
+	entry = newCachedRootfs()
+	cache.m[src] = entry
+	cache.mu.Unlock()
+
+	rootfs, err := toTempDirectory(ctx, src)
+	entry.Done(rootfs, err)
+	return entry, nil
```
FrimIdan, line 136: shouldn't you return also `entry.err` here?

pkg/shared/utils/containerrootfs/totempdirectory.go
```diff
@@ -0,0 +1,168 @@
...
+	if cache == nil {
+		return fmt.Errorf("uninitialized cache")
+	}
+	cache.mu.Lock()
+	defer cache.mu.Unlock()
+
+	errs := make([]error, len(cache.m), 0)
+	for key, entry := range cache.m {
+		if entry.rootfs == nil {
+			continue
```
FrimIdan, line 149: why you are not removing the key from the cache also here?

pkg/shared/utils/containerrootfs/todirectory.go
```diff
@@ -0,0 +1,98 @@
...
+			if numBytes >= perFileReadLimit || errors.Is(err, io.EOF) {
+				return fmt.Errorf("zip read limit hit (potential decompression bomb attack)")
+			}
+			if err != nil {
+				return fmt.Errorf("unable to copy file: %w", err)
+			}
+
+			err = output.Close()
+			if err != nil {
+				logger.WithError(err).Errorf("unable to close file %s", target)
```
FrimIdan, line 85: is it safe to continue here?

pkg/shared/utils/containerRootfs/fs.go
FrimIdan: instead of 
```
type cachedRootfs struct {
	complete chan struct{}
	err      error
	rootfs   Rootfs
}

type Cache struct {
	mu sync.Mutex
	m  map[string]*cachedRootfs
}
```
we can use
```
type cachedRootfs struct {
	once sync.Once
	err      error
	rootfs   Rootfs
}

type Cache struct {
	mu sync.Mutex
	m  map[string]*cachedRootfs
}
```

will it not simplified the complete/done/wait logic?

so you can getOrSet the cachedRootfs under mutex, and use the `once` to run only once the `toTempDirectory()`?

pkg/shared/utils/containerrootfs/totempdirectory.go
```diff
@@ -0,0 +1,168 @@
...
+func (cache *Cache) ToTempDirectory(ctx context.Context, src string) (Rootfs, error) {
+	if cache == nil {
+		return nil, fmt.Errorf("uninitialized cache")
+	}
+
+	cache.mu.Lock()
+	entry, ok := cache.m[src]
+	if ok {
+		cache.mu.Unlock()
+		entry.Wait()
```
fishkerez, line 127: Do we want to wait here on the ctx as well?
Maybe pass ctx to Wait and do:
```
select {
	case <-ctx.Done():
	case <-crfs.complete:
		
	}
```

pkg/shared/utils/containerrootfs/totempdirectory.go
Tehsmash: In the current implementation its expected that even if an error is returned defer rootfs.Cleanup() is expected to be called like:

```
rootfs, err := toTempDirectory(...)
defer rootfs.Cleanup()
if err != nil {
   ...
}
```

This does feel a bit weird though...

FrimIdan: since it's internal function I guess we can do whatever we feel like, but to me it feels like - if an `err` returned from a function, the function is responsible for cleaning and `roots` will be `nil`

pkg/shared/utils/containerrootfs/todirectory.go
Tehsmash: Opinions on the internet are mixed on this, I think in this case we probably should fail because we're intending to read it again before this problem exits, I'll update.

pkg/shared/utils/containerRootfs/fs.go
Tehsmash: I'm not sure that it will now, especially given Erez's comment about make sure that we abort Wait() if the context is canceled. I've updated the Wait and Done functions to make this nicer.

pkg/shared/utils/containerrootfs/totempdirectory.go
Tehsmash: I found a pattern online for handling defer's within a function only if the function is not successful, for example:
```
func setupThing() (func() error, error) {
   successful := false
   
   file, err := os.OpenFile(...)
   if err != nil {
       return func() error {}, err
   }
   defer func() {
      if !successful {
          if err := file.Close(); err != nil {
              log.Errorf(err)
          }
      }
   }()
   
   file2, err := os.OpenFile(...)
   if err != nil {
        return func() error {}, err
   }
   defer func() {
      if !successful {
          if err := file2.Close(); err != nil {
              log.Errorf(err)
          }
      }
   }()
   
   successful = true
   return func() error {
          if err := file.Close(); err != nil {
              return err
          }
          if err := file2.Close(); err != nil {
              return err
          }
   }
}
```

If the function returns before reaching `successful = true` then the defer's in this function trigger and clean up the files and any error's are logged, otherwise its the callers responsibility to call the returned cleanup function.

pkg/shared/utils/containerrootfs/todirectory.go
Tehsmash: switched to using a defer for this and logging it, if we run into issues then we can update it later to using the "!successful" defer kind of pattern as I defined above.

akpsgit (approved):

PR #705, opened 2023-09-21
fix(make): remove deprecated clean-backend and run go mod tidy in all modules

paralta:
## Description

* Remove deprecated `clean-backend` and add `clean-go` to remove all GO executables
* Change `gomod-tidy` to run in all directories with `go.mod` files

## Type of Change

[x] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,9 +205,14 @@ license-cache: bin/licensei ## Generate license cache
 .PHONY: check
 check: lint test helm-lint ## Run tests and linters
 
+TIDYGOMODULES = $(addprefix tidy-, $(GOMODULES))
+
+.PHONY: $(TIDYGOMODULES)
+$(TIDYGOMODULES):
+	cd $(dir $(@:tidy-%=%)) && go mod tidy
+
 .PHONY: gomod-tidy
-gomod-tidy:
-	go mod tidy
+gomod-tidy: $(TIDYGOMODULES)
 
 .PHONY: gen-api
 gen-api: gen-apiserver-api gen-uibackend-api ## Generating API code
--- a/Makefile
+++ b/Makefile
@@ -140,8 +140,12 @@ e2e: docker-apiserver docker-cli docker-orchestrator docker-ui docker-ui-backend
 clean-ui:
 	@(rm -rf ui/build ; echo "UI cleanup done" )
 
+.PHONY: clean-go
+clean-go:
+	@(rm -rf bin ; echo "GO executables cleanup done" )
+
 .PHONY: clean
-clean: clean-ui clean-backend ## Clean all build artifacts
+clean: clean-ui clean-go ## Clean all build artifacts
 
 $(BIN_DIR):
 	@mkdir -p $(BIN_DIR)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -140,8 +140,12 @@ e2e: docker-apiserver docker-cli docker-orchestrator docker-ui docker-ui-backend
 clean-ui:
 	@(rm -rf ui/build ; echo "UI cleanup done" )
 
+.PHONY: clean-go
+clean-go:
+	@(rm -rf bin ; echo "GO executables cleanup done" )
```
Tehsmash, line 145: might need to be more specific here, if we `rm -rf bin` we'll delete golangci-lint and license-check etc too which I'm not sure is what we want to happen.

Tehsmash (changes requested):

Makefile
paralta: Thanks for reviewing, @Tehsmash!
I think it would be good to clean every executable in the bin folder with `make clean` not only the ones we build, since we use curl to get those executables when we run `make check` etc. What do think?

Makefile
```diff
@@ -201,9 +205,14 @@ license-cache: bin/licensei ## Generate license cache
 .PHONY: check
 check: lint test helm-lint ## Run tests and linters
 
+TIDYGOMODULES = $(addprefix tidy-, $(GOMODULES))
+
+.PHONY: $(TIDYGOMODULES)
+$(TIDYGOMODULES):
+	cd $(dir $(@:tidy-%=%)) && go mod tidy
```
chrisgacsal, line 212:
```suggestion
	cd $(dir $(@:tidy-%=%)) && go mod tidy -go=1.20
```

chrisgacsal (dismissed):

Makefile
chrisgacsal: What would be even better if the GO version for `go mod tidy` would be passed as a variable and determined based on the `go ...` directive in the `go.mod` files.

Tehsmash: > If the -go flag is set, go mod tidy will update the [go directive](https://go.dev/ref/mod#go-mod-file-go) to the indicated version
 
It might be better to have this controlled in the Makefile, then it can force all our modules to use the same go version instead of maintaining it per go.mod

chrisgacsal: One way to get the go version from go.mod:

Makefile
Tehsmash: +1 I think that would be a good idea, have like clean-golangci-lint + clean-licensei

Makefile
chrisgacsal: > > If the -go flag is set, go mod tidy will update the [go directive](https://go.dev/ref/mod#go-mod-file-go) to the indicated version
> 
> It might be better to have this controlled in the Makefile, then it can force all our modules to use the same go version instead of maintaining it per go.mod

Agreed.

paralta: thanks for reviewing again 😄 please check the last commit and let me know what you think

chrisgacsal (approved):
Nice! :shipit:

Tehsmash (approved):

Issue #564, opened 2023-08-16
Kubernetes Container Image Discovery


Tehsmash:
Following design #527 we should implement container image discovery as part of the kubernetes provider added by #561.

The provider should list all the nodes in the cluster, and from the node's status field collect the list of containers which are currently stored in the node's container runtime. A ContainerImage asset should be returned for each discovered container.

We should use the reference/name with registry hash as the ID for the ContainerImage asset.
We should use the container image size as the ContainerImage asset size.
We should use the Node's architecture and OS as the ContainerImage architecture and os.

PR #711, opened 2023-09-25
feat(kubernetes): Discover container images and running containers

### LINKED ISSUE #562: Kubernetes Running Container Discovery
Following design #527, we should implement container discovery in the DiscoverAssets function added by #561.

The container discovery logic should fetch all pods from the cluster, scan for containerStatus' and return a new asset entry for each container which is not terminated:
- containerID this will be used as the "ID" field for the container asset.
- imageID this will be used to fill out the image part of the container asset using the digest not the tag.
- name this will be used to fill out the name part of the container asset

Tehsmash:
## Description

With this addition the kubernetes provider can now discover container images and running containers from the connected kubernetes cluster.

Closes #562 #564

## Type of Change

[ ] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


pkg/containerruntimediscovery/dockerdiscoverer.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (dd *DockerDiscoverer) Images(ctx context.Context) ([]models.ContainerImageInfo, error) {
+	// List all docker images
+	images, err := dd.client.ImageList(ctx, dtypes.ImageListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list images: %w", err)
+	}
+
+	// Convert images to container image info
+	result := make([]models.ContainerImageInfo, len(images))
```
fishkerez, line 62: Why do you not use the merge logic as you did in containerd discoverer?

pkg/containerruntimediscovery/containerddiscoverer.go
```diff
@@ -0,0 +1,191 @@
...
+
+	// NOTE(sambetts) We can not use image.Size as it gives us the size of
+	// the compressed layers and not the real size of the content.
+	snapshotter := cd.client.SnapshotService(containerd.DefaultSnapshotter)
+	size, err := imgutil.UnpackedImageSize(ctx, snapshotter, image)
+	if err != nil {
+		return models.ContainerImageInfo{}, fmt.Errorf("unable to determine size for image: %w", err)
+	}
+
+	repoTags, repoDigests := ParseImageReferences([]string{image.Name()})
```
fishkerez, line 106: why can't we take the tags and digests from the manifest/config? How does the name of the image helps us here? We will get in the most one tag/digest

pkg/containerruntimediscovery/containerddiscoverer.go
```diff
@@ -0,0 +1,191 @@
...
+		return models.ContainerInfo{}, fmt.Errorf("unable to get info for container %s: %w", id, err)
+	}
+	createdAt := info.CreatedAt
+
+	image, err := container.Image(ctx)
+	if err != nil {
+		return models.ContainerInfo{}, fmt.Errorf("unable to get image from container %s: %w", id, err)
+	}
+
+	imageInfo, err := cd.getContainerImageInfo(ctx, image)
```
fishkerez, line 178: It make more sense to me here to look for the image by id in the previously found images, rather than recalculate the data. WDYT?

pkg/orchestrator/provider/kubernetes/config.go
```diff
@@ -29,6 +29,10 @@ type Config struct {
 	// KubeConfig defines a path to a kubeconfig file to use to connect to
 	// the Kubernetes API
 	KubeConfig string `mapstructure:"kubeconfig"`
+
+	// Container Runtime Discovery Namespace is the namespace where the
+	// container runtime discovery daemonset is running.
+	ContainerRuntimeDiscoveryNamespace string `mapstructure:"container_runtime_discovery_namespace"`
```
fishkerez, line 35: This never being set

pkg/containerruntimediscovery/containerddiscoverer.go
```diff
@@ -0,0 +1,191 @@
...
+			digests = append(digests, parsed.String())
+		} else if _, ok := parsed.(reference.Tagged); ok {
+			tags = append(tags, parsed.String())
+		}
+	}
+	return tags, digests
+}
+
+func (cd *ContainerdDiscoverer) Containers(ctx context.Context) ([]models.ContainerInfo, error) {
+	containers, err := cd.client.Containers(ctx)
```
fishkerez, line 139: I thought we talked about taking the containers info from the pods status

pkg/orchestrator/provider/kubernetes/provider.go
```diff
@@ -17,14 +17,24 @@ package kubernetes
 
 import (
 	"context"
+	"errors"
```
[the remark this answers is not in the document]
Tehsmash: I actually think it should be the other way around, we no longer have anything called a client other than the internal clients for each provider so for a newcomer that doesn't know the history client.go is weird.

pkg/containerruntimediscovery/discoverer.go
```diff
@@ -0,0 +1,55 @@
...
+
+type Discoverer interface {
+	Images(ctx context.Context) ([]models.ContainerImageInfo, error)
+	Containers(ctx context.Context) ([]models.ContainerInfo, error)
+}
+
+type DiscovererFactory func(ctx context.Context) (Discoverer, error)
+
+var discovererFactories map[string]DiscovererFactory = map[string]DiscovererFactory{
+	"docker":     NewDockerDiscoverer,
```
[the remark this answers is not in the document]
Tehsmash: they can be added later as / when people need them. Its pluggable so that its easy to add them without changing the infrastructure.

cmd/vmclarity-cr-discovery-server/cmd/root.go
```diff
@@ -0,0 +1,96 @@
...
+				return fmt.Errorf("unable to create discoverer: %w", err)
+			}
+
+			abortCtx, cancel := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
+			defer cancel()
+
+			// TODO(sambetts) Get listen address from the viper
+			// configuration with a default
+			listenAddr := ":8080"
+			ids := containerruntimediscovery.NewContainerRuntimeDiscoveryServer(logger, listenAddr, discoverer)
```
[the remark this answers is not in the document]
Tehsmash: It used to be "Image Discovery Server", I'll change this to "CRDS"

charts/vmclarity/templates/cr-discovery-server/daemonset.yaml
```diff
@@ -0,0 +1,49 @@
...
+          resources: {{- toYaml .Values.crDiscoveryServer.resources | nindent 12 }}
+          {{- end }}
+          volumeMounts:
+            - mountPath: /var/run/containerd
+              name: containerd
+              readOnly: true
+            - mountPath: /var/lib/docker
+              name: docker
+              readOnly: true
+      volumes:
```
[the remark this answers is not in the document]
Tehsmash: 1. not true, by default hostPaths are optional mounts and if they are missing kubernetes ignores it.
2. this is a pattern used by other well known applications ([cadvisor](https://github.com/google/cadvisor) for example) which allows our code to detect which container runtime is enabled so users don't need to input it into the chart and allows for mixed CR environments with a single daemonset. The security concerns can be addressed by a) mounting as ReadOnly (already done) and by controlling the container security context.

pkg/orchestrator/provider/kubernetes/config.go
Tehsmash: Missed the BindEnv in the NewConfig function I'll add this

charts/vmclarity/templates/orchestrator/deployment.yaml
```diff
@@ -152,6 +152,8 @@ spec:
             {{- if eq .Values.orchestrator.provider "kubernetes" }}
             - name: PROVIDER
               value: "kubernetes"
+            - name: VMCLARITY_KUBERNETES_CONTAINER_RUNTIME_DISCOVERY_NAMESPACE
```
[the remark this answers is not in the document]
Tehsmash: kubernetes/config.go

cmd/vmclarity-cr-discovery-server/cmd/root.go
```diff
@@ -0,0 +1,96 @@
...
+			ids.Serve()
```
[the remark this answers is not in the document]
Tehsmash: golang http server doesn't support context canceling and for graceful shutdown we need to call Shutdown(). I don't necessarily want to make serve blocking either and I think it'll make it less useful for reuse in other places. I could add a blocking version "ServeAndBlock" which just moves the`<- abortCtx.Done()` inside Serve, but we'd still need to call shutdown outside of that logic.

You never need to wait on the parent context if you are waiting on its children, if a parent context is canceled it cascades to any and all children so in this case waiting on on ctx and abortCtx is redundant.

pkg/containerruntimediscovery/containerddiscoverer.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/openclarity/vmclarity/pkg/shared/utils"
+)
+
+type ContainerdDiscoverer struct {
+	client *containerd.Client
+	logger *logrus.Entry
+}
+
+func NewContainerdDiscoverer(ctx context.Context) (Discoverer, error) {
+	client, err := containerd.New("/var/run/containerd/containerd.sock", containerd.WithDefaultNamespace("k8s.io"))
```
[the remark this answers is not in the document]
Tehsmash: I've added a comment explaining it and also switched to using the constant from containerd which defines it

pkg/containerruntimediscovery/dockerdiscoverer.go
Tehsmash: Docker natively groups images by ImageID, containerd doesn't and stores a separate entry per manifest ID so in ContainerD we have to merge the entries together by ImageID outselves.

pkg/containerruntimediscovery/containerddiscoverer.go
Tehsmash: ContainerD images are not grouped by ImageID like in Docker, they are stored like:
```
[
   {
     "name": "docker.io/library/nginx:latest",
     ... 
   },
   {
     "name": "ghcr.io/library/nginx:latest",
     ... 
   }
]
```
and each unique tag or manifest has a separate entry.

We parse the name into tag or digest, and return it as a list so then its easier to merge with the other entries that we find.

pkg/containerruntimediscovery/dockerdiscoverer.go
```diff
@@ -0,0 +1,147 @@
...
+		Labels:       convertTags(image.Config.Labels),
+		RepoTags:     &image.RepoTags,
+		RepoDigests:  &image.RepoDigests,
+		ObjectType:   "ContainerImageInfo",
+		Os:           utils.PointerTo(image.Os),
+		Size:         utils.PointerTo(int(image.Size)),
+	}, nil
+}
+
+func convertTags(tags map[string]string) *[]models.Tag {
```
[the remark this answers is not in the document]
Tehsmash: +1 there is a lot of shared code between here and the docker provider we may need to refactor if we determine that they will always be the same

pkg/containerruntimediscovery/containerddiscoverer.go
Tehsmash: yeah, I mentioned this to @chrisgacsal I think we need to change the API model so that we just store the ID and then we can look it up later. However it would mean that if we find a container and we are missing the ContainerImageInfo asset that we'll be missing a lot of information.

pkg/containerruntimediscovery/containerddiscoverer.go
Tehsmash: PodStatus's don't give us all the information that we need without a lot of additional work, and actually not all the running containers are in the pod statuses either (for example the pause/sandbox container which makes up a pod is missing) therefore discovery through the container runtime is more accurate and efficient IMO

cmd/vmclarity-cr-discovery-server/cmd/root.go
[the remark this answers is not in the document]
fishkerez: I meant you can call ctx.Done() in case that ListenAndServe returns error. but never mind, didn't notice that you call Fatal in that case.

pkg/containerruntimediscovery/containerddiscoverer.go
fishkerez: I meant look in the imageIDs that you found as part of this discovery session. 
```
images := ListImages()
containers := ListContainers(images)
```

and then in ListContainers, we can match imageId to image data in images.

Tehsmash: that will require holding every image that we found during the discovery run in memory in the orchestrator as there is no access to the API server as part of the provider side of the discovery code. If we change the API model so that the container's just store a reference / imageID, then we can make the link in API / UI layer instead of reimplementing it in each provider.

pkg/orchestrator/provider/kubernetes/provider.go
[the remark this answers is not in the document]
fishkerez: @Tehsmash I have no problem to call it provider if you think it is better (or other name), I just want the names to be alligned across providers.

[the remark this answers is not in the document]
chrisgacsal: @fishkerez I agree with you on having the same naming convention applied across the code base, however I agree with @Tehsmash that the name `provider` is more suitable in this case as it implements the `provider.Provider` interface and we should probably update the rest of the providers to adopt this naming.

e2e/go.mod
```diff
@@ -193,6 +193,17 @@ require (
 replace (
 	github.com/openclarity/vmclarity => ../
 	github.com/openclarity/vmclarity/api => ../api
+	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc => go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.40.0
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp => go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.40.0
+
+	// These are replaced in order to be compatible with docker compose v2.22.0
```
chrisgacsal, line 199: I would group replaces by component instead of smashing everything together. Could you please move the docker-compose related replaces back to a separate group which we already have.
Also it would be really helpful to add a bit of context in comments why we need these replaces and which dependencies requiring them. It will make the job of the next person looking into this issue way easier when they need to figure out if the replaces are still valid or need to be updated, etc.

e2e/go.mod
```diff
@@ -206,9 +217,15 @@ replace (
...
-	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp => go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.41.0
-	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp => go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.15.1
+
+	// NOTE(adamtagscherer):
+	// For k8s dependencies, we use a replace directive, to prevent them being
+	// upgraded to the version specified in containerd, which is not relevant to the
+	// version needed.
+	// See https://github.com/docker/buildx/pull/948 for details.
+	// https://github.com/docker/buildx/blob/v0.9.1/go.mod#L62-L64
+	k8s.io/api => k8s.io/api v0.26.2
```
chrisgacsal, line 227: Are the k8s replaces needed because we are pulling in `k8s.io/api` in the root `go.mod` and docker-compose is not happy with that version?

pkg/containerruntimediscovery/containerddiscoverer.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/openclarity/vmclarity/api/models"
+	"github.com/openclarity/vmclarity/pkg/shared/log"
+	"github.com/openclarity/vmclarity/pkg/shared/utils"
+)
+
+type ContainerdDiscoverer struct {
+	client *containerd.Client
+	logger *logrus.Entry
```
chrisgacsal, line 36: I personally dislike the approach of storing the logger in any struct instead of retrieving the logger from context mainly because there is not way to provide request specific context to be logged if the initial logger is reused.

I think this is a bad practice so I'd prefer to remove Logger from struct and get it from context.Context whereever it is used.

chrisgacsal (changes requested):

pkg/orchestrator/provider/kubernetes/provider.go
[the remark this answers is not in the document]
Tehsmash: I think I'll open a ticket for this, I don't think this is the PR for this change to be be made.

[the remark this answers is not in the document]
Tehsmash: https://github.com/openclarity/vmclarity/issues/834

pkg/containerruntimediscovery/containerddiscoverer.go
```diff
@@ -0,0 +1,191 @@
...
+
+	result := []models.ContainerImageInfo{}
+	for _, image := range imageSet {
+		result = append(result, image)
+	}
+	return result, nil
+}
+
+func (cd *ContainerdDiscoverer) getContainerImageInfo(ctx context.Context, image containerd.Image) (models.ContainerImageInfo, error) {
+	configDescriptor, err := image.Config(ctx)
```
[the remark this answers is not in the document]
Tehsmash: not sure I understand, here are the docs for the containerd.Image object for reference: https://pkg.go.dev/github.com/containerd/containerd#Image

e2e/go.mod
Tehsmash: these were readded as part of a merge conflict, fixed in the latest

pkg/containerruntimediscovery/containerddiscoverer.go
Tehsmash: I think I caught all the new places in the code doing this and added the context where it was missing, please check.

Makefile
```diff
@@ -203,8 +206,16 @@ docker-ui-backend: ## Build UI Backend container image
...
 		-t $(DOCKER_IMAGE)-ui-backend:$(DOCKER_TAG) .
 
+.PHONY: docker-cr-discovery-server
+docker-cr-discovery-server: ## Build K8S Image Resolver Docker image
+	@(echo "Building cr-discovery-server docker image ..." )
+	docker build --file ./Dockerfile.cr-discovery-server --build-arg VERSION=${VERSION} \
+		--build-arg BUILD_TIMESTAMP=$(shell date -u +"%Y-%m-%dT%H:%M:%SZ") \
+		--build-arg COMMIT_HASH=$(shell git rev-parse HEAD) \
+		-t ${DOCKER_IMAGE}-cr-discovery-server:${DOCKER_TAG} .
+
```
chrisgacsal, line 216:
```suggestion
	$(info Building cr-discovery-server docker image ...)
	docker build --file ./Dockerfile.cr-discovery-server --build-arg VERSION=$(VERSION) \
		--build-arg BUILD_TIMESTAMP=$(BUILD_TIMESTAMP) \
		--build-arg COMMIT_HASH=$(COMMIT_HASH) \
		-t $(DOCKER_IMAGE)-cr-discovery-server:$(DOCKER_TAG) .

```

Dockerfile.cr-discovery-server
```diff
@@ -0,0 +1,40 @@
...
+
+RUN xx-verify bin/vmclarity-cr-discovery-server
+
+FROM alpine:3.18
+
+WORKDIR /app
+
+COPY --from=builder ["/build/bin/vmclarity-cr-discovery-server", "/app/vmclarity-cr-discovery-server"]
+
+ENTRYPOINT ["/app/vmclarity-cr-discovery-server"]
```
chrisgacsal, line 40:
```suggestion
COPY --from=builder ["/build/bin/vmclarity-cr-discovery-server", "/bin/vmclarity-cr-discovery-server"]

ENTRYPOINT ["/bin/vmclarity-cr-discovery-server"]
```

installation/kubernetes/helm/vmclarity/README.md
```diff
@@ -51,6 +51,26 @@ secrets.
...
+| crDiscoveryServer.containerSecurityContext.readOnlyRootFilesystem | bool | `true` | Mounts the container file system as ReadOnly |
+| crDiscoveryServer.containerSecurityContext.runAsGroup | int | `1001` | Group ID which the containers should run as |
+| crDiscoveryServer.containerSecurityContext.runAsNonRoot | bool | `true` | Whether the containers should run as a non-root user |
+| crDiscoveryServer.containerSecurityContext.runAsUser | int | `1001` | User ID which the containers should run as |
+| crDiscoveryServer.image.digest | string | `""` | Container Runtime Discovery Server image digest. If set will override the tag. |
+| crDiscoveryServer.image.pullPolicy | string | `"IfNotPresent"` | Container Runtime Discovery Server image pull policy |
+| crDiscoveryServer.image.registry | string | `"docker.io"` | Container Runtime Discovery Server container registry |
+| crDiscoveryServer.image.repository | string | `"tehsmash/vmclarity-cr-discovery-server"` | Container Runtime Discovery Server container repository |
+| crDiscoveryServer.image.tag | string | `"k8simagediscovery"` | Container Runtime Discovery Server container tag |
```
chrisgacsal, line 66: Docs needs to be regenerated.

installation/kubernetes/helm/vmclarity/values.yaml
```diff
@@ -819,3 +819,58 @@ yaraRuleServer:
...
         - ALL
+
+crDiscoveryServer:
+  image:
+    # -- Container Runtime Discovery Server container registry
+    registry: docker.io
+    # -- Container Runtime Discovery Server container repository
+    repository: tehsmash/vmclarity-cr-discovery-server
+    # -- Container Runtime Discovery Server container tag
+    tag: k8simagediscovery
```
chrisgacsal, line 830:
```suggestion
    registry: ghcr.io
    # -- Container Runtime Discovery Server container repository
    repository: openclarity/vmclarity-cr-discovery-server
    # -- Container Runtime Discovery Server container tag
    tag: latest
```

chrisgacsal (commented):
@Tehsmash It looks awesome! I have left a couple comments related to minor things.

chrisgacsal (approved):

chrisgacsal (approved):

Issue #637, opened 2023-09-06
Create new API objects


adamtagscherer:
API objects, introduction the new object for the provider, and create relations

1. create a new provider object
2. modify the asset object by adding a new field provider which tells which provider has discovered it (list of ids)
3. add a new field called - provider - to the asset scan object itself, asset scan scheduler will fill this field out

Issue #353, opened 2023-05-31
Multi-cloud multi account support


akpsgit:

lgecse:
@akpsgit @Tehsmash please provide the detailed description and architecture plan for this work piece.

ramizpolic:
To clarify: decoupling of compnents (just running discovery + scanner in environment, and have a central VMClarity instance without the discovery and scanner that collects results from different envs).

PR #714, opened 2023-09-26
feat(api): add new provider api object

paralta:
## Description

Part 1/3 of issue https://github.com/openclarity/vmclarity/issues/637. Please find more feature design details in discussion https://github.com/openclarity/vmclarity/discussions/581. 

* Add provider object to API
* Add providers path to API
* Add provider table to Database
* Add provider controller to API server
* Add provider methods to Backend client

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


fishkerez:
I think the PR and the issue lacks the motivation behind adding the new Provider object. Can we elaborate on this a bit more?

Tehsmash:
> I think the PR and the issue lacks the motivation behind adding the new Provider object. Can we elaborate on this a bit more?

The issue is part of an epic it is linked at the top of the issue, https://github.com/openclarity/vmclarity/issues/353 the design document that was completed gives the full motivation for the new provider object.

api/openapi.yaml
```diff
@@ -2350,6 +2554,107 @@ components:
...
+          properties:
+            state:
+              description: Describes the health state of the provider.
+              type: string
+              enum:
+                - Healthy
+                - NotHealthy
+                - Unknown
+
+    ProviderAnnotations:
```
Tehsmash, line 2649: Maybe just "Annotations", then we can reuse this schema to add annotations to other API objects in the future.

pkg/apiserver/database/gorm/provider.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	return nil
+}
+
+func (t *ProvidersTableHandler) checkUniqueness(provider models.Provider) (*models.Provider, error) {
+	var providers []Provider
+	// In the case of creating or updating an provider, needs to be checked whether other provider exists with same properties.
+	// TODO(paralta) Confirm uniqueness criteria
+	filter := fmt.Sprintf("id ne '%s' and displayName eq '%s'", *provider.Id, *provider.DisplayName)
```
Tehsmash, line 260: not sure we need to enforce displayName to be unique. I expected displayName to be an optional field and this criteria will fail if multiple Providers have empty DisplayName.

This code will also panic with a nil pointer exception if DisplayName is not set in the POST request so we need to either enforce DisplayName is set or handle == nil.

Tehsmash (changes requested):

pkg/apiserver/database/gorm/provider.go
paralta: good point on the second part!

regarding the first part, any thoughts/alternatives on checking provider uniqueness or should we just use provider id?

Tehsmash: The provider ID UUID is the only field that needs to be unique and that isn't currently externally controllable. Although that might change depending on the registration process so we can probably skip the uniqueness checks for now, wdyt? @chrisgacsal maybe you have some thoughts here?

pkg/apiserver/database/gorm/odata.go
```diff
@@ -1106,6 +1106,24 @@ var schemaMetas = map[string]odatasql.SchemaMeta{
...
 		},
 	},
+	providerSchemaName: {
+		Table: "assets",
+		Fields: odatasql.Schema{
+			"id":                     odatasql.FieldMeta{FieldType: odatasql.StringFieldType},
+			"revision":               odatasql.FieldMeta{FieldType: odatasql.NumberFieldType},
+			"displayName":            odatasql.FieldMeta{FieldType: odatasql.StringFieldType},
+			"lastHeartbeatTime":      odatasql.FieldMeta{FieldType: odatasql.DateTimeFieldType},
+			"status":                 odatasql.FieldMeta{FieldType: odatasql.StringFieldType},
```
akpsgit, line 1116: shouldn't be ProviderStatus object?

pkg/apiserver/database/gorm/odata.go
```diff
@@ -1106,6 +1106,24 @@ var schemaMetas = map[string]odatasql.SchemaMeta{
 			"state":              odatasql.FieldMeta{FieldType: odatasql.StringFieldType},
 		},
 	},
+	providerSchemaName: {
+		Table: "assets",
```
Tehsmash, line 1110: This should be providers, and there will need to be changes in database.go to add the new table as well as add indexes for the ID field, and any other commonly queried fields.

Tehsmash (changes requested):

api/openapi.yaml
```diff
@@ -2410,6 +2600,112 @@ components:
...
+        - lastTransitionTime
+      properties:
+        state:
+          description: Describes the health state of the provider.
+          type: string
+          enum:
+            - Healthy
+            - Unhealthy
+            - Unknown
+        reason:
```
adamtagscherer, line 2698: I think the reason field of the status object should be an enum just like for ResourceCleanupStatus: https://github.com/openclarity/vmclarity/blob/main/api/openapi.yaml#L2090

chrisgacsal: Also documentation is missing for `state` field which describes the different states.

paralta:
thank you so much for all the feedback 😄 I just updated the PR and added a small test, please have a look

chrisgacsal:
@paralta Could you pls rebase this PR to the latest main?

Tehsmash (approved):

chrisgacsal (approved):
LGTM! :shipit:

Issue #679, opened 2023-09-12
Helper function for run commands and parse their output.


pbalogh-sa:
## Problem Statement

I think we need to update command output parsing in order to use an io.Reader and a bufio.Scanner to process the output line by line to avoid huge memory consumption in the case of Yara.

## Proposed Solution

It would be nice if there was a helper function that has two args (command, parser_func) that runs the command and parse output line-by-line.

PR #716, opened 2023-09-27
chore: run command and parse output line by line

pbalogh-sa:
## Description

Instead of running a command and returning byte slice output, this PR addresses parse command output line-by-line during the execution time.

closes: #679 

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


pkg/shared/utils/utils.go
```diff
@@ -53,6 +58,39 @@ func RunCommand(cmd *exec.Cmd) ([]byte, error) {
...
+	}
+
+	scanner := bufio.NewScanner(r)
+	// Use the scanner to scan the output line by line and parse it
+	done := make(chan struct{})
+	go func() {
+		// Read line by line and process it
+		for scanner.Scan() {
+			line := scanner.Text()
+			resChan <- pfn(line, logger)
```
Tehsmash, line 75: I wouldn't add a channel here, the caller can handle their own result/error handling process in the closure they parse into the RunCommandAndParseOutputLineByLine function. For example in the Yara code:

```
detectedMalware := []common.DetectedMalware{}
parseErrs = []error{}
processFunc := func(ctx context.Context, line string) {
	malware, err := util.ParseYaraScanOutput(ctx, line)
    if err != nil {
    	errs = append(errs, err)
    	return
    }
    if malware != nil {
    	detectedMalware = append(detectedMalware, malware)
    }
}
err := RunCommandAndParseOutputLineByLine(cmd, processFunc)
if err != nil {
    return nil, fmt.Errorf("error running command: %w", err)
}
err := errors.Join(parseErrs...)
if err != nil {
   return nil, fmt.Errorf("error parsing lines: %w", err)
}
```

pkg/shared/families/malware/yara/yara.go
```diff
@@ -79,18 +78,21 @@ func (s *Scanner) Run(sourceType utils.SourceType, userInput string) error {
...
-		detectedMalware := util.ParseYaraScanOutput(string(out), s.logger)
-		retResults.Malware = detectedMalware
-		// In the case of Yara there is no scan summary.
-		retResults.Summary = &common.ScanSummary{}
-
-		s.sendResults(retResults, nil)
+		resChan := make(chan any, resultBufferSize)
+		ctx, cancel := context.WithCancel(context.Background())
+		go func() {
+			err := sharedutils.RunCommandAndParseOutputLineByLine(yaraCommand, util.ParseYaraScanOutput, resChan, s.logger)
```
Tehsmash, line 84: I think we should switch to using context to pass around the logger. For example:

```
ctx := context.TODO()
ctx = log.SetLoggerForContext(ctx, s.logger)
```

then in the code that needs the logger, as long as they have the context they can simply do:

```
logger := GetLoggerFromContextOrDiscard(ctx)
```

pkg/shared/utils/utils.go
```diff
@@ -53,6 +58,39 @@ func RunCommand(cmd *exec.Cmd) ([]byte, error) {
...
+		}
+		done <- struct{}{}
```
Tehsmash, line 77: Probably better to `close(done)` instead of putting a object on the channel to indicate doneness. Even better to do it with a defer to make sure that the code doesn't deadlock if there is a panic:

```
go func() {
   defer close(done)
   ...
}
```

pkg/shared/utils/utils.go
```diff
@@ -53,6 +58,39 @@ func RunCommand(cmd *exec.Cmd) ([]byte, error) {
...
+	}()
+
+	// Start the command and check for errors
+	if err := cmd.Start(); err != nil {
+		return fmt.Errorf("failed to start command: %w", err)
+	}
+
+	<-done
```
Tehsmash, line 85: nit: maybe a comment here like `// Wait for all stdout lines to be read`

Tehsmash: Been thinking and I'm not sure we need this to be a go routine unless we also want to capture any stderr messages in parallel with stdout. As the code is today the scanner could be inlined here.

Tehsmash (changes requested):

pkg/shared/families/malware/yara/yara.go
akpsgit, line 84: Maybe RunCommandAndReadOutputLineByLine?

pkg/shared/utils/utils.go
```diff
@@ -53,6 +57,39 @@ func RunCommand(cmd *exec.Cmd) ([]byte, error) {
...
+	}
+
+	// Merge stdout and stderr
+	merged := io.MultiReader(stderr, stdout)
+	scanner := bufio.NewScanner(merged)
+	// Use the scanner to scan the output line by line and parse it
+	for scanner.Scan() {
+		line := scanner.Text()
+		pfn(line)
+	}
```
Tehsmash, line 83: We need to check scanner.Err() here to make sure we log any errors while parsing the output

pkg/shared/utils/utils.go
```diff
@@ -53,6 +57,39 @@ func RunCommand(cmd *exec.Cmd) ([]byte, error) {
...
+		return fmt.Errorf("failed to get stderr pipe: %w", err)
+	}
+
+	// Start the command and check for errors
+	if err := cmd.Start(); err != nil {
+		return fmt.Errorf("failed to start command: %w", err)
+	}
+
+	// Merge stdout and stderr
+	merged := io.MultiReader(stderr, stdout)
```
Tehsmash, line 77: do we want to pass stdout and stderr to the process function without knowing which is which? It seems like we might get a bunch of unexpected log/info lines coming through which the process function will have to omit. It might not be as big of a problem for Yara, but I know other tools output a log of info stuff on stderr.

It might be better to have a separate scanner go routine, and just throw the stderr lines straight into a logger.Debug so that they appear in the VMClarity scanner output if you turn on debug mode so that we can see them, but we don't need to cache them in memory or pass them to the process function.

pbalogh-sa: I've added a simple stderr logging for now.

Tehsmash: Love it 👍

Tehsmash (approved):

FrimIdan (approved):
LGTM

Issue #718, opened 2023-09-28
Add annotations field to root level API objects


chrisgacsal:
## Problem Statement

Storing additional information (generated at runtime or external components) in the API objects require the API specification being extended with new field(s). The feature set of VMClarity is expending rapidly which means the API spec also getting bigger and more complex. It would be nice to allow additional information being stored in the API objects without exploding the API spec.

## Proposed Solution

Introduce `annotations` field (similarly to API objects in Kubernetes) which defines a map of key-values with `string` type that allows storing extra information in the API object.

## Alternatives Considered

Adding custom fields with`string` type to specific API objects in use-case basis however this could quickly get out of hand and explode the API spec.

## Additional Context

This issue is relevant to #703 and #637.

PR #719, opened 2023-09-28
feat(api): add annotations

paralta:
## Description

* Add annotations API object to close https://github.com/openclarity/vmclarity/issues/718
* `BaseAPIObject` is intended to include a `Metadata` object as well 
* All the main API objects (like assets, asset scans, etc) should contain this new `BaseAPIObject` property (to be added in a separate PR per object)

Please let me know if you have any naming alternatives specially for `BaseAPIObject `, thanks!

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


api/openapi.yaml
```diff
@@ -1277,6 +1277,21 @@ components:
...
+          $ref: '#/components/schemas/Annotations'
+
+    Annotations:
+      type: object
+      description: Generic map of string keys and string values to attach arbitrary non-identifying metadata to objects.
+      properties:
+        key:
+          type: string
+        value:
+          type: string
```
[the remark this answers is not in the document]
Tehsmash: dictionaries with unknown keys don't have any support in oDATA filtering etc, having an array with defined keys will let us do:

```
$filter=annotations/any(t: t/key eq 'blahblah' and t/value eq 'foo')
```

we can also $select to filter things like:

```
$select=annotations($filter=startswith(key, 'vmclarity.io')
```

api/openapi.yaml
```diff
@@ -1277,6 +1277,21 @@ components:
           readOnly: true
       description: An object that is returned in cases of success that returns nothing.
 
+    BaseAPIObject:
```
fishkerez, line 1280: Why not just calling it Metadata?
Don't you think this will be part of the Metadata of the object?

chrisgacsal: @fishkerez I will propose another object called `Metadata` which would be common field for all the root level API objects like the `Annotations`. The `BaseAPIObject` is used to combine all the mandatory/common fields into a single object definition which then can be used for composing new root level API objects and to extend the existing ones. I agree the `BaseAPIObject` is not the best name and I am open for any suggestions other then `Metadata`.

fishkerez: @chrisgacsal I am struggling to understand the difference between Metadata and BaseAPIObject.
**Metadata** - would be common field for all the root level API objects like the Annotations.
**BaseAPIObject** - is used to combine all the mandatory/common fields into a single object definition which then can be used for composing new root level API objects and to extend the existing ones.

So which fields are belong to Metadata and which to the Base? To me it sounds like annotations are a common field that will be used by all object, so not sure to where it belongs.

BTW, in Kubernetes the annotations are part of the metadata.

Tehsmash: correct me if I'm wrong but I would expect it to look something like:

```
BaseAPIObject:
   type: object
   properties:
      metadata:
         type: object
         properties:
            createdAt:
               type: string
               format: date-time
            updatedAt:
            ...
            ...
      annotations:
         type: array
         ...
      id:
        type: string
```

whether annotations is part of metadata or not I'm not bothered, then its used like:

```
Asset:
  allOf:
     - $ref: "..../BaseAPIObject"
     - type: object
        properties:
           assetInfo:
              $ref: ...      
```

chrisgacsal: Ahh @Tehsmash has already described what I did in the previous comment.
I still think it would be better to separate the `Annotations` from `Metadata` as the actor operating on these fields are different. Former is used/managed by components other than `API Server` while the latter must be only managed by the `API Server` only. Defining permissions for fields is easier if these two fields are not embedded in a high level field.

chrisgacsal: @Tehsmash  @fishkerez @paralta Wdyt?

Tehsmash: Honestly I'm not convinced that we need a Metadata wrapper key around any of them if we go with the composite object approach. ID, createdAt, updatedAt, annotations, revision, can all be top level keys for all objects without a need for another layer.

We can create an interface and Get/Set functions for each object if we want to be able to process them in a common way in the go code.

api/openapi.yaml
```diff
@@ -1277,6 +1277,23 @@ components:
           readOnly: true
       description: An object that is returned in cases of success that returns nothing.
 
+    Metadata:
```
chrisgacsal, line 1280:
```suggestion
Metadata:
  type: object
  properties:
    meta:
        type: object
        properties:
          annotations:
            $ref: '#/components/schemas/Annotations'
```

chrisgacsal (changes requested):

api/openapi.yaml
chrisgacsal: Personally, I think would be much cleaner to have a top level key (`metadata`) which includes the metadata properties like (`createdAt`, updatedAt, `revision`, etc). On the other hand the `annotations` definitely needs to be a top level field.

chrisgacsal: Alright. Let's keep it as is for now.

chrisgacsal (approved):
LGTM!

Issue #682, opened 2023-09-12
Error threshold for Yara family


pbalogh-sa:
## Problem Statement

In the case of Yara scan, some errors can occur because of file permissions. We need to define some kind of error threshold that indicates to us the whole scan is in a failure state.

## Proposed Solution

It needs to be researched what should be the proper solution.
1. Check what kind of errors can occur during the Yara scan.
2. Needs to be decided which errors can be skipped and which ones count on the final threshold.
3. Maybe it would be better to use `--scan-list` to compare the error count with the number of scanned files. (this solution slows down the scan because we need to list all of the files and create a file list before initiating a scan)

In the Yara source code, there are many types of errors. Most of them can happen during the compilation of the rules. If we are using raw Yara rules these rule compilation errors can also occur.
All the error messages are printed to the stderr.
There are some that indicate a whole scanner error. For example, if we are using the compiled rule file there can be different errors:

rules were compiled with a different version of YARA
```
invalid compiled rules file
corrupt compiled rules file
could not open file
```
In this case, the return value of the yara is not 0.
And there are some errors that can occur by reading files.
```
error scanning img/xxx: could not open file
possible_includes_base64_packed_functions [impact =5,hide=true,desc="Detects possible includes and packed functions"] img/tttt.svg
```
The return value of the yara is 0.

So I think checking the error threshold should happen in two layers:
1. Need to count all lines in stdout and stderr, and if the `stdout lines/stdout+stderr lines` reach the threshold we will send a scan error.
2. Need to count all lines in stdout and count the parse errors in stdout, and if the `stdout line-parse error count/stdout lines` reach the threshold we will send a parser error message.

The successful scan and parse percentage should be above 20%.

## Alternatives Considered

Just skip the errors.

pbalogh-sa:
In the Yara source code, there are many types of errors. Most of them can happen during the compilation of the rules. If we are using raw Yara rules these rule compilation errors can also occur.
All the error messages are printed to the stderr.
There are some that indicate a whole scanner error. For example, if we are using the compiled rule file there can be different errors:
```
rules were compiled with a different version of YARA
invalid compiled rules file
corrupt compiled rules file
could not open file
```
In this case, the return value of the yara is not 0.
And there are some errors that can occur by reading files.
```
error scanning img/xxx: could not open file
possible_includes_base64_packed_functions [impact =5,hide=true,desc="Detects possible includes and packed functions"] img/tttt.svg
```
The return value of the yara is 0.

[...]

PR #732, opened 2023-10-03
feat(yara): error threshold

pbalogh-sa:
## Description

This PR adds an error threshold for the Yara scan in order to mark the whole scan as failed.
There are two types of errors:
1. count all lines in stdout and stderr, and if the `stdout lines/stdout+stderr lines` reach the threshold we will send a scan error.
2. count all lines in stdout and count the parse errors in stdout, and if the `stdout line-parse error count/stdout lines` reach the threshold we will send a parser error.

fixes: #682 

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


pkg/shared/families/malware/yara/util/util.go
```diff
@@ -43,6 +44,18 @@ func (e *InvalidLineError) Error() string {
...
 }
 
+func IsErrorThresholdReached(successCount, allCount int) bool {
+	switch allCount {
+	case 0:
+	default:
+		if float64(successCount)/float64(allCount) < errThreshold {
+			return true
+		}
+	}
```
Tehsmash, line 54: Using a switch here is a really weird to me I think this can be:

```
if allCount != 0 && float64(successCount)/float64(allCount) < errThreshold {
   return true
}
return false
```

I think its a little confusing that errThreshold is called errThreshold when we're actually comparing it to number of successes. Perhaps it would be easier to understand if the number of errors was the input and we did something like:

```
errThreshold = 0.8 
if float64(errorCount) / float64(allCount) > errThreshold
```

pkg/shared/families/malware/yara/yara.go
```diff
@@ -112,6 +128,27 @@ func (s *Scanner) Run(sourceType utils.SourceType, userInput string) error {
 			return
 		}
 
+		// If the (stdout lines)/(stderr + stdout lines) is less than `errThreshold` the error threshold will be reached.
```
Tehsmash, line 131: " If stderr lines / stderr lines + stdout lines is greater than the err threshold..."

pkg/shared/families/malware/yara/yara.go
```diff
@@ -112,6 +128,27 @@ func (s *Scanner) Run(sourceType utils.SourceType, userInput string) error {
...
+		if util.IsErrorThresholdReached(scanErrCount, outputLines+scanErrCount) {
+			s.sendResults(retResults, fmt.Errorf(
+				"scanner error threshold (%.2f%%) is reached, nuber of errors=%d, sample of errors: %w",
+				util.ErrThreshold*100, // nolint: gomnd
+				scanErrCount,
+				errors.Join(scanErrSamples...),
+			))
+			return
+		}
+		// If the (successfully parsed lines)/(stdout lines) is less than `errThreshold` the error threshold will be reached.
```
Tehsmash, line 141: "If the parse failures / lines parsed is greater than the err threshold then ..."

Tehsmash (changes requested):

pkg/shared/families/malware/yara/yara.go
pbalogh-sa: https://github.com/openclarity/vmclarity/pull/732/commits/7d454c033f9373ffe7bed7259fb80514240d9037

pkg/shared/families/malware/yara/yara.go
pbalogh-sa: https://github.com/openclarity/vmclarity/pull/732/commits/7d454c033f9373ffe7bed7259fb80514240d9037

Tehsmash (approved):

PR #753, opened 2023-10-06
feat(orchestrator): adopt new provider in config and discoverer

paralta:
## Description

https://github.com/openclarity/vmclarity/issues/640

* Update orchestrator to use new provider API object in configuration
* Get provider from environment variable and post it to DB
* Add provider ID to asset once discovered or rediscovered

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/pkg/orchestrator/config.go
+++ b/pkg/orchestrator/config.go
@@ -32,6 +32,7 @@ import (
 	"github.com/openclarity/vmclarity/pkg/orchestrator/scanconfigwatcher"
 	"github.com/openclarity/vmclarity/pkg/orchestrator/scanestimationwatcher"
 	"github.com/openclarity/vmclarity/pkg/orchestrator/scanwatcher"
+	"github.com/openclarity/vmclarity/pkg/shared/utils"
 )
 
 const (
@@ -99,7 +100,7 @@ const (
 )
 
 type Config struct {
-	ProviderKind models.CloudProvider
+	Provider models.Provider
 
 	APIServerHost      string `json:"apiserver-host,omitempty"`
 	APIServerPort      int    `json:"apiserver-port,omitempty"`
@@ -161,6 +162,8 @@ func setConfigDefaults() {
 func LoadConfig() (*Config, error) {
 	setConfigDefaults()
 
+	// TODO(paralta) Needs to change to loop once multiple Providers are supported.
+	// Currently a name string is provided in the config, in the future the Orchestrator will receive a list of Provider IDs.
 	var providerKind models.CloudProvider
 	switch strings.ToLower(viper.GetString(ProviderKind)) {
 	case strings.ToLower(string(models.Azure)):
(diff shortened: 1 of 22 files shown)


pkg/orchestrator/discovery/discoverer.go
```diff
@@ -100,6 +101,10 @@ func (d *Discoverer) DiscoverAndCreateAssets(ctx context.Context) error {
 		// which matches the unique properties of this asset, in this
 		// case we'll patch the just AssetInfo and FirstSeen instead.
 		assetData.FirstSeen = nil
+		assetData.Providers = AppendProviderIfMissing(
+			conflictError.ConflictingAsset.Providers,
+			models.ProviderRelationship{Id: (*d.providerClient).Object().Id},
+		)
```
Tehsmash, line 107: I wonder if we need to handle the reverse case, we need to remove the provider ID from the asset if we do not find it. Alternatively we need to have a "Last Seen" field for each provider on the asset like:

```
{
   "id": "<uuid>",
   "assetInfo": {},
   "providers": [
        {
            "provider": { "id": "<uuid>" },
            "lastSeen": "<time stamp>"
        }
   ]
}
```

this would remove the need for the terminated field. In the orchestrator/scheduler we'd only schedule AssetScans onto providers which have seen the asset in the last 5 / 10 minutes or something like that.

pkg/orchestrator/config.go
```diff
@@ -178,6 +181,14 @@ func LoadConfig() (*Config, error) {
...
 		providerKind = models.AWS
 	}
+	provider := models.Provider{
+		DisplayName: utils.PointerTo(string(providerKind)),
+		Status: &models.ProviderStatus{
+			State:              models.ProviderStatusStateUnknown,
+			Reason:             models.NoHeartbeatReceived,
+			LastTransitionTime: time.Now(),
+		},
+	}
```
Tehsmash, line 191: I don't think we should be creating the provider in the configuration, we should be taking in the provider display name as a config input, I think defaulting to "vmclarity-orchestrator", we can pass through the CFN stack name or something in the installers.
The provider object itself should be created as part of the orchestrator boot up flow using the config.

pkg/orchestrator/orchestrator.go
```diff
@@ -100,12 +100,17 @@ func New(ctx context.Context, config *Config) (*Orchestrator, error) {
...
+	// TODO(paralta) Provider initialization needs to be removed from here once Providers are split from Orchestrator (Issue #643).
+	apiProvider, err := backendClient.PostProvider(ctx, config.Provider)
 	if err != nil {
-		return nil, fmt.Errorf("failed to initialize provider. Provider=%s: %w", config.ProviderKind, err)
+		return nil, fmt.Errorf("failed to post provider: %w", err)
+	}
+	p, err := NewProvider(ctx, apiProvider)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize provider. Provider=%s: %w", *apiProvider.DisplayName, err)
 	}
```
Tehsmash, line 111: @chrisgacsal @paralta I think we need to chat about the initial registration part to avoid creating new providers on orchestrator restart. i.e. we could use passed displayName to check to see if the provider already exists in the API and if it does save that ID for use in the rest of the flow. Alternatively check to see if /etc/vmclarity/providerID exists, and if it does not we POST a new provider and then save the ID into that file. If the file does exist then we PATCH the display name etc.

I think the later might make more sense in the long term but does require persistent storage of some kind (this could be different per provider i.e. configmap for kubernetes, file for VM on AWS, volume for docker. In the future we'll probably need to store some kind of auth certificate anyway.

Tehsmash (changes requested):

paralta:
closing as issue needs to be split between implementing provider registration and updating asset discovery

PR #762, opened 2023-10-09
fix(docker): run docker from within the container

fishkerez:
## Description

When I run docker on my computer (Mac, Intel chip) I can see the discovery of assets fails due to:
`
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
`

We need to mount the docker socket in order to connect to the daemon service from within a container.

## Type of Change

[ *] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)


installation/docker/docker-compose.yml
```diff
@@ -39,6 +39,10 @@ services:
       replicas: 1
       restart_policy:
         condition: on-failure
+    volumes:
+      - type: bind
+        source: /var/run/docker.sock
+        target: /var/run/docker.sock
```
Tehsmash, line 45: We'll probably need to remove this from the e2e compose overrides if we add it here, they also mount /tmp as part of that so we may also require that volume mount.

Tehsmash (dismissed):

installation/docker/docker-compose.yml
chrisgacsal: https://github.com/openclarity/vmclarity/blob/main/e2e/docker-compose.override.yml#L4-L10

fishkerez: Yeah I saw this, do you know the reason we mount /tmp?

fishkerez: @Tehsmash @chrisgacsal any idea why the /tmp mount is needed?

chrisgacsal: @paralta Do you happen to remember why we need `/tmp` to be mounted? Is it a E2E specific requirement?

chrisgacsal: @fishkerez I'd just mvoe the `docker.sock` to the main docker-compose file and leave the `/tmp` in the E2E override file for now. We will revisit the need for `/tmp` separately if that is ok with everyone.

chrisgacsal (approved):
LGTM! :shipit:

Issue #763, opened 2023-10-09
Docker provider - UI doesn't  show assets info


fishkerez:
## Description

On docker provider, our assetTypes are ContainerInfo and ContainerImageInfo.
When looking at the UI, it seems like information is missing:

<img width="1364" alt="Screenshot 2023-10-09 at 15 22 00" src="https://github.com/openclarity/vmclarity/assets/89929372/13b4f088-9912-481a-a579-4acfce680684">

Inspecting further what the UI gets from the API, it seems it gets the info:

```
{
    "count": 42,
    "items": [
        {
            "assetInfo": {
                "architecture": "amd64",
                "id": "sha256:ba43c26ac9c46391d5d7b727241f659618a88fb344809ebedfa0b9e16368b4c5",
                "labels": [],
                "name": "",
                "objectType": "ContainerImageInfo",
                "os": "linux",
                "size": 12161332
            },
            "id": "e45e78d2-7629-466e-8976-93a1c8b37a80",
            "lastSeen": "2023-10-09T12:22:49.068983583Z"
        },
        .....
        {
            "assetInfo": {
                "architecture": "amd64",
                "id": "sha256:76074e0542a31b04ee76ec957efef62d972e0182f388b52a98cc97dc3d87c7fc",
                "labels": [
                    {
                        "key": "org.opencontainers.image.created",
                        "value": "2023-04-28T04:54:41Z"
                    },
                    {
                        "key": "org.opencontainers.image.revision",
                        "value": "1be1e2e6380efd9b63913721db1b9d61e3800126"
                    },
                    {
                        "key": "org.opencontainers.image.title",
                        "value": "trivy"
                    },
                    {
                        "key": "org.opencontainers.image.version",
                        "value": "0.41.0"
                    },
                    {
                        "key": "org.opencontainers.image.vendor",
                        "value": "Aqua Security"
                    },
                    {
                        "key": "org.opencontainers.image.description",
                        "value": "A Fast Vulnerability Scanner for Containers"
                    },
                    {
                        "key": "org.opencontainers.image.documentation",
                        "value": "https://aquasecurity.github.io/trivy/v0.41.0/"
                    },
                    {
                        "key": "org.opencontainers.image.source",
                        "value": "https://github.com/aquasecurity/trivy"
                    },
                    {
                        "key": "org.opencontainers.image.url",
                        "value": "https://www.aquasec.com/products/trivy/"
                    }
                ],
                "name": "",
[...]

PR #772, opened 2023-10-11
fix(provider): wrong container image name

paralta:
## Description

The Container Image name is available inside the `RepoTag` field. Included in this PR:
* Make container and container image id a required field in the API spec
* Remove prefix slash from container name
* Replace image name with repotags and repodigest fields
* Add container info and container image info to UI asset type items
* Extend basic test to scan a docker image

Related to https://github.com/openclarity/vmclarity/issues/763, but does not fix UI issue.

## Type of Change

[x] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


paralta:
Currently looking into the new test failure 👀

api/openapi.yaml
```diff
@@ -2214,6 +2214,7 @@ components:
           nullable: true
       required:
         - objectType
+        - id
```
fishkerez, line 2217: What does "id" represent here? it is not the database id, right?

e2e/basic_scan_test.go
```diff
@@ -45,66 +44,29 @@ var _ = ginkgo.Describe("Running a basic scan (only SBOM)", func() {
...
+	ginkgo.Context("which scans a docker image", func() {
+		ginkgo.It("should finish successfully", func(ctx ginkgo.SpecContext) {
+			ginkgo.By("waiting until test asset is found")
+			filter := "assetInfo/repoTags/any(s:startswith(s, 'alpine:3.18.2'))"
 			reportFailedConfig.objects = append(
 				reportFailedConfig.objects,
-				APIObject{"scanConfig", fmt.Sprintf("id eq '%s'", *apiScanConfig.Id)},
+				APIObject{"asset", filter},
 			)
-
```
fishkerez, line 66: Are these changes related to the PR?

api/openapi.yaml
fishkerez: I think we should call it imageID, to differentiate from the id fields we have on all our resources which represents the database id

fishkerez: containerID in that case

api/openapi.yaml
```diff
@@ -2236,7 +2243,8 @@ components:
         size:
           type: integer
       required:
-        - objectType 
+        - objectType
+        - id
```
fishkerez, line 2247: Let's change the name to imageID

api/openapi.yaml
paralta: Thanks for reviewing, @fishkerez ! Agree with you here, it will also make it more consistent with `VMInfo` where we have a `instanceID` 👍

chrisgacsal (approved):
LGTM! :shipit:

api/models/containerimageinfo.go
```diff
@@ -0,0 +1,40 @@
...
+// limitations under the License.
+
+package models
+
+// GetRepoTag returns the first repo tag if it exists. Otherwise, returns false.
+func (c *ContainerImageInfo) GetRepoTag() (string, bool) {
+	var tag string
+	var ok bool
+
+	if len(*c.RepoTags) > 0 {
```
fishkerez, line 23: need to check for c == nil && c.c.RepoTags == nil ?

api/models/containerimageinfo.go
```diff
@@ -0,0 +1,40 @@
...
+
+	return tag, ok
+}
+
+// GetRepoDigest returns the first repo digest if it exists. Otherwise, returns false.
+func (c *ContainerImageInfo) GetRepoDigest() (string, bool) {
+	var digest string
+	var ok bool
+
+	if len(*c.RepoDigests) > 0 {
```
fishkerez, line 35: need to check for c == nil && c.RepoDigests == nil ?

api/models/containerimageinfo.go
```diff
@@ -0,0 +1,40 @@
...
+	var ok bool
+
+	if len(*c.RepoTags) > 0 {
+		tag, ok = (*c.RepoTags)[0], true
+	}
+
+	return tag, ok
+}
+
+// GetRepoDigest returns the first repo digest if it exists. Otherwise, returns false.
```
fishkerez, line 30: why not returning all the repo digests and let the caller decide if he wants all or just the first one?

api/models/containerimageinfo.go
```diff
@@ -0,0 +1,40 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package models
+
+// GetRepoTag returns the first repo tag if it exists. Otherwise, returns false.
+func (c *ContainerImageInfo) GetRepoTag() (string, bool) {
```
fishkerez, line 19: why not returning all the repo tags and let the caller decide if he wants all or just the first one?

api/models/containerimageinfo.go
chrisgacsal: @fishkerez We can check if `c` is `nil` to make sure the receiver is not `nil`, however `len()` returns `0` if `c.RepoTags` is `nil` so the 2nd check is not needed as `len` does that implicitly.

api/models/containerimageinfo.go
chrisgacsal: @fishkerez Same as above.

api/models/containerimageinfo.go
chrisgacsal: This is just a helper to get the first `RepoDigest` in the list. If the caller want to get all then it can use the `c.RepoDigests`. But we can add another helper which returns all of the `RepoTags` and `RepoDigests` if there are any.

api/models/containerimageinfo.go
chrisgacsal: @fishkerez Same as above.

api/models/containerimageinfo.go
fishkerez: yeah, but we are dereferencing c.RepoTags (by doing *c.RepoTags), so if c.RepoTags are nil, we will crush

chrisgacsal: Testing the receiver for `nil` is not something we do in other parts of the code base as the assumption is that the caller won't call any method on `nil` pointer. Or at least it makes sure that safeguards are in place to avoid this issue. Personally I do not mind to test the receiver as well, but that we need to make sure that we follow the same principle in other parts of the code base as well.

The problem is that what should be the return value if the the receiver is null? An error or `0` value?

chrisgacsal: > yeah, but we are dereferencing c.RepoTags (by doing *c.RepoTags), so if c.RepoTags are nil, we will crush

That is valid point. @paralta Could you please fix this?

api/models/containerimageinfo.go
fishkerez: Ahh ok now I see. So maybe better calling it GetFirstRepoDigest/Tag ?

pkg/orchestrator/provider/docker/container.go
```diff
@@ -103,15 +104,17 @@ func (c *Client) getContainerInfo(ctx context.Context, containerID, imageID stri
 	}
 
 	// Get container image info
-	imageInfo, err := c.getContainerImageInfo(ctx, imageID)
+	imageInfo, err := c.getContainerImageInfo(ctx, info.Image)
 	if err != nil {
 		return models.ContainerInfo{}, err
 	}
 
+	containerName := strings.Trim(info.Name, "/")
```
fishkerez, line 112: Do you want Trim or TrimSuffix here?

paralta: It would be `TrimPrefix` in this case, but I think we are fine since docker does not allow container names with slashes

Tehsmash (changes requested):
I agree with Erez's points, lets get those fixed otherwise LGTM.

pkg/uibackend/rest/dashboard_riskiest_assets_test.go
```diff
@@ -321,7 +321,7 @@ func createContainerImageInfo(t *testing.T) *backendmodels.AssetType {
 	t.Helper()
 	info := backendmodels.AssetType{}
 	err := info.FromContainerImageInfo(backendmodels.ContainerImageInfo{
-		Name: utils.PointerTo("ghcr.io/openclarity/vmclarity-orchestrator"),
+		RepoTags: utils.PointerTo([]string{"ghcr.io/openclarity/vmclarity-orchestrator"}),
```
Tehsmash, line 324: missing the tag part here

pkg/uibackend/rest/dashboard_riskiest_assets.go
```diff
@@ -231,8 +231,15 @@ func containerInfoToAssetInfo(info backendmodels.ContainerInfo) (*models.AssetIn
...
-	repo, name := path.Split(*info.Name)
-	repo = strings.TrimSuffix(repo, "/")
+	var repo, name string
+
+	tag, ok := info.GetRepoTag()
+	if ok {
+		repo, name = path.Split(tag)
+	} else {
+		name = info.ImageID
+	}
```
Tehsmash, line 241: In the kubernetes work I handled this slightly differently: Name is always the ID, Location is the first repo digest (if there is one). The reason for using the whole digest as the location is because the repo is like the street name, but the digest is like the house number. Its an address where you can find the image with this ID if you curl it.

chrisgacsal: Lets have the same logic as @Tehsmash discribed above.

Tehsmash (approved):

Issue #610, opened 2023-08-30
Exploits family scan is failing


fishkerez:
## Description

When running VMClarity scan, with sbom, vulnerabilities and exploits enabled, In the VMClarity UI I see the following error:
`failed to scan input "CVE-2023-0590,CVE-2022-3544.....  for exploits: 1 error occurred: * "exploitdb" job failed: failed to get exploits from cve ids: failed to get exploits via HTTP: failed to fetch exploit: [retry count exceeded retry count exceeded retry count exceeded ....`

## Expected Behavior

Exploits family should finish successfully.

## Actual Behavior

Exploits family scan is failing.

## Affected Version

exploits db version: v0.2.3 

## Steps to Reproduce

1. Deploy VMClarity on AWS CloudFormation.
2. Connect to the UI.
3. Create a new scan configuration with SBOM, vuls and exploits families enabled.
4. Wait for the scan to finish.

Tehsmash:
Did you see any requests making it to the exploit db server?
If you SSH into the VMClarity server can you make a query to the exploit db server port locally and it succeed?

fishkerez:
> Did you see any requests making it to the exploit db server?

No requests were making it to the  exploit db server

> If you SSH into the VMClarity server can you make a query to the exploit db server port locally and it succeed?

Yes

PR #779, opened 2023-10-12
fix(scan): change exploit db address to the apiserver public ip

fishkerez:
## Description

Fix: https://github.com/openclarity/vmclarity/issues/610

The exploit scanner was failing because the he could not make http requests to the server.
This is because he tried to communicate to `http://apiserver:1326` which could not be resolved by the DNS server.
This address will only work on docker installation, where the cli runs on the same host as the server.

For other providers, it should be the apiserver public ip. 

## Type of Change

[* ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)


pkg/orchestrator/config.go
```diff
@@ -179,6 +179,9 @@ func LoadConfig() (*Config, error) {
 		providerKind = models.AWS
 	}
 
+	// TODO this is a bug since it is being resolved to "apiserver" on all of our providers installations.
```
akpsgit, line 182: Not sure it's clear what the bug is in this context. Do you mean that APIServerHost is populated incorrectly by installers, or there is a bug in the code?

fishkerez: I mean the APIServerHost is populated incorrectly by installers

chrisgacsal: I am planning to address this in the PR which aims for refactoring the Orchestrator config to make sure that Orchestrator has a dedicated config option for API server host instead of reusing the config option for API server itself.

chrisgacsal (approved):
LGTM!

Tehsmash (approved):

pkg/orchestrator/config.go
chrisgacsal: Resolving this discussion so we can merge this PR.

PR #791, opened 2023-10-13
fix(ui): name, location, platform, image, launch time of assets

gallotamas:
## Description

Name, location, platform, image and launch time was wrong for ContainerImageInfo and ContainerInfo assets. This PR is fixing them.

Related issue: https://github.com/openclarity/vmclarity/issues/763

Screenshots:
- ContainerInfo
image

- ContainerImageInfo
image

## Type of Change

[x] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


chrisgacsal (approved):
Awesome! Thanks for fixing it! :shipit:

paralta (dismissed):

ui/src/layout/Assets/AssetsTable.js
```diff
@@ -67,7 +67,7 @@ const AssetsTable = () => {
             Header: "Location",
             id: "location",
             sortIds: LOCATION_SORT_IDS,
-            accessor: "assetInfo.location"
+            accessor: (original) => original.assetInfo.location || original.assetInfo.repoDigests?.[0] || original.assetInfo.image?.repoDigests?.[0]
```
Tehsmash, line 70: I'm not sure about using the image repoDigest as the location for a Container that doesn't seem like the right approach. I think ideally if assetInfo.location doesn't exist, then we should put "unknown" or something. Once the multi-cloud stuff merges I think we'll want to replace all of this with the provider(s) which found the asset. @paralta   is adding a note about this to the multi-cloud design doc.

chrisgacsal: @gallotamas Could you pls update the PR with @Tehsmash suggestion. I have overlooked this part. :/

gallotamas: sure, the location will be empty in this case since we don't get the data from the API but you can populate it with the right information and then it will be displayed properly 🙂

chrisgacsal (approved):
Thanks @gallotamas for the fixes! LGTM! :shipit:

chrisgacsal:
@Tehsmash Could you please take another look on this one?

Tehsmash:
This might be out of scope for this PR, but looking at the screenshots shared in the description I'm not sure at this is solving the UX issue that we have. It feels like we're trying to squash all the container and container image specific information into the fields that are specific to the VM asset type. For example why does Container Image Info have an "IMAGE" field...

It feels to me like the page needs to adapt to the type of asset that its displaying, and output the common fields "Type", "First Seen", "Last Seen", "Terminated On" first. Then have an Asset Info section which includes:
- for VMs, we should use the existing fields although Name should be renamed "Instance ID".
- for Container Image assets we should show ImageID, RepoTags and Digests, Architechture, OS, Size.
- for Containers it should have "Labels", "Container Name", "Container ID" + Container Image Info like above.

chrisgacsal:
> This might be out of scope for this PR, but looking at the screenshots shared in the description I'm not sure at this is solving the UX issue that we have. It feels like we're trying to squash all the container and container image specific information into the fields that are specific to the VM asset type. For example why does Container Image Info have an "IMAGE" field...
> 
> It feels to me like the page needs to adapt to the type of asset that its displaying, and output the common fields "Type", "First Seen", "Last Seen", "Terminated On" first. Then have an Asset Info section which includes:
> 
>     * for VMs, we should use the existing fields although Name should be renamed "Instance ID".
> 
>     * for Container Image assets we should show ImageID, RepoTags and Digests, Architechture, OS, Size.
> 
>     * for Containers it should have "Labels", "Container Name", "Container ID" + Container Image Info like above.

Yeah, I've discussed this with @gallotamas last week. The main goal here was to fix what is broken. The presentation layer needs a refactor which will include the rework of the details page for assets based on their types.

But this is definitely out of scope for this PR.

@Tehsmash

Tehsmash:
> Yeah, I've discussed this with @gallotamas last week. The main goal here was to fix what is broken. The presentation layer needs a refactor which will include the rework of the details page for assets based on their types.
> 
> But this is definitely out of scope for this PR.
> 
> @Tehsmash

OK cool as long as we know because honestly these screens make little to no sense right now

Tehsmash (approved):

Tehsmash:
> Yes, @gallotamas showed me the different data objects of types in the openapi.yaml file. This is not just out of scope; unfortunately, much longer work is required to fix it.

fixing it properly maybe, but javascript supports `If objectType == "ContainerImageInfo" { return one set of HTML} else { return other set of HTML }`

PR #805, opened 2023-10-17
fix: azure deploy template have wrong values as defaults

fishkerez:
## Description

The azure deploy template was still using the old backend name in the advanced section where we should be able to override the defaults. This PR fixes it to use the new services (orchestrator, apiserver etc.)

## Type of Change

[* ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)

### DIFF
--- a/installation/azure/vmclarity-UI.json
+++ b/installation/azure/vmclarity-UI.json
@@ -168,12 +168,60 @@
                     "label": "Advanced",
                     "elements": [
                         {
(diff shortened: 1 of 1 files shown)


chrisgacsal (approved):
LGTM! But would love to get others to take a look as well.

Tehsmash (approved):

PR #826, opened 2023-10-18
feat(orchestrator): implement provider registration

paralta:
## Description

Implement https://github.com/openclarity/vmclarity/issues/780

* Currently the provider is not an independent service and communication with the backend (apiserver) service is done via the orchestrator, so for now add the backend client to each provider.
* When an IAM service is available, a bootstrap token will be given to each provider and it will use it perform registration with the apiserver service from which it will get a authentication token that will be used to communicate with this service. For now, just post the provider to the apiserver service and receive back an UUID which will be used as authentication token.
* While persistent storage is not available for most providers, for the docker provider a persistent volume can be easily implemented and the "current registration method" illustrated in the following diagram is implemented here.

Screenshot 2023-10-18 at 15 01 48

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/installation/docker/docker-compose.yml
+++ b/installation/docker/docker-compose.yml
@@ -43,6 +43,9 @@ services:
       - type: bind
         source: /var/run/docker.sock
         target: /var/run/docker.sock
+      - type: volume
+        source: provider-registration-data
+        target: /var/opt/vmclarity
     healthcheck:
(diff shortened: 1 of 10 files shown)


paralta:
closing as new approach will be adopted and the first task will be to split the provider from the orchestrator.

PR #864, opened 2023-10-27
feat(k8s): Asset scan support

Tehsmash:
## Description

Add RunAssetScan implementations for running containers and container images to the Kubernetes Provider. It builds on the foundation introduced by the Kubernetes Provider discovery logic to export containers and container images from the container runtime and then run a scanner job to analyse the resulting tar.

## Type of Change

[ ] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


installation/kubernetes/helm/vmclarity/values.yaml
```diff
@@ -819,3 +819,58 @@ yaraRuleServer:
       # -- List of capabilities to be dropped
       drop:
         - ALL
+
+crDiscoveryServer:
+  image:
+    # -- Container Runtime Discovery Server container registry
+    registry: docker.io
+    # -- Container Runtime Discovery Server container repository
+    repository: tehsmash/vmclarity-cr-discovery-server
```
akpsgit, line 828: should be openclarity/vmclarity-cr-discovery-server?

pkg/orchestrator/provider/kubernetes/container.go
```diff
@@ -72,11 +79,20 @@ func (p *Provider) discoverContainersFromDiscoverer(ctx context.Context, outputC
 		return fmt.Errorf("unable to decode response from discoverer: %w", err)
 	}
 
+	logger := log.GetLoggerFromContextOrDiscard(ctx)
 	for _, container := range containerResponse.Containers {
+		container := container
```
paralta, line 84: not required?

pkg/containerruntimediscovery/docker/discoverer.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (d *discoverer) Image(ctx context.Context, imageID string) (models.ContainerImageInfo, error) {
+	info, err := d.getContainerImageInfo(ctx, imageID)
+	if dclient.IsErrNotFound(err) {
+		return info, types.ErrNotFound
+	}
+	return info, err
+}
+
+func (d *discoverer) getContainerImageInfo(ctx context.Context, imageID string) (models.ContainerImageInfo, error) {
```
ramizpolic, line 79: nit: same method ordering here

pkg/containerruntimediscovery/types/cleanup.go
```diff
@@ -0,0 +1,63 @@
...
+// conditionally. Usage:
+//
+//		 cu := cleanup.Make(func() { f.Close() })
+//		 defer cu.Clean() // failure before release is called will close the file.
+//		 ...
+//	   cu.Add(func() { f2.Close() })  // Adds another cleanup function
+//	   ...
+//		 cu.Release() // on success, aborts closing the file.
+//		 return f
+type Cleanup struct {
```
ramizpolic, line 30: note: `Cleanup` is not concurrently safe, not sure if it would be useful to call concurrently though (e.g. `Add` could definitely be called in parallel)

pkg/containerruntimediscovery/types/discoverer.go
```diff
@@ -0,0 +1,37 @@
...
+type DiscovererFactory func() (Discoverer, error)
+
+type Discoverer interface {
+	Images(ctx context.Context) ([]models.ContainerImageInfo, error)
+	Image(ctx context.Context, imageID string) (models.ContainerImageInfo, error)
+	ExportImage(ctx context.Context, imageID string) (io.ReadCloser, error)
+
+	Containers(ctx context.Context) ([]models.ContainerInfo, error)
+	Container(ctx context.Context, containerID string) (models.ContainerInfo, error)
+	ExportContainer(ctx context.Context, containerID string) (io.ReadCloser, func(), error)
```
ramizpolic, line 34: nit: would be good to document the returned function usage

pkg/orchestrator/provider/kubernetes/config.go
```diff
@@ -23,16 +23,18 @@ import (
...
 
 type Config struct {
-	// KubeConfig defines a path to a kubeconfig file to use to connect to
-	// the Kubernetes API
+	// KubeConfig defines a path to a kubeconfig file to use to connect to the Kubernetes API.
 	KubeConfig string `mapstructure:"kubeconfig"`
 
-	// Container Runtime Discovery Namespace is the namespace where the
-	// container runtime discovery daemonset is running.
+	// ContainerRuntimeDiscoveryNamespace defines the name of the namespace where container runtime discovery DaemonSet is deployed.
```
ramizpolic, line 33:
```suggestion
	// ContainerRuntimeDiscoveryNamespace defines the namespace where container runtime discovery DaemonSet is deployed.
```

pkg/orchestrator/provider/kubernetes/provider.go
```diff
@@ -113,10 +120,215 @@ func (p *Provider) DiscoverAssets(ctx context.Context) provider.AssetDiscoverer
...
+	switch value := objectType.(type) {
+	case models.ContainerImageInfo:
+		var pickedClient *containerruntimediscovery.Client
+		for _, client := range clients {
+			_, err := client.GetImage(ctx, value.ImageID)
+			if err == nil {
+				pickedClient = client
+				break
+			}
+		}
```
ramizpolic, line 152: should we parallelize this? is it okay to not log anything in case all ops fail?

ramizpolic: we could also move it to a separate `selectClient(clients []client, selector func(client) error) (client, error)` function which accepts functor that performs an operation (e.g. `GetImage`, `GetContainer`) and returns a client on success so that we can reuse it for other ops

pkg/orchestrator/provider/kubernetes/provider.go
```diff
@@ -113,10 +120,215 @@ func (p *Provider) DiscoverAssets(ctx context.Context) provider.AssetDiscoverer
...
+						{
+							Name: "asset-data",
+							VolumeSource: corev1.VolumeSource{
+								EmptyDir: &corev1.EmptyDirVolumeSource{},
+							},
+						},
+					},
+				},
+			},
+			BackoffLimit: &backOffLimit,
```
ramizpolic, line 306: `utils.PointerTo(int32(0))`

pkg/orchestrator/provider/kubernetes/provider.go
```diff
@@ -113,10 +120,215 @@ func (p *Provider) DiscoverAssets(ctx context.Context) provider.AssetDiscoverer
...
 
-func (p *Provider) RemoveAssetScan(context.Context, *provider.ScanJobConfig) error {
-	return fmt.Errorf("not implemented")
+func (p *Provider) runScannerJob(ctx context.Context, config *provider.ScanJobConfig, sourceURL string) error {
+	configBytes, err := p.generateScanConfig(config)
+	if err != nil {
+		return fmt.Errorf("failed to generate scanner config yaml: %w", err)
+	}
+
+	jobName := fmt.Sprintf("vmclarity-scan-%s", config.AssetScanID)
```
ramizpolic, line 218: `scanName` seems a bit better because its the name for both the pod, container, and configmap

ramizpolic (dismissed):
lgtm,  quite a lot of work here, well done :+1:
just added a couple of smaller notes to consider

pkg/containerruntimediscovery/containerd/discoverer.go
```diff
@@ -0,0 +1,516 @@
...
+		if err != nil {
+			return false, fmt.Errorf("unable to convert image %s to container image info: %w", image.Name(), err)
+		}
+
+		result, err = result.Merge(cii)
+		if err != nil {
+			return false, fmt.Errorf("unable to merge image %v with %v: %w", result, cii, err)
+		}
+
+		return false, nil
```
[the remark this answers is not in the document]
chrisgacsal: Actually the logic implemented here is correct, it just really confusing. So this walk function should never return with `true` stopping the walk as we want to collect all the images with the same ID and merge their information together into a single `ContainerImageInfo` object which can represent a single image with multiple different repo tags. In case of `containerd` we need to iterate over all the repo tags to find the ones which belong to the same image ID.

I'll refactor this a bit to make it more readable.

e2e/testenv/kubernetes/helm/config.go
```diff
@@ -56,6 +56,17 @@ func WithNamespace(ns string) ConfigOptFn {
 	}
 }
 
+func WithReleaseName(release string) ConfigOptFn {
+	return func(c *Config) error {
+		if c.ReleaseName != "" {
+			return nil
```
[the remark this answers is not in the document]
chrisgacsal: I will update the name of the function to indicate that it is a defaulter.

chrisgacsal:
Thanks for the review @paralta  and @ramizpolic. I'll address your comments asap. Also wanted to mention that my main goal was to ensure correctness of what is implemented in this PR by limiting the number of changes added on top of the original code. All the optimizations and improvements should be added in following PRs in my honest opinion. Ty

pkg/containerruntimediscovery/docker/discoverer.go
chrisgacsal: Same as above.

pkg/containerruntimediscovery/containerd/discoverer.go
```diff
@@ -0,0 +1,516 @@
...
+			return nil, func() {}, fmt.Errorf("image has no known repo digests can not safely fetch it")
+		}
+
+		// TODO(sambetts) Maybe try all the digests in case one has gone missing?
+		ref := (*imageInfo.RepoDigests)[0]
+		_, err = d.client.Pull(ctx, ref)
+		if err != nil {
+			return nil, func() {}, fmt.Errorf("failed to fetch image %s: %w", ref, err)
+		}
+	}
```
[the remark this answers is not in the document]
chrisgacsal: There are parts which are common but they are not the same. Some code optimizations can be added here, but I'd definitely do that in an a followup PR.

pkg/orchestrator/provider/kubernetes/container.go
chrisgacsal: Yup. I am aware that this is not required from Go 1.22, but wanted to make sure that the code is backward compatible (can be built with older Go versions where the toolchain does not enforce the Go version defined in `go.mod`) and also make sure that the code base is aligned. We can remove these at some point though.

pkg/containerruntimediscovery/types/cleanup.go
chrisgacsal: Since is it no used concurrently I'd keep it as is. I'd expect the caller to ensure that either the `Cleanup` is wrapped with proper locking or `Cleanup` is extended with internal locking. We can add locking mechanism to the `Cleanup` struct later on, but I'd not extend the scope of the current PR with this.

pkg/containerruntimediscovery/client.go
```diff
@@ -0,0 +1,169 @@
...
+	case http.StatusOK:
+		var info models.ContainerImageInfo
+		decoder := json.NewDecoder(resp.Body)
+		err := decoder.Decode(&info)
+		if err != nil {
+			return models.ContainerImageInfo{}, fmt.Errorf("failed to decode response: %w", err)
+		}
+		return info, nil
+	case http.StatusNotFound:
+		return models.ContainerImageInfo{}, types.ErrNotFound
```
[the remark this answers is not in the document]
ramizpolic: Created https://github.com/openclarity/vmclarity/issues/1139

pkg/containerruntimediscovery/types/cleanup.go
ramizpolic: Lets just add that as a comment to the Cleanup struct for now

pkg/orchestrator/provider/kubernetes/provider.go
chrisgacsal: I believe it is called `jobName` because it is a batch job object. So I think the naming is correct in this case.

paralta (approved):

pkg/orchestrator/provider/kubernetes/provider.go
ramizpolic: I am okay with the other notes. But we do ignore the context of why the other clients failed since we return a generic error (unable to find image ID). it can happen that the id exists, but there is some other issue with the client. maybe aggregate all the errors together and only log them in case no client can be selected. anyhow, I'm okay with this.

ramizpolic (approved):

Issue #894, opened 2023-11-02
Completed Scans should be changed to Scans in Dashboard


paralta:
## Description

In the main page (dashboard) the `Completed Scans` count actually refers to the total count of `Scans`, including pending and done/completed scans. Thus, either `Completed Scans` should be changed to `Scans` or the count implementation should be fixed.

![Screenshot 2023-11-02 at 18 08 45](https://github.com/openclarity/vmclarity/assets/46568597/0b869dcd-0eec-48f9-8301-cefb9c1312aa)

## Affected Version

Branch main

## Steps to Reproduce

1. Install docker
2. Trigger default scan with empty scope
3. Check Dashboard

## Checklist

- [x] I have read the [contributing guidelines](/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing issue

PR #898, opened 2023-11-03
fix(ui): calculate completed scans count correctly

adamtagscherer:
## Description

Fix for: https://github.com/openclarity/vmclarity/issues/894
On the Dashboard menu the `Completed scans` widgets shows every scan not only the completed ones.
This PR fixes this issue by checking the scans for their state property.

## Type of Change

[x] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[x] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/ui/src/layout/Dashboard/CounterDisplay/index.js
+++ b/ui/src/layout/Dashboard/CounterDisplay/index.js
@@ -1,16 +1,28 @@
-import React from 'react';
+import React, {useMemo} from 'react';
 import { useFetch } from 'hooks';
 import { formatNumber } from 'utils/utils';
 
 import './counter-display.scss';
 
 const CounterDisplay = ({url, title, background}) => {
-    const [{data, error, loading}] = useFetch(url, {queryParams: {"$count": true, "$top": 1, "$select": "id"}});
-    
+    const [{data, error, loading}] = useFetch(url, {queryParams: {"$select": "state"}});
+
+    const completedScans = useMemo(() => {
+        let ret = 0;
+
+        data?.items.forEach((element) => {
+            if (element.state === "Aborted" || element.state === "Failed" || element.state === "Done") {
+                ret++
+            }
+        })
+
+        return ret
+    }, [data])
+
     return (
         <div className="dashboard-counter" style={{background}}>
-            {loading || error ? "" : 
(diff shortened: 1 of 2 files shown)


paralta (approved):
LGTM 👍  thanks for fixing this!

lgecse (approved):
LGTM, thanks @adamtagscherer! :)

ui/src/layout/Dashboard/CounterDisplay/index.js
```diff
@@ -1,19 +1,46 @@
...
+        let ret = 0;
+
+        data?.items.forEach((element) => {
+            if (element.state === "Aborted" || element.state === "Failed" || element.state === "Done") {
+                ret++
+            }
+        })
+
+        return ret
+    }, [data])
```
gallotamas, line 23: This logic can be simplified:
```suggestion
    const completedScans = useMemo(
        () => data?.items.filter(item => item.state === "Aborted" || item.state === "Failed" || item.state === "Done").length ?? 0,
        [data]
    );
```

gallotamas (approved):

gallotamas (approved):

Tehsmash (changes requested):
I think we should be using the $filter here instead of loading every scan into the UI and then filtering it. e.g. 
```
$filter=state eq 'Aborted' or state eq 'Failed' or state eq 'Done'
```

gallotamas (approved):

Tehsmash (approved):

akpsgit:
I see that "Risky Findings" was changed to "Findings" in https://github.com/openclarity/vmclarity/pull/898/files#diff-a7f123acafc6f879519bf54b3e58eb3ebab608f6d1c23a8107b883eee501e8cfR18

The idea was not to count packages, but only findings with Risks.

adamtagscherer:
> I see that "Risky Findings" was changed to "Findings" in https://github.com/openclarity/vmclarity/pull/898/files#diff-a7f123acafc6f879519bf54b3e58eb3ebab608f6d1c23a8107b883eee501e8cfR18
> 
> The idea was not to count packages, but only findings with Risks. Also, maybe worth making such changes if needed in a separated PR.

The problem with this was that the label said "Risky Findings" although we queried every finding without narrowing the query at all.
We can change this query to only GET the Risky Findings but then we have to decide which Findings are risky, so what is the definition of a Risky Finding?

Issue #1041, opened 2023-12-14
Allow multiple assets in findings


ramizpolic:
As as user, I would like to see only the unique findings grouped by assets when viewing the findings page. 
When checking a specific finding, all assets linked to it will be displayed in a table instead of showing a single asset.

Tasks:
- [x] #1042
- [x] #1253 
- [x] #1257 
- [x] #1258 
- [x] #1259 
- [x] #1260 
- [x] #1766

PR #1042, opened 2023-12-14
docs: change asset-findings relationships rfc

ramizpolic:
## Description

Proposes using many-to-many relationship between assets and findings to improve efficiency and achieve API consistency with the existing features. 
This also enables feature-parity between VMClarity and KubeClarity findings.

Epic: #1041 

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[x] Documentation  
[ ] Other (please describe)  

## Checklist


ramizpolic:
Ready for review :no_good:

rfc/change-asset-findings-relationships.md
```diff
@@ -0,0 +1,236 @@
...
+R(assetScan, finding, version, schedule)        = 10^2 * 100^2 - Scans can be scheduled to occur more often than the versioning changes on other models
+```
+
+Therefore, it makes sense to keep track of associations in separate tables between these models to reduce size and improve efficiency.
+Otherwise, sharding or more complex approaches must be used which are not viable options yet.
+
+### Non-goals
+
+This RFC does not intend to propose changes regarding the relationship between findings and asset scans.
+It is assumed that the asset scan in a given finding represents the latest scan that discovered it.
```
paralta, line 87: what about the `foundOn` timestamp? will we lose the timestamp when a certain finding was first found?

paralta: just saw that we only keep the last timestamp below, maybe worth adding a new field?

paralta (dismissed):
LGTM 👍 just a small remark

rfc/change-asset-findings-relationships.md
ramizpolic: The property pair `(foundOn, foundBy)` in this case represents when the finding was last detected because we do not keep the historical data at all. 

Maybe adding an extra field to the summary for `totalDiscovered`  or something similar to indicate how many scans rediscovered the same finding (initial value would be 1). The `(foundOn, foundBy)` can then be used only for the first scan that discovered it. WDYT?

ramizpolic: > what about the foundOn timestamp? will we lose the timestamp when a certain finding was first found?

Yes, it will be replaced. The property pair `(foundOn, foundBy)` in this case represents when the finding was last detected because we do not keep the historical data. The assumption is that a newer scan might update the data for this finding that does not uniquely describe the finding (e.g. updated description). Can also be switched to denote the first scan. Maybe `refreshedOn` could be used to indicate rediscovery/update by a scan while the other fields can be preserved.

In addition, maybe adding an extra field to the summary for `totalDiscovered`  or something similar to indicate how many scans rediscovered the same finding (initial value would be 1). WDYT?

paralta: my comment was more on the user perspective where it might be helpful to check how old a finding really is as well as check if it was just found in the last scan.

ramizpolic: makes sense, thanks for addressing this, will update

ramizpolic: the new changes should provide a bit more clarity. let me know if this addresses @paralta

paralta (approved):

chrisgacsal (approved):

PR #1048, opened 2023-12-15
refactor(api): refactor Scan status property

adamtagscherer:
## Description

This PR refactors the `status` field of the `Scan` API object to be aligned with other objects.

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[x] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


pkg/apiserver/rest/scan_controller.go
```diff
@@ -98,6 +109,27 @@ func (s *ServerImpl) PatchScansScanID(ctx echo.Context, scanID models.ScanID, pa
 		return sendError(ctx, http.StatusBadRequest, fmt.Sprintf("failed to bind request: %v", err))
 	}
 
+	// check that an asset with that id exists
```
paralta, line 112: The comment here should state `// check that a scan with that id exists`.

paralta: Actually, do we need this check? I think we only need to move the valid transition check code block down to line 139, right?

pkg/apiserver/rest/scan_controller.go
```diff
@@ -131,13 +163,39 @@ func (s *ServerImpl) PatchScansScanID(ctx echo.Context, scanID models.ScanID, pa
...
 
+// nolint:cyclop
 func (s *ServerImpl) PutScansScanID(ctx echo.Context, scanID models.ScanID, params models.PutScansScanIDParams) error {
 	var scan models.Scan
 	err := ctx.Bind(&scan)
 	if err != nil {
 		return sendError(ctx, http.StatusBadRequest, fmt.Sprintf("failed to bind request: %v", err))
 	}
 
+	// check that a scan with that id exists
```
paralta, line 174: same comment as above

pkg/orchestrator/scanwatcher/watcher.go
```diff
@@ -399,24 +403,40 @@ func (w *Watcher) reconcileInProgress(ctx context.Context, scan *models.Scan) er
...
+				models.ScanStatusStateFailed,
+				models.ScanStatusReasonAssetScanFailed,
+				utils.PointerTo(
+					fmt.Sprintf(
+						"%d succeeded, %d failed out of %d total asset scans",
+						*assetScans.Count-failedAssetScans,
+						failedAssetScans,
+						*assetScans.Count,
+					),
+				),
```
paralta, line 416: We could avoid repetition of this code block if we define this message where it was previously defined

api/openapi.yaml
```diff
@@ -1685,34 +1661,81 @@ components:
...
+            | Failed     | AssetScanFailed   | At least one AssetScan has ended with an error             |
+            | Failed     | Timeout           | Scan has been timed out                                    |
+            | Done       | NothingToScan     | There are no assets that need to be scanned                |
+            | Done       | Success           | Scan ended successfully without any errors                 |
+          type: string
+          enum:
+            - Created
+            - AssetsDiscovered
+            - AssetScansRunning
+            - Cancellation
```
paralta, line 1717: I suggest `Cancelled` here to make these words a bit more consistent. Maybe also `Timed out` and `Succeeded`?

paralta (commented):
Couple of comments and suggestions, but looks good to me overall 😄

api/models/scanstatus.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (a *ScanStatus) isValidReason(aa *ScanStatus) error {
+	reasons, ok := scanStatusReasonMapping[aa.State]
+	if ok {
+		for _, reason := range reasons {
+			if reason == aa.Reason {
+				return nil
+			}
+		}
+	}
```
ramizpolic, line 97: this can serve as a validator since it does not require another object for checking the reason
```suggestion
func (a *ScanStatus) isValid() error {
    // check if the reason is allowed for the state
	reasons, _ := scanStatusReasonMapping[a.State]
	for _, reason := range reasons {
		if reason == a.Reason {
			return nil
		}
	}

	return fmt.Errorf("invalid reason for state: state=%s reason=%s", a.State, a.Reason)
}
```

pkg/orchestrator/scanwatcher/watcher.go
```diff
@@ -132,44 +132,42 @@ func (w *Watcher) Reconcile(ctx context.Context, event ScanReconcileEvent) error
...
-
-		err = w.backend.PatchScan(ctx, *scan.Id, &models.Scan{
-			State:        scan.State,
-			StateMessage: scan.StateMessage,
-			StateReason:  scan.StateReason,
-		})
+		scan.Status = models.NewScanStatus(
+			models.ScanStatusStateFailed,
+			models.ScanStatusReasonTimeout,
+			utils.PointerTo("Scan has been timed out"),
```
ramizpolic, line 138:
```suggestion
			utils.PointerTo("Scan has timed out"),
```

pkg/apiserver/rest/scan_controller.go
```diff
@@ -131,13 +163,39 @@ func (s *ServerImpl) PatchScansScanID(ctx echo.Context, scanID models.ScanID, pa
...
+		}
+		return sendError(ctx, http.StatusInternalServerError, fmt.Sprintf("failed to get scan: scanID=%v: %v", scanID, err))
+	}
+
+	// check for valid state transition
+	status, ok := scan.GetStatus()
+	if !ok {
+		return sendError(ctx, http.StatusBadRequest, err.Error())
+	}
+	if ok {
```
ramizpolic, line 188: no need for this check

api/models/scanstatus.go
```diff
@@ -0,0 +1,115 @@
...
+
+	return fmt.Errorf("invalid reason for state: state=%s reason=%s", a.State, a.Reason)
```
ramizpolic, line 99: can we make the function signatures consistent for clarity `func (a *ScanStatus) NAME(aa *ScanStatus)` to avoid mixing the objects

ramizpolic (changes requested):

api/models/scanstatus.go
```diff
@@ -0,0 +1,115 @@
...
+	return &ScanStatus{
+		State:              s,
+		Reason:             r,
+		Message:            m,
+		LastTransitionTime: time.Now(),
+	}
+}
+
+func (a *ScanStatus) Equals(aa *ScanStatus) bool {
+	if a.Message == nil && aa.Message != nil {
```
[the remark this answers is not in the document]
chrisgacsal: The `api` package should have 0 3rd party dependencies. So either the `ValueOrZero` must be added to the `api` package or keep it as is. I am good with the current state though.

pkg/apiserver/rest/scan_controller.go
chrisgacsal: _check **if** a scan with that id exists_

api/openapi.yaml
```diff
@@ -1685,34 +1661,81 @@ components:
...
-              type: string
-            stateReason:
-              description: Machine-readable, UpperCamelCase text indicating the reason for the condition's last transition.
-              type: string
-              enum:
-                - Aborted
-                - TimedOut
-                - OneOrMoreAssetFailedToScan
-                - DiscoveryFailed
-                - Unexpected
```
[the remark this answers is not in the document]
adamtagscherer: Yep, it isn't used anywhere if you check it on the main branch.
Same for the `DiscoveryFailed` state.
https://github.com/openclarity/vmclarity/blob/f0834fb37b7ad41f3b26911c67bc95f3abc138f7/api/models/models.gen.go#L146C16-L146C16

e2e/basic_scan_test.go
```diff
@@ -139,9 +139,9 @@ func RunSuccessfulScan(ctx ginkgo.SpecContext, report *ReportFailedConfig, filte
 	ginkgo.By("waiting until scan state changes to done")
 	scanParams = models.GetScansParams{
 		Filter: utils.PointerTo(fmt.Sprintf(
-			"scanConfig/id eq '%s' and state eq '%s'",
+			"scanConfig/id eq '%s' and status/state eq '%s'",
 			*apiScanConfig.Id,
-			models.ScanStateDone,
+			models.AssetScanStatusStateDone,
```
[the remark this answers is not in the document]
adamtagscherer: I added the reason checks and created an issue to add more e2e tests.

https://github.com/openclarity/vmclarity/issues/1066

I'd like to add more outside of this PR because I think that would delay the merging of this PR a lot.

api/models/scanstatus.go
adamtagscherer: Not sure how this function can replace the current one, can you elaborate on the caller side of your version please?

api/openapi.yaml
paralta: > the scan state is aborted due to cancellation

following this format and given that we kept the `cancellation` reason, should we then change `Created` to `Creation`?

akijakya: Just chiming in with a quick question: how come `Cancellation` can result in both `Aborted` and `Failed` states?

api/openapi.yaml
```diff
@@ -1685,34 +1661,81 @@ components:
...
+            Describes the state of Scan.
+            
+            | State      | Description                                       |
+            | ---------- | ------------------------------------------------- |
+            | Pending    | Scan is waiting to be scheduled                   |
+            | Discovered | Assets have been found for the Scan               |
+            | InProgress | The AssetScans are running on the target machines |
+            | Aborted    | Scan has been aborted through the API             |
+            | Failed     | At least one AssetScan ended with an error        |
+            | Done       | Every AssetScan ended successfully                |
```
ramizpolic, line 1688: Similar to Kubernetes, we could have a bit more descriptive states:
- **Pending** - The initial state when the scan is created, and is working to discover assets
- **Discovered**: The asset scans have been discovered and are waiting to be scheduled
- **InProgress**: The asset scans have been scheduled and the assets are being scanned
- **Aborted**: The scan has been manually aborted
- **Failed**: One or more assets could not be successfully scanned
- **Done**: All assets have been scanned successfully

Later on, additional more granular states can be added to indicate transitions between different states e.g. `Discovered` to `InProgress` through `Scheduling`, `AssetScanCreating`, etc.

api/openapi.yaml
```diff
@@ -1685,34 +1661,81 @@ components:
...
+          description: |
+            Machine readable reason for state transition.
+            
+            | State      | Reason            | Description                                                |
+            | ---------- | ----------------- | ---------------------------------------------------------- |
+            | Pending    | Created           | Scan is pending as it has been newly created               |
+            | Discovered | AssetsDiscovered  | Assets that fall into the scope of the Scan are discovered |
+            | InProgress | AssetScansRunning | AssetScans are currently running                           |
+            | Aborted    | Cancellation      | Scan has been aborted due to cancellation request          |
+            | Failed     | Cancellation      | Scan has been aborted due to cancellation request          |
```
ramizpolic, line 1707: How do we get the scan being cancelled and ending up in `Aborted` or `Failed` state? I think it should be mentioned why it ends up in the given state due to cancellation. If there is no difference, then one can be removed (either _Aborted_ can act as _Failed_ state, or _Failed_ state should not include cancellation event/reason)

ramizpolic (dismissed):

api/models/scanstatus.go
ramizpolic: Sure, added an example for this in the new review

api/openapi.yaml
ramizpolic: @chrisgacsal please add input here because the `reason` messages do not need to be so contextually strict to understand the `state` on the status

ramizpolic: @chrisgacsal feel free to add input here because the `reason` messages do not need to be so contextually strict to understand the `state` on the status. wdyt?

api/openapi.yaml
adamtagscherer: The Scan first gets PATCH-ed to `Aborted` state and it starts updating the related AssetScans to `Aborted` state as well. After it finished updating every AssetScan it moves to `Failed` state. The difference before this PR was that the `Aborted` state of the Scan didn't had any Reason, and the Reason for the `Failed` state was `Aborted`, now both has `Cancellation` state. I extended the description with this information.

api/openapi.yaml
adamtagscherer: I answered under Ramiz's comment: https://github.com/openclarity/vmclarity/pull/1048#discussion_r1445152125

ramizpolic (approved):

PR #1111, opened 2024-01-21
feat: add docker bake

### LINKED ISSUE #1085: Use Bake for Docker builds
## Problem Statement

Using a more modular approach is to be expected for the project which would benefit from using Bake for Docker Builds. This could also be a great opportunity to refine the labeling of our docker images.

## Proposed Solution

Utilize Docker Bake, and do some housekeeping around how we build docker images.

## Alternatives Considered

-

## Additional Context

-

akijakya:
## Description

Added Docker Bake to streamline docker builds, also modified the Makefile use bake for building and pushing images to the repository.

Modifying the CI will come in another PR.

Build time remained the same, but using Docker Bake comes with some advantages:

### modularity 🧱 

- from now on Dockerfiles can exist in any place, number, or name within the repo (or even remote URLs can be used); they can be referred to properly in `docker-bake.hcl` with the `context` and `dockerfile` fields
- multiple images can be built at once via target groups

### tidiness ✨ 

- Makefile is a bit cleaner now by not having to specify the same build commands per each image

Fixes #1085 

## Type of Change

[ ] Bug Fix  
[X] New Feature  
[ ] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,13 +30,6 @@ HELM_CHART_DIR := $(INSTALLATION_DIR)/kubernetes/helm
 HELM_OCI_REPOSITORY := ghcr.io/openclarity/charts
 DIST_DIR ?= $(ROOT_DIR)/dist
 
-VMCLARITY_APISERVER_IMAGE = $(DOCKER_REGISTRY)/vmclarity-apiserver:$(DOCKER_TAG)
-VMCLARITY_ORCHESTRATOR_IMAGE = $(DOCKER_REGISTRY)/vmclarity-orchestrator:$(DOCKER_TAG)
-VMCLARITY_UI_IMAGE = $(DOCKER_REGISTRY)/vmclarity-ui:$(DOCKER_TAG)
-VMCLARITY_UIBACKEND_IMAGE = $(DOCKER_REGISTRY)/vmclarity-ui-backend:$(DOCKER_TAG)
-VMCLARITY_SCANNER_IMAGE = $(DOCKER_REGISTRY)/vmclarity-cli:$(DOCKER_TAG)
-VMCLARITY_CR_DISCOVERY_SERVER_IMAGE = $(DOCKER_REGISTRY)/vmclarity-cr-discovery-server:$(DOCKER_TAG)
-
 ####
 ## Load additional makefiles
 ####
@@ -146,13 +139,13 @@ $(FIXGOMODULES):
 fix: bin/golangci-lint $(FIXGOMODULES) ## Fix linter errors in Go source code
 
 .PHONY: e2e
-e2e: docker-apiserver docker-cli docker-orchestrator docker-ui docker-ui-backend docker-cr-discovery-server ## Run end-to-end test suite
-	export VMCLARITY_E2E_APISERVER_IMAGE=$(VMCLARITY_APISERVER_IMAGE) \
-           VMCLARITY_E2E_ORCHESTRATOR_IMAGE=$(VMCLARITY_ORCHESTRATOR_IMAGE) \
-           VMCLARITY_E2E_UI_IMAGE=$(VMCLARITY_UI_IMAGE) \
-           VMCLARITY_E2E_UIBACKEND_IMAGE=$(VMCLARITY_UIBACKEND_IMAGE) \
-           VMCLARITY_E2E_SCANNER_IMAGE=$(VMCLARITY_SCANNER_IMAGE) \
-           VMCLARITY_E2E_CR_DISCOVERY_SERVER_IMAGE=$(VMCLARITY_CR_DISCOVERY_SERVER_IMAGE) && \
+e2e: docker ## Run end-to-end test suite
+	export VMCLARITY_E2E_APISERVER_IMAGE=$(DOCKER_REGISTRY)/vmclarity-apiserver:$(DOCKER_TAG) \
+           VMCLARITY_E2E_ORCHESTRATOR_IMAGE=$(DOCKER_REGISTRY)/vmclarity-orchestrator:$(DOCKER_TAG) \
+           VMCLARITY_E2E_UI_IMAGE=$(DOCKER_REGISTRY)/vmclarity-ui:$(DOCKER_TAG) \
+           VMCLARITY_E2E_UIBACKEND_IMAGE=$(DOCKER_REGISTRY)/vmclarity-ui-backend:$(DOCKER_TAG) \
+           VMCLARITY_E2E_SCANNER_IMAGE=$(DOCKER_REGISTRY)/vmclarity-cli:$(DOCKER_TAG) \
+           VMCLARITY_E2E_CR_DISCOVERY_SERVER_IMAGE=$(DOCKER_REGISTRY)/vmclarity-cr-discovery-server:$(DOCKER_TAG) && \
 	cd e2e && \
 	go test -v -failfast -test.v -test.paniconexit0 -timeout 2h -ginkgo.v .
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-push-component.yaml
```diff
@@ -24,34 +20,24 @@ on:
...
         type: string
         description: 'The build timestamp to be used for binaries.'
+      bake_target:
+        required: true
+        type: string
+        description: 'Docker Bake target name.'
+      cache-name-prefix:
+        required: true
+        type: string
+        description: 'A prefix for image names and digests to be cached temporarily.'
```
ramizpolic, line 30: Any particular reason why it's required to set the cache name? Can this not be done automatically?

.github/workflows/build-and-push-component.yaml
```diff
@@ -24,34 +20,24 @@ on:
...
+      post_image_tags:
+        required: false
+        type: boolean
+        description: |
+          If set to true the image tags pushed to the repository are posted as comment for the Pull Request.
+          Only works if the event type is `pull_request`.
```
ramizpolic, line 36: Is this really needed for this action? afaik this will never be hit as this is executed manually. only the reusable action requires this

.github/workflows/build-and-push-component.yaml
```diff
@@ -24,34 +20,24 @@ on:
...
-          ##    $ basename ghcr.io/openclarity/vmclarity-ui-dev
-          ##    vmclarity-ui-dev
-          ##
-          image_name="$(basename ${{ inputs.image_name }})"
-          
-          ##
-          ## Set digests cache name
-          ##
-
-          echo "digests-cache-name=digests-${{ github.run_id }}-${image_name}" >> "$GITHUB_OUTPUT"
```
ramizpolic, line 50: can this be preserved to avoid having to specify it manually?

.github/workflows/build-and-push-component.yaml
```diff
@@ -81,35 +67,33 @@ jobs:
...
-          cache-to: type=local,dest=/tmp/.buildx-cache
-          build-args: |
-            VERSION=${{ inputs.image_tag }}
-            BUILD_TIMESTAMP=${{ inputs.timestamp }}
-            COMMIT_HASH=${{ github.sha }}
+          targets: ${{ inputs.bake_target }}
+          set: |
+            *.platform=${{ matrix.platform }}
+            *.output=type=image,name=${{ inputs.image_name }},push-by-digest=true,name-canonical=true,push=${{ inputs.push }}
+            *.tags=
```
ramizpolic, line 77: I would configure caching here rather than specifying it in the bake target

Makefile
```diff
@@ -191,92 +184,87 @@ test: ## Run Go unit tests
...
 	$(info Building apiserver docker image ...)
-	docker build --file ./Dockerfile.apiserver --build-arg VERSION=$(VERSION) \
-		--build-arg BUILD_TIMESTAMP=$(BUILD_TIMESTAMP) \
-		--build-arg COMMIT_HASH=$(COMMIT_HASH) \
-		-t $(VMCLARITY_APISERVER_IMAGE) .
-
-ifneq ($(strip $(VMCLARITY_TOOLS_BASE)),)
-VMCLARITY_TOOLS_CLI_DOCKER_ARG=--build-arg VMCLARITY_TOOLS_BASE=${VMCLARITY_TOOLS_BASE}
-endif
+	docker buildx bake apiserver
```
ramizpolic, line 202: `make docker-apiserver` returns `ERROR: failed to find target apiserver`
The proper name is `vmclarity-apiserver` as defined in the bake. Same should be updated for all other docker related make targets.
```suggestion
	docker buildx bake vmclarity-apiserver
```

docker-bake.hcl
```diff
@@ -0,0 +1,108 @@
...
+	targets = ["vmclarity-apiserver", "vmclarity-cli", "vmclarity-orchestrator", "vmclarity-ui", "vmclarity-ui-backend"]
+}
+
+target "_common" {
+	labels = {
+		"org.opencontainers.image.url" = "https://github.com/openclarity/vmclarity"
+		"org.opencontainers.image.licenses" = "Apache-2.0"
+	}
+	cache-to = ["type=local,dest=/tmp/.buildx-cache"]
+	cache-from = ["type=local,src=/tmp/.buildx-cache"]
```
ramizpolic, line 29: By default, the build is locally cached. It can be improved a bit with this, but I haven't noticed many changes. For CICD, this is definitely useful (GHA cache driver is supported but in beta, which can avoid the local dir cache driver). Not sure if you managed to benchmark this locally. 

I would prefer if these configurations were empty for local build, but configured for CICD only. WDYT?

Also, to enable local cache, I had to create another builder via:
```
$ docker buildx bake vmclarity-apiserver
# ERROR: Cache export is not supported for the docker driver.

$ docker buildx create --name container --driver=docker-container

$ docker buildx bake vmclarity-apiserver --builder=container
# ... works normally
```

docker-bake.hcl
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+target "vmclarity-cli" {
+	context = "."
+	dockerfile = "Dockerfile.cli"
+	tags = get_tag("${target.vmclarity-cli.name}")
+	inherits = ["_common", "_common_args_for_go"]
+	args = {
+		VMCLARITY_TOOLS_BASE = "${VMCLARITY_TOOLS_BASE}"
```
ramizpolic, line 58: Default `make docker-cli` complains that `ERROR: failed to solve: base name (${VMCLARITY_TOOLS_BASE}) should not be blank` which overwrites the default value specified as ARG in CLI docker image.
Commenting these args resolves the error.

ramizpolic (changes requested):

docker-bake.hcl
akijakya: On second thought, @chrisgacsal had bad experiences with `gha` cache type, we should stick to `local`.

docker-bake.hcl
```diff
@@ -0,0 +1,110 @@
...
+variable "COMMIT_HASH" {default = null}
+variable "VMCLARITY_TOOLS_BASE" {default = null}
+
+function "get_tag" {
+  params = [name]
+  result = ["${DOCKER_REGISTRY}/${name}${SUFFIX}:${DOCKER_TAG}"]
+}
+
+group "default" {
+	targets = ["vmclarity-apiserver", "vmclarity-cli", "vmclarity-cr-discovery-server", "vmclarity-orchestrator", "vmclarity-ui", "vmclarity-ui-backend"]
```
chrisgacsal, line 20: I'd make the `targets` definition multiline.

docker-bake.hcl
```diff
@@ -0,0 +1,110 @@
...
+}
+
+group "e2e" {
+	targets = ["vmclarity-apiserver", "vmclarity-cli", "vmclarity-orchestrator", "vmclarity-ui", "vmclarity-ui-backend"]
```
chrisgacsal, line 24: Same as above.

docker-bake.hcl
```diff
@@ -0,0 +1,110 @@
...
+target "_common" {
+	labels = {
+		"org.opencontainers.image.url" = "https://github.com/openclarity/vmclarity"
+		"org.opencontainers.image.licenses" = "Apache-2.0"
+	}
+	output = ["type=image"]
+}
+
+target "_common_args_for_go" {
+  args = {
```
chrisgacsal, line 36: Indentation is off.

docker-bake.hcl
```diff
@@ -0,0 +1,110 @@
...
+	}
+}
+
+target "vmclarity-cr-discovery-server" {
+	context = "."
+	dockerfile = "Dockerfile.cr-discovery-server"
+	tags = get_tag("${target.vmclarity-cr-discovery-server.name}")
+	inherits = ["_common", "_common_args_for_go"]
+	labels = {
+		"org.opencontainers.image.title" = "K8S Image Resolver"
```
chrisgacsal, line 74:
```suggestion
		"org.opencontainers.image.title" = "Container Runtime Discovery Server"
```

docker-bake.hcl
chrisgacsal: Also do we really need this group? I think we should be ok with only having the `default` group. We do plan to extend E2E to be performed on K8s deployments as well.

Makefile
```diff
@@ -205,92 +198,88 @@ test: $(TESTGOMODULES) ## Run Go unit tests
...
 .PHONY: docker
-docker: docker-apiserver docker-cli docker-orchestrator docker-ui docker-ui-backend docker-cr-discovery-server ## Build All Docker images
+docker: ## Build All Docker images
+	$(info Building all docker images ...)
+	docker buildx bake
+
+.PHONY: docker-e2e
+docker-e2e: ## Build Docker images for e2e tests
+	$(info Building docker images for e2e tests ...)
+	docker buildx bake e2e
```
chrisgacsal, line 212: I am not sure if we need this. See my comment below.

Makefile
```diff
@@ -205,92 +198,88 @@ test: $(TESTGOMODULES) ## Run Go unit tests
 
 ##@ Docker
 
+# Export params required in Docker Bake
+export DOCKER_REGISTRY DOCKER_TAG VERSION BUILD_TIMESTAMP COMMIT_HASH VMCLARITY_TOOLS_CLI_DOCKER_ARG=$(VMCLARITY_TOOLS_BASE)
```
chrisgacsal, line 202: I'd prefer using `docker buildx bake --set` argmunet to provide variables to the build instead of setting this variable for all the targets in the Makefile. Like:

```makefile
BAKE_OPTS =
BAKE_OPTS += --set DOCKER_REGISTRY=$(DOCKER_REGISTRY)
BAKE_OPTS += --set DOCKER_TAG=$(DOCKER_TAG)
...
```

```Makefile
docker: ## Build All Docker images
      	$(info Building all docker images ...)
        docker buildx bake $(BAKE_OPTS)
```

chrisgacsal (changes requested):

.github/workflows/build-and-push-component.yaml
akijakya: CI changes will be part of another PR

.github/workflows/build-and-push-component.yaml
akijakya: CI changes will be part of another PR

.github/workflows/build-and-push-component.yaml
akijakya: CI changes will be part of another PR

docker-bake.hcl
akijakya: when I created the group, building the `cr-discovery-server` image was not yet part of the CI and e2e workflows, now it is not necessary indeed

.github/workflows/build-and-push-component.yaml
ramizpolic: could you please collect these comments that are planned for another PR in a separate issue @akijakya

ramizpolic (approved):
lgtm, tested make & bake and it works properly now, thanks @akijakya!

PR #1114, opened 2024-01-22
docs: extend misconfiguration scanners rfc

ramizpolic:
## Description

This PR includes an RFC that aims to:
- extends necessary logic to enable generic misconfiguration scanning capabilities, in order to
- provide a simple way to migrate CIS Docker Benchmark scanner from KubeClarity

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[x] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/rfc/extend-misconfiguration-scanners.md
+++ b/rfc/extend-misconfiguration-scanners.md
@@ -12,34 +12,34 @@
 
 ---
 
-This RFC proposes the extension of misconfiguration scanning capabilities to integrate CIS Docker Benchmark and enrich security findings on assets.
+This RFC proposes the extension of misconfiguration scanning logic to integrate CIS Docker Benchmark and enrich security findings on assets.
 
 ## Background
 
 > [!NOTE]
 > The scanning logic relies on using explicit scopes such as vulnerabilities and misconfigurations to categorize security findings on assets.
 Generally, this works well when scanners have a well-defined boundary used to determine a specific scope.
-However, some scanners cannot directly categorize findings by a single scope which can limit integration options.
+However, some scanners cannot directly categorize findings using a single or existing scope which can limit integration options.
 This behavior, along with the lack of dynamic- and multi-scope options, also underlines an important limitation of how findings are being described, categorized, processed, and analyzed.
 Note that this RFC does not intend to resolve this behavior, but rather draw attention to it.
 
 The integration of [CIS Docker Benchmark](https://github.com/goodwithtech/dockle) scanner requires additional changes to address the scope-based categorization limitations.
-In KubeClarity, this scanner defines its own findings' model as described in the [API specifications](https://github.com/openclarity/kubeclarity/blob/5ac3048b7a782c900a9bef846a91a7735ba77e24/api/swagger.yaml#L243C26-L243C26).
-This makes the migration of scanning capabilities to VMClarity problematic for two main reasons:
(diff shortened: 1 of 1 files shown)


rfc/extend-misconfiguration-scanners.md
```diff
@@ -0,0 +1,106 @@
...
+
+The model preserves most of its original properties but is generalized to enable direct conversion from CIS Docker Benchmark results.
+
+```yaml
+Misconfiguration:
+  type: object
+  properties:
+    scannerName: # preserved
+      type: string
+    code: # replaces `testID`; maps CISDockerBenchmarkResultsEX.code
```
paralta, line 69: `code` is not very intuitive for me here but tbh I do not have a better suggestion 😞 `testID` or something like `checkID` are also not the best options here.

Your description is clear anyway 😄

paralta (approved):
Nice work 👍

rfc/extend-misconfiguration-scanners.md
ramizpolic: I agree, left it as `code` due to a lack of a better alternative. `checkID` makes more sense to me, but I wanted to avoid the `ID` similar to other findings.

Tehsmash: "TestID" was picked because it was somewhat generic, i.e. Lynis might have an ID like "DB034" and Docker Bench has something like "1.1.4". In all cases its a specific to the scanner i.e. Scanner + TestID uniquely identifies a specific test (does a file exist, or is a config set on github) that failed.

This was something we wanted to revisit because if two scanners in the system detect the same issue, they are reported as different findings (essentially every test for every unique location for every scanner that fails is a unique finding). Ideally we should be able to merge them together somehow, for example if two scanners find that a file as two many permissions then it could be a Finding structured like:

All failures for a specific file are bundled together:

```
Misconfiguration:
   Location: <file path>
   FailedTests:
      - Scanner: XYZ
         TestID: FooBar
      - Scanner: Ham
         TestID: Cheese
      - Scanner: Ham
         TestID: Mayo
```
       
Or alternatively we somehow understand which tests are for a specific misconfiguration and then bundle them per name and location:

```
Misconfiguration:
   MisconfigurationName: Too Many Permissions
   Location: <file path>
   Foundby:
      - Scanner: XYZ
         TestID: FooBar
      - Scanner: Ham
         TestID: Cheese
      - Scanner: Ham
         TestID: Mayo

ramizpolic: A similar approach can be reused for other findings as well @Tehsmash. I also have concerns regarding findings in general due to these limitations (merging issues, data duplication, provider dependencies, extension limitations, etc) which I am currently working on. Thanks a lot for addressing this @Tehsmash!

The plan here is to integrate the CIS Docker quickly with minimal changes to the API, while the other proposal will tackle these finding-related challenges in one go.

ramizpolic: A similar approach can be reused for other findings as well. I also have concerns regarding findings in general due to these limitations (merging, data duplication, provider dependencies, extension limitations, etc) which I am currently working on. Thanks a lot for addressing this @Tehsmash!

The plan for this proposal is to integrate the CIS Docker quickly with minimal changes to the API, while the other proposal will tackle these finding-related challenges in one go.

chrisgacsal: I'd use plain `id` instead of `code`. Latter is kinda misleading.

chrisgacsal (commented):
Tiny remark, otherwise looks good to me.

paralta (approved):

chrisgacsal (approved):
LGTM! :shipit:

PR #1185, opened 2024-02-08
ci: use Docker Bake for build and push

### LINKED ISSUE #1157: GHA to use docker bake for building and publishing images
Use Docker Bake for building and pushing images in Github Actions.

akijakya:
## Description

Use Docker Bake for build and push workflows.

Build and push is now image name agnostic, meaning that it only takes a `bake group` name as input and builds the target images defined in the group as long as the bake target name is the same as the image name. The `reusable-build-and-push` now splits the bake group into its targets and builds them in a matrix with the `build-and-push-component`, which also builds the target in a matrix of the defined platforms. 

As a result of this, the image names are not calculated beforehand, instead, they are the output of the bake action, so only parts from `reusable-image-names.yml` were used instead of the workflow itself in this build process (I kept it for the `reusable-end-to-end-testing` workflow but I think those two now could also be merged). Also, image names are uploaded as artifacts for other jobs to use.

Fixes #1157 

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


.github/workflows/reusable-build-and-push.yml
```diff
@@ -39,112 +52,88 @@ jobs:
...
+          } >> "$GITHUB_ENV"
+
       - name: Post comment with image tags
         uses: marocchino/sticky-pull-request-comment@331f8f5b4215f0445d3c07b4967662a32a2d3e31 # v2
         with:
           hide_and_recreate: true
           hide_classify: "OUTDATED"
           skip_unchanged: true
           header: image-tags
+          append: true
```
ramizpolic, line 134: what if we used `recreate: true` instead of adding new ones to ensure that only the latest built images are present instead of showing the comment for every commit?

ramizpolic (dismissed):
added one note to consider, otherwise lgtm :rocket:

.github/workflows/reusable-build-and-push.yml
akijakya: Oh, I must have left this line here by accident btw, thanks for noticing!

chrisgacsal (approved):

PR #1351, opened 2024-02-22
refactor: Azure provider v2

akijakya:
## Description

Move code of Azure provider to `provider/v2`.

Tested using [this](https://openclarity.io/docs/vmclarity/getting-started/deploy-azure/) guide with images created with the `GO_BUILD_TAGS=providerv2 DOCKER_PUSH=true DOCKER_REGISTRY=<myregistry> make docker` command.

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/provider/v2/azure/common/common.go
+++ b/provider/v2/azure/common/common.go
@@ -1,0 +1,51 @@
+// Copyright © 2023 Cisco Systems, Inc. and its affiliates.
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+
+	"github.com/openclarity/vmclarity/provider"
+)
+
+func HandleAzureRequestError(err error, actionTmpl string, parts ...interface{}) (bool, error) {
+	action := fmt.Sprintf(actionTmpl, parts...)
+
+	var respError *azcore.ResponseError
+	if !errors.As(err, &respError) {
+		// Error should be an azcore.ResponseError otherwise something
+		// bad has happened in the client.
+		return false, provider.FatalErrorf("unexpected error from azure while %s: %w", action, err)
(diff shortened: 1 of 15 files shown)


provider/v2/azure/common/common.go
ramizpolic, line 1: this can be moved under `provider/v2/azure/utils`

provider/v2/azure/common/config.go
ramizpolic, line 1: config should be moved to the parent level under `provider/v2/azure`

provider/v2/azure/discoverer/discoverer.go
```diff
@@ -17,15 +17,142 @@ package discoverer
...
+	ret.Encrypted = isEncrypted(osDisk)
+	ret.SizeGB = int(to.ValueOrZero(osDisk.Disk.Properties.DiskSizeGB))
+
+	return ret
+}
+
+func getVMInfoFromVirtualMachine(vm *armcompute.VirtualMachine, rootVol *apitypes.RootVolume) (apitypes.AssetType, error) {
+	assetType := apitypes.AssetType{}
+	err := assetType.FromVMInfo(apitypes.VMInfo{
+		ObjectType:       "VMInfo",
```
ramizpolic, line 109: note: we should probably make these items constants in `vmclarity/api/types` in another PR

provider/v2/azure/common/common.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+
+	"github.com/openclarity/vmclarity/provider"
+)
+
+func HandleAzureRequestError(err error, actionTmpl string, parts ...interface{}) (bool, error) {
```
ramizpolic, line 29: note: this return obscures what is being returned when writing business code. I think this could be split into two helpers `HandleAzureRequestError` that only returns err, and `HandleAzureRequestStatusCodeError` that returns HTTP status code (its easier and much clearer to do a check against `http.StatusNotFound`) and the error. We can do this in another PR for all providers, but I am not finding it easy to understand what is happening in the implementation part when dealing with these error helpers.

provider/v2/azure/common/config.go
akijakya: It is added as a field to the `Scanner` struct so it would cause an import cycle. But now that I have taken a better look at it, the `Config` struct is basically a config for the `Scanner` and could be moved to the `scanner` package instead? Only the `SubscriptionId` field is used for creating the `Discoverer` struct.

provider/v2/azure/discoverer/discoverer.go
ramizpolic: created https://github.com/openclarity/vmclarity/issues/1367

ramizpolic (approved):
lgtm 🚀

chrisgacsal (approved):

PR #1381, opened 2024-02-29
feat: add initial version of workflow library

chrisgacsal:
## Description

Add initial version of the workflow manager which allows to run group of tasks operating a shared state in the right order based on their dependencies.

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


lgecse (approved):

workflow/internal/dispatcher/status.go
```diff
@@ -0,0 +1,81 @@
...
+		return "Done"
+	case FAILED:
+		return "Failed"
+	default:
+		return "Unknown"
+	}
+}
+
+type TaskStatus struct {
+	id    string
```
ramizpolic, line 47: wdy if we remove the ID from the task status? as i noticed, we use this in `Dispatcher.set` and `Dispatcher.update`, but i think it might be cleaner to explicity pass the id to those two methods than infer from the status data. `TaskStatus.String` is not used anywhere where we dont already have ID anyway.

workflow/workflow.go
```diff
@@ -0,0 +1,124 @@
...
+type visitor struct {
+	Order []string
+}
+
+func (v *visitor) Visit(vertex dag.Vertexer) {
+	if v.Order == nil {
+		v.Order = make([]string, 0)
+	}
+	id, _ := vertex.Vertex()
+	v.Order = append(v.Order, id)
```
ramizpolic, line 123: since we already know the dependencies between tasks and have access to the visitor struct, did you consider creating the tasks (as goroutines) here and use channels directly to handle starting and ending their runs (based on the dependency execution state passed back via these channels)? this would make the dispatcher redundant and the implementation a bit more straightforward. in addition, the usage of both channels and mutex would drop to just one of those.

workflow/internal/dispatcher/status.go
chrisgacsal: Originally I used that approach. I'll revert to that.

workflow/internal/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,183 @@
...
+			}
+		}
+	}
+}
+
+func (d *Dispatcher[S, R]) stop() {
+	select {
+	case d.exitChan <- struct{}{}:
+	default:
+	}
```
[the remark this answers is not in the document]
chrisgacsal: Used for testing and forgot to remove.

workflow/internal/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,183 @@
...
+
+	ok := true
+	for _, depID := range task.Dependencies() {
+		depStatus, err := d.get(depID)
+		if err != nil {
+			return false, fmt.Errorf("failed to get status for task with ID: %s", id)
+		}
+		if !depStatus.Finished() {
+			ok = false
```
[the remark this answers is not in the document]
chrisgacsal: Agreed.

workflow/workflow.go
chrisgacsal: I have a couple of ideas how this can be improved and I do plan to make the current implementation more robust, but I am not sure if the `dispatcher`'s functionality would go away. The `workflow` should not store/handle the "state" of any `workflow.Run` invocation. We need to handle that state separately.

Also I am not sure if using only channels to trigger invocations would actually make the logic more straightforward, but as I mentioned I do plan to give a try to something more event driven with less orchestration in place. But definitely not in this PR as to much work depends on this and I wanted to make sure that the current solution is stable enough to be used in other components.

ramizpolic (approved):
lgtm, great work @chrisgacsal :shipit:

PR #1394, opened 2024-03-03
fix: subst spaces with commas in GO_BUILD_TAGS

akijakya:
## Description

Create a comma-separated list instead of a space-separated list from `GO_BUILD_TAGS` in the Makefile. The previous solution worked for go-related make targets (build, test, vet stc.) but was not sufficient for the `make docker*` targets.

## Type of Change

[X] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,9 @@ build: ui build-all-go ## Build all components
 build-all-go: bin/vmclarity-apiserver bin/vmclarity-cli bin/vmclarity-orchestrator bin/vmclarity-ui-backend bin/vmclarity-cr-discovery-server ## Build all go components
 
 BUILD_OPTS = -race
-BUILD_OPTS += -tags="$(GO_BUILD_TAGS)"
+ifneq ($(strip $(GO_BUILD_TAGS)),)
+	BUILD_OPTS += -tags=$(call subst-space-with-comma,$(GO_BUILD_TAGS))
+endif
 
 LDFLAGS = -s -w
 LDFLAGS += -X 'github.com/openclarity/vmclarity/core/version.Version=$(VERSION)'
--- a/makefile.d/10-helpers.mk
+++ b/makefile.d/10-helpers.mk
@@ -33,3 +33,12 @@ else
 		ARCHTYPE = arm64
 	endif
 endif
+
+####
+## Helper to create a comma-separated list from a space-separated list
+####
+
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -62,7 +62,7 @@ build: ui build-all-go ## Build all components
 build-all-go: bin/vmclarity-apiserver bin/vmclarity-cli bin/vmclarity-orchestrator bin/vmclarity-ui-backend bin/vmclarity-cr-discovery-server ## Build all go components
 
 BUILD_OPTS = -race
-BUILD_OPTS += -tags="$(GO_BUILD_TAGS)"
+BUILD_OPTS += -tags=$(call subst-space-with-comma,$(GO_BUILD_TAGS))
```
chrisgacsal, line 65:
```suggestion
ifneq ($(strip $(GO_BUILD_TAGS)),)
	BUILD_OPTS += -tags=$(call subst-space-with-comma,$(GO_BUILD_TAGS))
endif
```

To make sure we only append go build tags if they are set.

Makefile
```diff
@@ -494,3 +494,9 @@ endif
 
 $(DIST_DIR)/CHANGELOG.md: $(ROOT_DIR)/cliff.toml bin/git-cliff | $(DIST_DIR)
 	$(GITCLIFF_BIN) --config $(ROOT_DIR)/cliff.toml --output $@ $(GITCLIFF_OPTS)
+
+# helper to create a comma-separated list from a space-separated list
+null  :=
+space := $(null) #
+comma := ,
+subst-space-with-comma = $(subst $(space),$(comma),$(strip $(1)))
```
chrisgacsal, line 502: I'd move this part to `makefiles.d/10-helpers.mk` to avoid polluting the `Makefile`.

chrisgacsal (changes requested):

Makefile
akijakya: Thanks, I didn't notice we have this file 👍

Makefile
akijakya: Right now we always have two, but I'll add it for extra safety in case those will be removed in the future

chrisgacsal (approved):

PR #1456, opened 2024-03-14
test(aws): implement aws test environment

paralta:
## Description

Implement AWS test environment
* Create new environment for AWS
* Environment setup consists in creating a Cloudformation stack from the available template at https://github.com/openclarity/vmclarity/blob/main/installation/aws/VmClarity.cfn and creating an asset to be scanned (simple EC2 instance)
* Connection with the VMClarity server via an SSH tunnel using local existing keys (as environment variables)
* Make E2E tests provider agnostic (until now the only provider was Docker) and increase setup+scan timeouts
* All tests are running successfully 🥳 

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


testenv/aws/config.go
```diff
@@ -0,0 +1,75 @@
...
+package aws
+
+import (
+	"context"
+	"errors"
+
+	envtypes "github.com/openclarity/vmclarity/testenv/types"
+)
+
+const DefaultAWSRegion = "eu-central-1"
```
akijakya, line 25: Is this being used somewhere? I see a validation to force providing this env, but wouldn't it be nicer if there were a default region instead of throwing an error?

paralta: nice catch! I remember I used it initially but then deprecated its use..

Not sure what would be the best practice here, is it is ok for us to force a certain location if the user does not specify it? I think the worst case scenario here would be if the user as all the resources in a certain region and then the stack is created in another region.

akijakya: This test runs on its own bubble meaning it also creates the asset VM it scans so I don't think it matters (or it shouldn't matter) what other resources are in the region besides the ones created by the e2e test, right?

paralta: that is true, I will add a default region 👍 thanks for reviewing!

testenv/aws/utils.go
```diff
@@ -0,0 +1,258 @@
...
+		CreateBucketConfiguration: &s3types.CreateBucketConfiguration{
+			LocationConstraint: s3types.BucketLocationConstraint(e.region),
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create bucket: %w", err)
+	}
+
+	// Read template file
+	f, err := os.Open("../installation/aws/VmClarity.cfn")
```
chrisgacsal, line 73: Please use the `installation.AWSManifestBundle` for the CFN tempate here. The same way as the docker-compose and helm chart based deployments does.

chrisgacsal (changes requested):

testenv/utils/ssh.go
```diff
@@ -0,0 +1,171 @@
...
+	"path/filepath"
+
+	"golang.org/x/crypto/ssh"
+)
+
+type SSHKeyPair struct {
+	PublicKeyFile  string
+	PrivateKeyFile string
+	Temporary      bool
+}
```
chrisgacsal, line 37:
```suggestion
type SSHKeyPair struct {
	PublicKey  []byte
	PrivateKey []byte
	Temporary bool
}

func (k *SSHKeyPair) Export(privKeyFile, pubKeyFile string) error {}
func (k *SSHKeyPair) Load(privKeyFile, pubKeyFile string) error {}
```
I'd make the `SSHKeyPair` simple container which supports key-pairs to be loaded from/exported to the filesystem.

testenv/utils/docker_ssh.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/docker/cli/cli/connhelper"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
+)
+
+func GetRemoteDockerContainerList(ctx context.Context, privateKeyFile, remoteHost string) (*[]types.Container, error) {
```
chrisgacsal, line 29: Let's talk about this offline.

chrisgacsal (changes requested):

e2e/helpers.go
```diff
@@ -16,18 +16,42 @@
...
+	scope := "assetInfo/%s/any(t: t/key eq 'scanconfig' and t/value eq 'test')"
+	switch cfg.TestEnvConfig.Platform {
+	case types.EnvironmentTypeAWS, types.EnvironmentTypeGCP, types.EnvironmentTypeAzure:
+		return fmt.Sprintf(scope, "tags")
+	case types.EnvironmentTypeDocker, types.EnvironmentTypeKubernetes:
+		return fmt.Sprintf(scope, "labels")
+	default:
+		return ""
+	}
+}
```
chrisgacsal, line 42:
```suggestion
func getDefaultScope(t types.EnvironmentType) string {
	scope := "assetInfo/%s/any(t: t/key eq 'scanconfig' and t/value eq 'test')"
	switch t {
	case types.EnvironmentTypeAWS, types.EnvironmentTypeGCP, types.EnvironmentTypeAzure:
		return fmt.Sprintf(scope, "tags")
	case types.EnvironmentTypeDocker, types.EnvironmentTypeKubernetes:
		return fmt.Sprintf(scope, "labels")
	default:
		return ""
	}
}
```
There is not need to pass the whole `Config` struct when only a single parameter is needed.

e2e/helpers.go
```diff
@@ -16,18 +16,42 @@
...
+func getDefaultScanTimeout(cfg *config.Config) time.Duration {
+	switch cfg.TestEnvConfig.Platform {
+	case types.EnvironmentTypeAWS, types.EnvironmentTypeGCP, types.EnvironmentTypeAzure:
+		return 20 * time.Minute
+	case types.EnvironmentTypeDocker, types.EnvironmentTypeKubernetes:
+		return 2 * time.Minute
+	default:
+		return time.Minute
+	}
+}
```
chrisgacsal, line 54:
```suggestion
func getDefaultScanTimeout((t types.EnvironmentType) time.Duration {
	switch t {
	case types.EnvironmentTypeAWS, types.EnvironmentTypeGCP, types.EnvironmentTypeAzure:
		return 20 * time.Minute
	case types.EnvironmentTypeDocker, types.EnvironmentTypeKubernetes:
		return 2 * time.Minute
	default:
		return time.Minute
	}
}
```

Same as above.

e2e/suite_test.go
```diff
@@ -76,12 +76,24 @@ func beforeSuite(ctx context.Context) {
...
+	switch cfg.TestEnvConfig.Platform {
+	case types.EnvironmentTypeDocker:
+		servicesReadyTimeout = 5 * time.Second
+	case types.EnvironmentTypeAWS:
+		servicesReadyTimeout = 10 * time.Minute
+	case types.EnvironmentTypeGCP, types.EnvironmentTypeAzure, types.EnvironmentTypeKubernetes:
+	default:
+		servicesReadyTimeout = time.Minute
+	}
+
```
chrisgacsal, line 90: Shall this be a helper function the same way as the `getDefaultScanTimeout` and `getDefaultScanTimeout`?

Probably I'd combine these 3 parameter into a single struct generated by a single helper function based on the `types.EnvironmentType`.

```go
type TestSuiteParams struct {
	Scope string
	ServiceReadynessTimeout time.Duration
	ScanTimeout time.Duration
}

func TestSuiteParamsForEnv(t types.EnvironmentType) *TestSuiteParams {
	p := &TestSuiteParams{}
	scope := "assetInfo/%s/any(t: t/key eq 'scanconfig' and t/value eq 'test')"

	switch t {
	case types.EnvironmentTypeAWS, types.EnvironmentTypeGCP, types.EnvironmentTypeAzure:
		...
	case types.EnvironmentTypeDocker, types.EnvironmentTypeKubernetes:
		...
	default:
		...
	}

	return p
}
```

chrisgacsal (changes requested):

testenv/utils/ssh.go
```diff
@@ -0,0 +1,171 @@
...
+	// Listen on local port.
+	listener, err := net.Listen("tcp", "localhost:"+localPort)
+	if err != nil {
+		logger.Fatalf("failed to listen: %s\n", err)
+	}
+	defer listener.Close()
+
+	for {
+		// Accept local connection.
+		local, err := listener.Accept()
```
chrisgacsal, line 130:
```suggestion
		local, err := listener.Accept()
		defer local.Close()
```

testenv/utils/ssh.go
```diff
@@ -0,0 +1,171 @@
...
+		if err != nil {
+			logger.Fatalf("failed to accept: %s\n", err)
+		}
+
+		// Dial remote server.
+		remote, err := client.Dial("tcp", "localhost:"+remotePort)
```
chrisgacsal, line 136:
```suggestion
		remote, err := client.Dial("tcp", "localhost:"+remotePort)
		defer remote.Close()
```

testenv/utils/ssh.go
```diff
@@ -0,0 +1,171 @@
...
+		runTunnel(ctx, local, remote)
+	}
+}
+
+// runTunnel runs a tunnel between two connections.
+func runTunnel(ctx context.Context, local, remote net.Conn) {
+	logger := GetLoggerFromContextOrDiscard(ctx)
+
+	defer local.Close()
+	defer remote.Close()
```
chrisgacsal, line 151:
```suggestion
```

chrisgacsal: If some reason the `runTunnel` is not called the connections will be leaked.

testenv/utils/ssh.go
```diff
@@ -0,0 +1,171 @@
...
+	done := make(chan struct{}, 2) //nolint:gomnd
```
chrisgacsal, line 152: I'd use a `sync.WaitGroup` here instead of the done channel.

testenv/aws/aws.go
```diff
@@ -0,0 +1,242 @@
...
+	s3Client            *s3.Client
+	testAsset           *asset.Asset
+	server              *Server
+	stackName           string
+	templateURL         string
+	workDir             string
+	region              string
+	sshKeyPair          *utils.SSHKeyPair
+	sshPortForwardInput *utils.SSHForwardInput
+	meta                map[string]interface{}
```
akijakya, line 61: Is this field necessary? :)

testenv/aws/utils.go
```diff
@@ -0,0 +1,290 @@
...
+		return errors.New("infrastructure is not ready")
+	}
+
+	if err = e.server.WaitForSSH(ctx, DefaultSSHPortReadyTimeout); err != nil {
+		return fmt.Errorf("failed to wait for the SSH port to become ready: %w", err)
+	}
+
+	e.sshPortForwardInput = &utils.SSHForwardInput{
+		PrivateKey:    e.sshKeyPair.PrivateKey,
+		User:          "ubuntu",
```
akijakya, line 93: nit: this could be a const as it is being used someplace else as well, right?

akijakya (dismissed):
LGTM, thanks for pioneering in this testenv topic! 🙏

testenv/aws/aws.go
paralta: @chrisgacsal I think I remember you mentioned that this field was required, can you remind me where?

testenv/aws/utils.go
paralta: thanks for reviewing, @akijakya ! good point 👍

e2e/suite_test.go
ramizpolic: I like this idea too, have all the env-specific configs flow through a shared struct

testenv/aws/aws.go
```diff
@@ -0,0 +1,243 @@
...
+	InstanceID string
+	PublicIP   string
+}
+
+func (s Server) WaitForSSH(ctx context.Context, t time.Duration) error {
+	// TODO(paralta): Implement method to check if SSH port is ready
+	time.Sleep(t) // nolint:gomnd
+
+	return nil
+}
```
ramizpolic, line 77: potential way (needs to be tested)
```suggestion
func (s Server) WaitForSSH(ctx context.Context, timeout time.Duration) error {
	conn, err := net.DialTimeout("tcp", net.JoinHostPort(host, port), timeout) // requires host and port to be available
	if err != nil {
	    return fmt.Errorf("connection error: %w", err)
	}
	defer conn.Close()
	
	return nil // ok
}
```

testenv/aws/aws.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// Create a new CloudFormation stack from template
+	_, err = e.client.CreateStack(
+		ctx,
+		&cloudformation.CreateStackInput{
+			StackName:    &e.stackName,
+			Capabilities: []cloudformationtypes.Capability{cloudformationtypes.CapabilityCapabilityIam},
+			TemplateURL:  &e.templateURL,
+			Parameters: []cloudformationtypes.Parameter{
+				{ParameterKey: aws.String("KeyName"), ParameterValue: &e.stackName},
```
ramizpolic, line 99: multiline split for clarity
```suggestion
				{
					ParameterKey: aws.String("KeyName"), 
					ParameterValue: &e.stackName,
				},
```

testenv/aws/aws.go
```diff
@@ -0,0 +1,243 @@
...
+		return fmt.Errorf("failed to create stack: %w", err)
+	}
+
+	waiter := cloudformation.NewStackCreateCompleteWaiter(e.client)
+	descStackInput := &cloudformation.DescribeStacksInput{
+		StackName: &e.stackName,
+	}
+	if err = waiter.Wait(ctx, descStackInput, DefaultStackCreationTimeout); err != nil {
+		return fmt.Errorf("failed to wait for the stack to be created: %w", err)
+	}
```
ramizpolic, line 113:
```suggestion
	if err = waiter.Wait(ctx, &cloudformation.DescribeStacksInput{
		StackName: &e.stackName,
	}, DefaultStackCreationTimeout); err != nil {
		return fmt.Errorf("failed to wait for the stack to be created: %w", err)
	}
```

testenv/aws/utils.go
```diff
@@ -0,0 +1,291 @@
...
+		RemotePort:    80, //nolint:gomnd
+	}
+
+	sshPF, err := utils.NewSSHPortForward(e.sshPortForwardInput)
+	if err != nil {
+		return fmt.Errorf("failed to setup SSH port forwarding: %w", err)
+	}
+
+	// Use non-inherited context to avoid cancelling the port forward with timeout
+	if err = sshPF.Start(context.Background()); err != nil { //nolint:contextcheck
```
ramizpolic, line 107: if this is the case, port forward will remain running? should we have a context ontop of the AWSEnv passed from testenv to ensure that port forwarding stops after the tests have been executed?

testenv/aws/utils.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	if status != ec2types.InstanceStateNameRunning {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+// Get EC2 instance status by instance ID.
+func (e *AWSEnv) getEC2InstanceStatus(ctx context.Context, instanceID string) (ec2types.InstanceStateName, error) {
```
ramizpolic, line 252: if this is only used to check if the instance is ready, maybe this could be changed into `isEC2InstanceReady`

testenv/utils/ssh.go
```diff
@@ -0,0 +1,291 @@
...
+		return fmt.Errorf("failed to listen: %w", err)
+	}
+
+	// TODO: wait until the BannerCallBack is invoked
+
+	go func() {
+		var callbackErr error
+		for callbackErr == nil && !f.stop.Load() {
+			defer client.Close()
+			defer listener.Close()
```
ramizpolic, line 169: shouldnt this be outside of the loop?

testenv/utils/ssh.go
```diff
@@ -0,0 +1,291 @@
...
+	if _, err := os.Stat(privateKeyFile); errors.Is(err, os.ErrNotExist) {
+		if err := keys.Save(privateKeyFile, publicKeyFile); err != nil {
+			return fmt.Errorf("failed to save SSH keys to filesystem: %w", err)
+		}
+	}
+
+	dst := input.User + "@" + input.Host
+	journalctlCmd := "journalctl -u " + input.Service + ".service --since='" + startTime.Format("2006-01-02 15:04:05") + "'"
+
+	args := []string{dst, "-i", privateKeyFile, journalctlCmd}
```
ramizpolic, line 281: can we move all the arguments into args slice for clarity?

ramizpolic (changes requested):

testenv/aws/utils.go
ramizpolic: another note: ssh port forwarding is never stopped (either explicitly via stop, or via context), which is needed as otherwise the port will stay in use

testenv/utils/ssh.go
```diff
@@ -0,0 +1,291 @@
...
+		return nil, fmt.Errorf("failed to parse private key: %w", err)
+	}
+
+	// Create SSH client config.
+	config := ssh.ClientConfig{
+		User: input.User,
+		Auth: []ssh.AuthMethod{
+			ssh.PublicKeys(signer),
+		},
+		HostKeyCallback:   ssh.InsecureIgnoreHostKey(), //nolint:gosec
```
ramizpolic, line 245: Did you try verifying host key?
```suggestion
		HostKeyCallback:   ssh.FixedHostKey(signer.PublicKey()),
```

e2e/basic_scan_test.go
```diff
@@ -24,43 +24,50 @@ import (
...
 	ginkgo.Context("which scans a docker image", func() {
 		ginkgo.It("should finish successfully", func(ctx ginkgo.SpecContext) {
+			if cfg.TestEnvConfig.Platform != types.EnvironmentTypeDocker {
+				ginkgo.By("skipping test because it's not running on docker")
+				return
+			}
+
+			containerInfo, err := (*assets.Items)[0].AssetInfo.AsContainerInfo()
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			imageID = containerInfo.Image.ImageID
```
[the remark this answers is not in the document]
paralta: this test case only runs for the docker provider that is why this logic is contained in this section after the provider check

testenv/utils/ssh.go
paralta: this change fails with `failed to run after setup: failed to wait for the SSH port to become ready: failed to dial:ssh: handshake failed: ssh: host key mismatch`

testenv/utils/ssh.go
```diff
@@ -0,0 +1,291 @@
...
+
+			wg.Wait()
+		}
+	}()
+
+	return nil
+}
+
+func (f *SSHPortForward) Stop() {
+	f.stop.Store(true)
```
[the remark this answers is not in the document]
ramizpolic: resolved

testenv/utils/ssh.go
ramizpolic: addressed below in review

testenv/aws/aws.go
ramizpolic: in other envs, meta fields are used for logging, nothing more

testenv/utils/ssh.go
```diff
@@ -0,0 +1,299 @@
...
+		defer client.Close()
+		defer listener.Close()
+
+		for callbackErr == nil && !f.stop.Load() {
+			func() {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+				}
```
ramizpolic, line 177: this would need to be outside of the function (inside the loop), otherwise we might have just the loop running without returning. 
```go
for condition {
	select {} // allows it to return if the ctx is done
	func() { } // executes the logic to stream between ssh and host
}

```

paralta: fixed 👍 thank you so much for reviewing this!

ramizpolic (approved):
lgtm, great work @paralta! 🚀

chrisgacsal (approved):
Thanks for this! Awesome work! :shipit:

PR #1601, opened 2024-04-19
feat: add gcp e2e test env

adamtagscherer:
## Description

This PR adds the GCP E2E test environment to the codebase.

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/installation/gcp/dm/components/roles.py
+++ b/installation/gcp/dm/components/roles.py
@@ -21,10 +21,13 @@ from hashlib import sha1
 def GenerateConfig(context):
     """Creates the custom roles"""
 
-    discoverer_role_name = f"vmclarity-{context.env['deployment']}-discoverer-snapshotter"
+    deployment_name_hash = sha1(context.env["deployment"].encode("utf8")).hexdigest()[:10]
+    prefix = f"vmclarity-{deployment_name_hash}"
+
+    discoverer_role_name = f"{prefix}-discoverer-snapshotter"
     discoverer_role_id = discoverer_role_name.replace("-", "_")
 
-    scanner_role_name = f"vmclarity-{context.env['deployment']}-scanner"
+    scanner_role_name = f"{prefix}-scanner"
     scanner_role_id = scanner_role_name.replace("-", "_")
 
     resources = [
@@ -95,7 +98,7 @@ def GenerateConfig(context):
             },
         },
         {
-            "name": f"vmclarity-{context.env['deployment']}-discoverer-snapshotter-role-binding",
+            "name": f"{prefix}-discoverer-snapshotter-role-binding",
             "type": "gcp-types/cloudresourcemanager-v1:virtual.projects.iamMemberBinding",
             "properties": {
                 "resource": context.env["project"],
@@ -107,7 +110,7 @@ def GenerateConfig(context):
             },
         },
         {
-            "name": f"vmclarity-{context.env['deployment']}-scanner-role-binding",
+            "name": f"{prefix}-scanner-role-binding",
             "type": "gcp-types/cloudresourcemanager-v1:virtual.projects.iamMemberBinding",
             "properties": {
                 "resource": context.env["project"],
--- a/installation/gcp/dm/components/roles.py
+++ b/installation/gcp/dm/components/roles.py
@@ -21,13 +21,10 @@ from hashlib import sha1
 def GenerateConfig(context):
     """Creates the custom roles"""
 
-    deployment_name_hash = sha1(context.env["deployment"].encode("utf8")).hexdigest()[:10]
-    prefix = f"vmclarity-{deployment_name_hash}"
-
-    discoverer_role_name = f"{prefix}-discoverer-snapshotter"
+    discoverer_role_name = f"vmclarity-{context.env['deployment']}-discoverer-snapshotter"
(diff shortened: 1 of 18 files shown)


installation/gcp/dm/components/roles.py
```diff
@@ -21,13 +21,10 @@
 def GenerateConfig(context):
     """Creates the custom roles"""
 
-    deployment_name_hash = sha1(context.env["deployment"].encode("utf8")).hexdigest()[:10]
```
ramizpolic, line 24: im not familiar fully with why we use hash here, but will this change impact any logic (name hitting the size limit)?

testenv/gcp/asset/asset.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/openclarity/vmclarity/core/to"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	"cloud.google.com/go/compute/apiv1/computepb"
+)
+
+func Create(ctx context.Context, instancesClient *compute.InstancesClient, projectID string, zone string, assetName string) error {
+	var diskSizeGb int64 = 10
+
+	req := &computepb.InsertInstanceRequest{
```
ramizpolic, line 30: lets declare values used for instance creation as consts on top level (machineType, sourceImage, networkInterfaceName). do we need to use e2-standard-2 image for testing? can one of micro instances do the job?

testenv/utils/docker/docker.go
```diff
@@ -82,6 +83,27 @@ func (e *DockerHelper) Services(ctx context.Context) (envtypes.Services, error)
 	return serviceCollection.AsServices(), nil
 }
 
+func (e *DockerHelper) WaitForDockerReady(ctx context.Context) error {
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
ramizpolic, line 87: its okay to redeclare ctx here (reducing to 2 minutes should also be okay)
```suggestion
	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
```

testenv/utils/ssh.go
```diff
@@ -225,6 +225,25 @@ func (f *SSHPortForward) Stop() {
 	f.cancel()
 }
 
+func (f *SSHPortForward) DialWithTimeout() (*ssh.Client, error) {
```
ramizpolic, line 228: we should handle context cancellation from parent here to avoid waiting full timeout
```suggestion
func (f *SSHPortForward) DialWithTimeout(ctx context.Context) (*ssh.Client, error) {
```

ramizpolic (changes requested):

testenv/gcp/gcp.go
```diff
@@ -0,0 +1,353 @@
...
+		}
+	} else {
+		sshKeyPair, err = utils.GenerateSSHKeyPair()
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate ssh key pair: %w", err)
+		}
+	}
+
+	return &GCPEnv{
+		workDir:         config.WorkDir,
```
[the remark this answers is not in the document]
adamtagscherer: Currently everything in GCP gets the context from the Config, I don't see how including the `Config.ctx` here would help us making it more robust.

testenv/utils/docker/docker.go
adamtagscherer: 2 minutes is not enough for GCP :S

testenv/gcp/utils.go
```diff
@@ -0,0 +1,70 @@
...
+		RemoteAddress: "localhost",
+		RemotePort:    80, //nolint:gomnd
+	}
+
+	e.sshPortForward, err = utils.NewSSHPortForward(e.sshPortForwardInput)
+	if err != nil {
+		return fmt.Errorf("failed to setup SSH port forwarding: %w", err)
+	}
+
+	if err = e.sshPortForward.Start(context.Background()); err != nil { //nolint:contextcheck
```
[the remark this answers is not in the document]
adamtagscherer: Okay so, we have a problem with passing contexts to these functions from the parent.

The issue why we can't use the `ctx` variable here from the `afterSetup` function's parameter is because if you follow up the call chain then that `ctx` is the `setupCtx` variable in the `beforeSuit` function, see here: https://github.com/openclarity/vmclarity/blob/main/e2e/suite_test.go#L68

That context ends when the `beforeSuit` function runs to completion and after that the context is not available, but with the current solution we want to use that context in actual tests (because we want to have an open SSH connection to VMClarity control plane while the tests are running).

This is not going to work, we have to rework how we pass context from ginkgo's test suite creation to the actual test environments, but I think that's out of the scope of this PR.

The situation is the same here in your previous comment - https://github.com/openclarity/vmclarity/pull/1601/files/de050880ac466a03a92bbf25a6497d4bde060beb#diff-7be911ff144b8813d4374c3f85695689dd505e9ace576d90e7f0f2b7a071b421R327 - the context that we pass to the `iamService` is cancelled when the `beforeSuit` ends, but we want to use this context in the `TearDown` where we want to undelete the 2 roles.

ramizpolic (approved):
lgtm. thanks @adamtagscherer 🚀

PR #1679, opened 2024-05-15
feat: add dependency on healthchecks to docker-compose services

### LINKED ISSUE #1612: Installing VMClarity with Postgresql database fails on cloud providers
## Description

Deployment of VMClarity on Azure fails if the database to use is set to `Postgresql` instead of `SQLite` on the step when the deploy script is being run on the server because the `PostgreSQL` container starts slower than the `apiserver` container, which tries to connect to the database but fails and restarts a couple of times, causing the `vmclarity.service` to fail and thus ending up failing the whole installation.

This probably affects the install script of other cloud providers as well.

## Expected Behavior

The docker compose waits with spinning up the `apiserver` container until the `PostgreSQL` container is healthy.

## Checklist

- [X] I have verified this does not duplicate an existing issue

akijakya:
## Description

- switch from using compose profiles to docker-compose override file so an additional healthcheck could also be added to the `apiserver` service when the stack is used with PostgreSQL database instead of SQLite. This is necessary to avoid service failures coming from the apiserver because of slow postgres startup, failing the whole VMClarity installation on cloud providers.
- add startup dependency on healthcheck to `orchestrator`, `UI` and `uibackend` services which depend on the `apiserver` service
- remove obsolete version from docker-compose files

Tested with Docker, Azure and AWS providers.

Fixes #1612 

## Type of Change

[X] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


paralta (dismissed):
LGTM 🚀

ramizpolic (approved):
LGTM!

paralta (approved):

PR #1682, opened 2024-05-16
ci: run k8s tests

paralta:
## Description

* Add new `make e2e-docker` target and use `make e2e` to run end to end tests on both Docker and Kubernetes
* Run end to end tests running on Kubernetes in the CI pipeline. This should not increase the CI pipeline run time significantly since Kubernetes tests (~15min) and Docker tests (~14min) will run in parallel.
* Required steps for merge in this repository need to be updated

## Type of Change

[ ] Bug Fix  
[x] New Feature  
[ ] Breaking Change  
[ ] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist

### DIFF
--- a/.github/workflows/reusable-end-to-end-testing.yml
+++ b/.github/workflows/reusable-end-to-end-testing.yml
@@ -111,4 +111,4 @@ jobs:
           VMCLARITY_E2E_CR_DISCOVERY_SERVER_IMAGE: ${{ format('{0}:{1}', needs.images.outputs.cr-discovery-server-image, inputs.image_tag) }}
           VMCLARITY_E2E_SCANNER_PLUGIN_KICS: ${{ format('{0}:{1}', needs.images.outputs.plugin-kics-image, inputs.image_tag) }}
         run: |
-          make e2e
+          make e2e-docker
--- a/.github/workflows/reusable-end-to-end-testing.yml
+++ b/.github/workflows/reusable-end-to-end-testing.yml
@@ -17,6 +17,10 @@ on:
         type: boolean
         description: 'If set to true the image published to the release repository is used otherwise the development.'
         default: false
+      platform:
+        required: true
+        type: string
+        description: 'Platform used to run end-to-end tests. Supported values are `docker` and `kubernetes`.'
 
 jobs:
   images:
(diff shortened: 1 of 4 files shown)


testenv/kubernetes/kind/dockerimageloader.go
```diff
@@ -52,7 +52,27 @@ func (l *DockerImageLoader) imageIDsFromRepoTags(ctx context.Context, repoTags [
...
+				Filters: filters.NewArgs(
+					filters.Arg("reference", repoTag)),
+			})
+			if err != nil {
+				return nil, fmt.Errorf("failed to list image %s: %w", repoTag, err)
+			}
+
+			if len(result) <= 0 {
+				return nil, fmt.Errorf("failed to find image: %s", repoTag)
+			}
```
ramizpolic, line 75: we can continue here meaning that `imageIDsFromRepoTags` can return empty slice. we can move the responsibility to pull image to `Load` method from line 100

```go
		ids, err := l.imageIDsFromRepoTags(ctx, []string{image})
		if err != nil {
			return fmt.Errorf("failed to get ImageID for RepoTag %s: %w", image, err)
		}
		if len(ids) == 0 {
			// pull logic
			// ids = from pull info
		}
```

ramizpolic: lets also log when we pull the image

ramizpolic (approved):

.github/workflows/reusable-end-to-end-testing.yml
```diff
@@ -110,5 +124,12 @@ jobs:
           VMCLARITY_E2E_SCANNER_IMAGE: ${{ format('{0}:{1}', needs.images.outputs.cli-image, inputs.image_tag) }}
           VMCLARITY_E2E_CR_DISCOVERY_SERVER_IMAGE: ${{ format('{0}:{1}', needs.images.outputs.cr-discovery-server-image, inputs.image_tag) }}
           VMCLARITY_E2E_SCANNER_PLUGIN_KICS: ${{ format('{0}:{1}', needs.images.outputs.plugin-kics-image, inputs.image_tag) }}
+          VMCLARITY_E2E_PLATFORM: ${{ inputs.platform }}
         run: |
```
akijakya, line 128: I know I am a bit late but it seems like duplication to me, we don't need two separate make targets if the only difference between them is the value of the `VMCLARITY_E2E_PLATFORM`, which we already set here in the CI, what do you think?

paralta: but if you mean here, we need to reuse this workflow for each one in order to run the two environments in parallel

akijakya: What I propose is to have only one `make e2e` make target because the difference between running it with docker vs kubernetes provider is just the value of the `VMCLARITY_E2E_PLATFORM` env var. So the `run` command here in the github action can also remain `make e2e` since this env is being set for this step.

akijakya: Alternatively, in the case we stick to having two different make targets for the convenience of running both docker and kubernetes at once with `make e2e`, then setting the `VMCLARITY_E2E_PLATFORM` env here in this GHA step is redundant since the make targets will start the test with the correct platform anyways :)

PR #1741, opened 2024-05-27
refactor(scanner): propagate context to scanners/analyzers

zsoltkacsandi:
## Description

Passing context to scanners/analyzers.

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[X] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


paralta (dismissed):
LGTM 🚀

scanner/families/exploits/exploitdb/exploitdb.go
```diff
@@ -45,8 +46,8 @@ func New(_ string, c job_manager.IsConfig, logger *log.Entry, resultChan chan jo
 	}
 }
 
-func (a *Scanner) Run(sourceType utils.SourceType, userInput string) error {
-	go func() {
+func (a *Scanner) Run(ctx context.Context, sourceType utils.SourceType, userInput string) error {
+	go func(_ context.Context) {
```
ramizpolic, line 50: although we use context in the goroutine body, the ctx from the gororutine itself is not used. we should use one approach to ensure consistency by either using goroutine ctx or using parent ctx outside of goroutine 
```go
 Run(ctx context.Context, sourceType utils.SourceType, userInput string) error {
	go func() {
		someOperation(ctx) // 1. ok, makes sense where ctx is coming from
	}()

	go func(ctx context.Context) {
		someOperation(ctx) // 2. ok, makes sense where ctx is coming from
	}(ctx)

	go func(_ context.Cotnext) {
		someOperation(ctx) // 3. confusing since ctx is available locally but being used from parent
	}(ctx)
}
```
We should also decide for one of these approaches (either 1 or 2) and use it across scanner logic to avoid confusion and enforce consistency.

ramizpolic (approved):
LGTM! 🚀

ramizpolic (dismissed):

ramizpolic:
Note: CIS Docker Benchmark scanner needs to be updated upstream to support contexts being passed when used as a library. It should either be added as part of Config or directly in calling function at https://github.com/Portshift/dockle/blob/4f3b165086b7adbaf853ece809c7c61497bb52d7/pkg/run.go#L44. Without this, we cannot force exit for CIS Docker scanner. This can be done in another PR once there is support for passing ctx in Dockle.

ramizpolic (approved):

ramizpolic (dismissed):

akijakya:
> Note: CIS Docker Benchmark scanner needs to be updated upstream to support contexts being passed when used as a library. It should either be added as part of Config or directly in calling function at https://github.com/Portshift/dockle/blob/4f3b165086b7adbaf853ece809c7c61497bb52d7/pkg/run.go#L44. Without this, we cannot force exit for CIS Docker scanner. This can be done in another PR once there is support for passing ctx in Dockle.

Wouldn't that be a breaking change for the Dockle scanner? Time for our own fork? 😃 A ticket should be created for this anyway if there isn't one already...

ramizpolic:
> > Note: CIS Docker Benchmark scanner needs to be updated upstream to support contexts being passed when used as a library. It should either be added as part of Config or directly in calling function at https://github.com/Portshift/dockle/blob/4f3b165086b7adbaf853ece809c7c61497bb52d7/pkg/run.go#L44. Without this, we cannot force exit for CIS Docker scanner. This can be done in another PR once there is support for passing ctx in Dockle.
> 
> 
> 
> Wouldn't that be a breaking change for the Dockle scanner? Time for our own fork? 😃 A ticket should be created for this anyway if there isn't one already...

It wont be a breaking change for Dockle. @zsoltkacsandi has a way to address this in a backwards compatible way by making the run function that performs the scan in Dockle public.

ramizpolic (approved):

PR #1774, opened 2024-06-10
feat: enhance yara scanner

adamtagscherer:
## Description

This PR revamps the Yara malware scanner to run under a reasonable amount of time.
The tests were run with an Ubuntu 24.04 image on AWS.

### Scanning the whole mounted filesystem:

Yara config on the scanner VM:
Screenshot 2024-06-13 at 11 32 15

Scan still in progress after 1 hour, not sure when it will finish:
Screenshot 2024-06-11 at 17 41 52

### Scanning just specific files:

Malware config on scanner VM:
Screenshot 2024-06-13 at 10 51 13

Malware finished in ~16mins:
Screenshot 2024-06-13 at 11 02 23

## Type of Change

[ ] Bug Fix  
[ ] New Feature  
[ ] Breaking Change  
[x] Refactor  
[ ] Documentation  
[ ] Other (please describe)  

## Checklist


paralta (dismissed):
LGTM 👍 just one small comment

scanner/families/malware/yara/yara.go
```diff
@@ -91,7 +94,7 @@ func (s *Scanner) Run(sourceType utils.SourceType, userInput string) error {
 		defer cleanup()
 
 		// Define the yara args to run
-		args := []string{"-C", s.compiledRuleFile, "-r", "-w", "-m", fsPath}
```
ramizpolic, line 94: Couple of suggestions here:
- We can fully remove the _per file timeout_ field and use the context passed to `Run` to control the cancellation/timeout
- We dont need to scan each file independently but switch it to per-directory scan
- We actually dont need to parallelise the execution on our own but rather use `--threads=runtime.NumCPU()` to pass this responsibility to Yara since its supported on them
- Instead of running multiple processes parallely to scan each dir, we can do this sequentially since the parallelisation is supported on Yara
- Scanning a directory by Yara should use recursive flag `-r` to ensure that the whole directory depth is scanned

scanner/families/malware/yara/yara.go
```diff
@@ -130,12 +133,38 @@ func (s *Scanner) Run(sourceType utils.SourceType, userInput string) error {
...
+					err = utils.RunCommandAndParseOutputLineByLine(yaraCommand, parserFunc, errCheckFunc)
+				})
+				if err != nil {
+					s.sendResults(retResults, fmt.Errorf("failed to run yara command: %w", err))
+					return fmt.Errorf("failed to run yara command: %w", err)
+				}
+			}
+
+			return nil
+		})
```
ramizpolic, line 165:
```suggestion
		if !s.needsDirToBeScanned(path, fsPath) {
			return nil
		}

		yaraCommand := exec.CommandContext(ctx, yaraBinaryPath, append(args, path)...)
		err = utils.RunCommandAndParseOutputLineByLine(yaraCommand, parserFunc, errCheckFunc)
		if err != nil {
			s.sendResults(retResults, fmt.Errorf("failed to run yara command: %w", err))
			return fmt.Errorf("failed to run yara command: %w", err)
		}
```

scanner/families/malware/yara/yara.go
```diff
@@ -91,7 +94,7 @@ func (s *Scanner) Run(sourceType utils.SourceType, userInput string) error {
 		defer cleanup()
 
 		// Define the yara args to run
-		args := []string{"-C", s.compiledRuleFile, "-r", "-w", "-m", fsPath}
+		args := []string{"-C", s.compiledRuleFile, "-w", "-m", "-a", strconv.Itoa(s.config.PerFileTimeout)}
```
ramizpolic, line 97:
```suggestion
		args := []string{"-C", s.compiledRuleFile, "-r", "-w", "-m", "--threads", strconv.Itoa(runtime.NumCPU())}
```
this way, we can remove `MaxGoRoutineNumber` since we have the actual runtime value

scanner/families/malware/yara/yara.go
```diff
@@ -130,13 +127,17 @@ func (s *Scanner) Run(ctx context.Context, sourceType utils.SourceType, userInpu
...
-		err = utils.RunCommandAndParseOutputLineByLine(yaraCommand, parserFunc, errCheckFunc)
-		if err != nil {
-			s.sendResults(retResults, fmt.Errorf("failed to run yara command: %w", err))
-			return
+		// Define the yara args to run
+		args := []string{"-C", s.compiledRuleFile, "-r", "-w", "-m", "-p", strconv.Itoa(runtime.NumCPU())}
+		s.logger.Infof("Running yara...")
+
+		for _, d := range s.config.DirectoriesToScan {
+			yaraCommand := exec.CommandContext(ctx, yaraBinaryPath, append(args, fsPath+d)...)
```
ramizpolic, line 135: not really sure why we have the "directories to scan" at all since we can achieve the same behaviour by specifying those dirs through existing scan inputs for malware family. whats the difference between defining things to scan now? could you provide example configs?

adamtagscherer: The `Inputs` that you are referring to I guess - https://github.com/openclarity/vmclarity/blob/main/scanner/families/malware/config.go#L27 ? - is the mount point where the whole filesystem will be mounted on the scanner VM.

This `Inputs` field gets set when the disks get attached to the scanner VM as stated in the comment here: https://github.com/openclarity/vmclarity/blob/main/orchestrator/watcher/assetscan/families.go#L171

So the `DirectoriesToScan` field is needed to narrow down the whole mounted filesystem in the `Inputs` field.

After this PR gets merged you can configure the malware scanner family to only scan these directories when `POST`-ing a new scan config.

```
"scanFamiliesConfig": {
        ...
        },
        "malware": {
          "enabled": true,
          "yara_directories_to_scan": ["/home", "/opt", ...]
        },
        ...
      }
}
```

ramizpolic: thanks @adamtagscherer, makes sense!

ramizpolic (approved):

PR #1794, opened 2024-06-13
feat(plugin): enable binary plugin support

zsoltkacsandi:
## Description

Added top-level `binary_mode` config to the plugin scanner family, and also to the individual plugin scanners, defaulting to `false`. If the top-level `binary_mode` config is explicitly specified, it'll overwrite the individual plugin-level ones.

If binary mode is enabled, it results in running the scanner plugins without the docker daemon, which means downloading the image (without docker pull), extracting its content to the machine, mounting the necessary directories of the host machine under the plugin filesystem and executing its entrypoint with `chroot` (details are in `plugins/runner/internal/runtimehandler/binary/handler.go`).

This solution is not intended for using it in production, but for testing purposes in dockerless environments (e.g. CI/CD pipelines).

## Type of Change

- [ ] Bug Fix  
- [X] New Feature  
- [ ] Breaking Change  
- [X] Refactor  
- [ ] Documentation  
- [ ] Other (please describe)  

## Checklist


plugins/runner/runner.go
```diff
@@ -34,35 +34,36 @@ import (
...
+	runtimeHandler runtimehandler.PluginRuntimeHandler
+	client         runnerclient.ClientWithResponsesInterface
 }
 
 func New(ctx context.Context, config types.PluginConfig) (types.PluginRunner, error) {
 	// Create docker container
 	// TODO: switch to factory once the support for more container engines is added
-	manager, err := docker.New(ctx, config)
+	//handler, err := docker.New(ctx, config)
+	handler, err := binary.New(ctx, config)
```
akijakya, line 46: Note: we need to create handler according to the config (whether binary mode is enabled or not)

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,186 @@
...
+		err = containerrootfs.ToDirectory(ctx, h.config.ImageName, pluginDir)
+		if err != nil {
+			return fmt.Errorf("unable to extract image(%s): %w", h.config.ImageName, err)
+		}
+	}
+
+	// Mount input from host
+	// /tmp/vmclarity-plugins/kics + /input + /host-dir-to-scan
+	inputDirMountPoint := filepath.Join(pluginDir, runtimehandler.RemoteScanInputDirOverride, h.config.InputDir)
+	err = os.MkdirAll(inputDirMountPoint, 0755)
```
ramizpolic, line 67: input dir that needs to be scanned should be mounted in read-only mode

.families.yaml
```diff
@@ -1,139 +1,9 @@
...
-    - input: "/"
-      input_type: "rootfs"
-  scanners_config:
-    chkrootkit:
-      binary_path: chkrootkit
-
 plugins:
-  enabled: false
+  enabled: true
+  binary_mode: true
```
ramizpolic, line 3: we can add `binary_source_dir` which will override the location where the plugins are saved. When empty, it should default to temp dir. This way we can also setup caching to prevent redownloading on every re-run if its needed.

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,171 @@
...
+		return fmt.Errorf("no entrypoint or command found in the config")
+	}
+
+	// Find a port
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return fmt.Errorf("unable to find port")
+	}
+	port := listener.Addr().(*net.TCPAddr).Port
+	listener.Close()
```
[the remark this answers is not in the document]
ramizpolic: we can use a global mutex and lock before finding a port. we can later extract this into utils and create a package to reserve/free up/check available ports.

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,171 @@
...
+func (h *binaryRuntimeHandler) Start(ctx context.Context) error {
+	pluginDir := filepath.Join(os.TempDir(), "vmclarity-plugins", h.config.Name)
+
+	image, err := stereoscope.GetImage(ctx, h.config.ImageName)
+	if err != nil {
+		return fmt.Errorf("unable to get image(%s): %w", h.config.ImageName, err)
+	}
+
+	if _, err := os.Stat(pluginDir); os.IsNotExist(err) {
+		err = containerrootfs.ToDirectory(ctx, h.config.ImageName, pluginDir)
```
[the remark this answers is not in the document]
akijakya: https://github.com/openclarity/vmclarity/pull/1823

plugins/sdk-python/example/Dockerfile.test
```diff
@@ -40,4 +40,4 @@ COPY ./plugins/sdk-python/example ./src
 USER 65534
 
 # Run example scanner
-CMD ["python3", "/src/scanner.py"]
+CMD ["/usr/local/bin/python3", "/src/scanner.py"]
```
zsoltkacsandi, line 43: To use the binary plugin support it's a requirement to provide the full path of the binary in the image of the plugin container image. (this is also advantageous from a security point of view)

This is a limitation that comes from the Go's exec package: https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/os/exec/exec.go;drc=f7f266c88598398dcf32b448bcea2100e1702630;l=418

Whether you run a command in a chroot or not, Go will attempt to resolve the path of the binary based on the PATH set before the chroot.

plugins/runner/internal/runtimehandler/binary/handler.go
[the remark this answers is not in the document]
ramizpolic: mutex will at least resolve multiple parallel binary handlers hijacking the same port, but yes, it wont resolve potential port hijacking outside of this scope (external programs)

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,202 @@
...
+	h.cmd.SysProcAttr = &syscall.SysProcAttr{
+		Chroot: h.pluginDir,
+	}
+
+	h.cmd.Stdin = &bytes.Buffer{}
+	h.cmd.Stdout = &h.cmdStdOut
+	h.cmd.Stderr = &h.cmdStdErr
+
+	// Start command
+	openPortListener.Close()
```
zsoltkacsandi, line 139: Moving the close call of the TCP listener here to minimize the chance of a potential race condition with another processes (let that be a plugin or anything else out of our control).

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,202 @@
...
+	if workDir == "" {
+		workDir = "/"
+	}
+
+	// Initialize command
+	h.cmd = exec.CommandContext(ctx, args[0], args[1:]...)
+	h.cmd.Env = env
+	h.cmd.Dir = workDir
+
+	h.cmd.SysProcAttr = &syscall.SysProcAttr{
```
zsoltkacsandi, line 130: Instead of an explicit syscall.Chroot call in the plugin handler, we are only chrooting the plugin process.

plugins/runner/runtimehandler_linux.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/openclarity/vmclarity/plugins/runner/internal/runtimehandler"
+	"github.com/openclarity/vmclarity/plugins/runner/internal/runtimehandler/binary"
+	"github.com/openclarity/vmclarity/plugins/runner/internal/runtimehandler/docker"
+	"github.com/openclarity/vmclarity/plugins/runner/types"
+)
+
+func getPluginRuntimeHandler(ctx context.Context, config types.PluginConfig) (runtimehandler.PluginRuntimeHandler, error) {
+	if config.BinaryMode {
+		return binary.New(ctx, config)
```
zsoltkacsandi, line 31: Should we print some kind of warning here to use this method with caution?

.families.yaml
zsoltkacsandi: I have been thinking about this and I am not sure if we should use the temp dir as default.

On Linux/Unix systems it is a security best practice to mount /tmp as a separate partition/volume, often with noexec flag (so we won't be able to execute binaries from there).

There may also be a distribution dependent /tmp cleaning solution (for example: https://www.freedesktop.org/software/systemd/man/latest/tmpfiles.d.html) that may clean up the /tmp directory while we are running plugins.

.families.yaml
```diff
@@ -1,139 +1,9 @@
-sbom:
```
zsoltkacsandi, line 1: This file has to be reverted.

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	// https://lwn.net/Articles/281157/
+	// "the read-only attribute can only be added with a remount operation afterward"
+	err = syscall.Mount(h.config.InputDir, h.inputDirMountPoint, "", syscall.MS_BIND|syscall.MS_REMOUNT|syscall.MS_RDONLY, "")
+	if err != nil {
+		return fmt.Errorf("unable to remount input directory as read-only (%s - %s): %w", h.config.InputDir, h.inputDirMountPoint, err)
+	}
+
+	// TODO: handle panics - unmount
```
zsoltkacsandi, line 94: If there is a panic, the mount will be stuck.

This can also happen with the docker binary handler (it won't stop the running plugin container on a panic).

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	// Find a port
+	openPortListener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return fmt.Errorf("unable to find port")
+	}
+	port := openPortListener.Addr().(*net.TCPAddr).Port
+
+	h.pluginServerEndpoint = fmt.Sprintf("http://127.0.0.1:%d", port)
```
zsoltkacsandi, line 113: Should we make this a constant?

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func New(ctx context.Context, config types.PluginConfig) (runtimehandler.PluginRuntimeHandler, error) {
+	return &binaryRuntimeHandler{
+		config: config,
+	}, nil
+}
+
+func (h *binaryRuntimeHandler) Start(ctx context.Context) error {
+	h.pluginDir = filepath.Join(os.TempDir(), "vmclarity-plugins", h.config.Name)
```
zsoltkacsandi, line 59: Should we make this a constant?

plugins/runner/internal/runtimehandler/binary/handler.go
zsoltkacsandi: This is done.

utils/command/command.go
```diff
@@ -57,14 +57,16 @@ func (c *Command) Start(ctx context.Context) (*State, error) {
 		cmd.Dir = c.WorkDir
 	}
 
-	var stdOut, stdErr bytes.Buffer
+	var stdIn, stdOut, stdErr bytes.Buffer
+	cmd.Stdin = &stdIn
```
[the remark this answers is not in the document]
zsoltkacsandi: TODO: revert. We don't need this anymore.

.families.yaml
ramizpolic: we could use `{home dir}/vmclarity-plugins` as a default option or something else, wdyt @zsoltkacsandi?

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,233 @@
...
+
+func (h *binaryRuntimeHandler) Start(ctx context.Context) error {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	image, cleanup, err := containerrootfs.GetImageWithCleanup(ctx, h.config.ImageName)
+	if err != nil {
+		return fmt.Errorf("unable to get image(%s): %w", h.config.ImageName, err)
+	}
+	defer cleanup()
```
zsoltkacsandi, line 68: We shouldn't do the cleanup here (currently it recursively deletes the host filesystem).

This cleanup runs at the end of the `Start` function when there is still an existing mount from the host's filesystem. Before we delete anything we should be able to ensure that there is no mount from the host fs.

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,233 @@
...
+
+	go func() {
+		io.Copy(stderrPipeWriter, h.stderrPipe)
+	}()
+
+	return io.NopCloser(io.MultiReader(stdoutPipeReader, stderrPipeReader)), nil
+}
+
+func (h *binaryRuntimeHandler) Result(ctx context.Context) (io.ReadCloser, error) {
+	f, err := os.Open(filepath.Join(h.pluginDir, runtimehandler.RemoteScanResultFileOverride))
```
zsoltkacsandi, line 208: Right now this is not a problem because the plugins are running sequentially on the inputs, but if we will run them concurrently we will need to create separate result files to avoid the collision.

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,238 @@
...
+
+	if _, err := os.Stat(h.pluginDir); os.IsNotExist(err) {
+		err = containerrootfs.ToDirectory(ctx, image, h.pluginDir)
+		if err != nil {
+			return fmt.Errorf("unable to extract image(%s): %w", h.config.ImageName, err)
+		}
+	}
+
+	// Mount input from host
+	// /tmp/vmclarity-plugins/kics + /input + /host-dir-to-scan
```
akijakya, line 86:
```suggestion
	// /home/ubuntu/.vmclarity/plugins/kics/<id> + /input + /host-dir-to-scan
```
Right?

plugins/runner/go.mod
```diff
@@ -6,65 +6,112 @@ require (
 	github.com/deepmap/oapi-codegen/v2 v2.2.0
 	github.com/docker/docker v26.1.4+incompatible
 	github.com/docker/go-connections v0.5.0
+	github.com/hashicorp/go-multierror v1.1.1
 	github.com/openclarity/vmclarity/core v0.7.1
 	github.com/openclarity/vmclarity/plugins/sdk-go v0.7.1
+	github.com/openclarity/vmclarity/scanner v0.7.1
```
ramizpolic, line 12: we shouldnt import scanner here since we import runner in the scanner itself (also, we should use replaces for internal code rather than from releases). this is also why the vet is failing

my suggestion is to move `scanner/utils/containerrootfs` to `utils/containerrootfs` module, and import `github.com/openclarity/vmclarity/utils` here instead, but using replace directives

akijakya: There is also a `utils/fsutils`, could that also be its new home?

plugins/runner/runner.go
```diff
@@ -35,35 +34,39 @@ import (
...
+		config:         config,
+		runtimeHandler: handler,
 	}, nil
 }
 
 func (r *pluginRunner) Start(ctx context.Context) error {
-	if err := r.containerManager.Start(ctx); err != nil {
+	go func(ctx context.Context) {
+		<-ctx.Done()
+		r.Remove(ctx) //nolint:errcheck
```
ramizpolic, line 58: this will call remove multiple times. for docker, its okay, not sure if its okay for binary mode

utils/go.mod
```diff
@@ -3,15 +3,89 @@ module github.com/openclarity/vmclarity/utils
 go 1.22.4
 
 require (
+	github.com/anchore/stereoscope v0.0.2
 	github.com/go-viper/mapstructure/v2 v2.0.0
 	github.com/moby/sys/mountinfo v0.7.1
 	github.com/onsi/gomega v1.33.1
+	github.com/openclarity/vmclarity/core v0.7.1
```
ramizpolic, line 10: should use replace directive in all places that import vmclarity packages, unrelated if indirect or direct import

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+	}
+
+	// Unmount input directory
+	if err := syscall.Unmount(h.inputDirMountPoint, 0); err != nil {
+		removeErr = multierror.Append(removeErr, fmt.Errorf("failed to kill plugin process: %w", err))
+	} else {
+		// Call the cleanup function for the image only after the input directory is unmounted, or else it will also remove
+		// the root filesystem mounted under input
+		h.imageCleanup()
```
ramizpolic, line 237: in case we want to preserve the image between runs, would this have any effect on the rest of the flow when binary plugin scan is re-executed for the same input?

plugins/runner/internal/runtimehandler/binary/handler.go
```diff
@@ -0,0 +1,241 @@
...
+		return fmt.Errorf("unable to get image(%s): %w", h.config.ImageName, err)
+	}
+	h.imageCleanup = cleanup
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("unable to determine user's home directory: %w", err)
+	}
+
+	h.pluginDir = filepath.Join(home, ".vmclarity/plugins", h.config.Name, image.Metadata.ID)
```
ramizpolic, line 79: it would be good if we could expose two more options on the plugin config to allow users to
a) specify custom dir if they want the plugin data to be saved/loaded to allow caching via `binary_artifacts_path` which would default to `${HOME}/.vmclarity/plugins` if not specified
b) specify cleanup flag if they want to preserve the data or not between runs `binary_artifacts_keep` to allow reuse when used with flag above. We could also use `binary_artifacts_clean` to default to preserve behaviour when not specified

For example
```yaml
## reruns preserve new and reuse existing artifacts
binary_artifacts_path: /tmp/plugin-artifacts
binary_artifacts_keep: true

## reruns reuse existing artifacts between runs but do not preserve the new artifacts
binary_artifacts_path: /tmp/plugin-artifacts
binary_artifacts_keep: false

## reruns neither reuse nor preserve
# empty binary_artifact configs
```

plugins/runner/internal/runtimehandler/binary/handler.go
ramizpolic: also, is there a possibility that the root filesystem will ever be deleted with the existing approach?

plugins/runner/internal/runtimehandler/binary/handler.go
ramizpolic: we can do this in a separate PR, added an issue https://github.com/openclarity/vmclarity/issues/1854 to track this

plugins/runner/internal/runtimehandler/binary/handler.go
akijakya: I also have my concerns, if it is okay not to clean up the plugin scanner image from the machine after the scan, we could drop the image cleanup altogether from the binary plugin handler.

Also during implementation, we optimized for the fact that within VMClarity, one scan will only have one input, but in the light of recent developments, we should rethink some parts of the code.

zsoltkacsandi: I think there are multiple things here that we should think it through.

1. preserving the image

If the goal is to have a local cache of the image and save execution time/bandwidth by not re-downloading the image every time, then we can store the tar file somewhere else, and extract it on a plugin execution.

With the current approach we are mixing this local cache functionality with the plugin execution environment (the `/home/ubuntu/.vmclarity/plugins/kics` directory is the image cache and the root filesystem of the chroot environment where the kics plugin runs), that has multiple potential consequences:
- at some point we need to invalidate the cache, check if there is a new version, which can be a source of race condition if there is a running plugin in the directory
- it's not safe to run concurrently (currently the plugin family runner executes the plugins one after the other)

To solve this we can have a separate plugin cache (the downloaded tar file of the images), and on plugin execution we can extract the image tar file from our local cache, and create a dedicated and unique directory that we clean up after one plugin execution.

2. accidental deletion of root filesystem

With the read-only bind mount this is not supposed to happen in any case, yet somehow it does. I have a feeling that something is off here and we should investigate this further, understand what is happening in the background, so we can prevent from happening in the future.

We have other options to solve this issue, but that would complicate the implementation, and we would getting closer to a full-blown container runtime implementation.

plugins/runner/runner.go
zsoltkacsandi: I don't think it is okay, we should track the internal lifecycle of the plugin execution.

ramizpolic: marking resolved as we can call remove multiple times per discussion

plugins/runner/internal/runtimehandler/binary/handler.go
ramizpolic: @akijakya will create an issue to summarise expected changes after the scanner refactor ticket and we will track all these expected changes there. it is not relevant to address it now

ramizpolic (approved):
great work everyone!! :shipit:

PR #1820, opened 2024-06-19
fix(plugin): add healthz check retry mechanism

akijakya:
## Description

Instead of exiting with an error, print out the error message and continue with the retries in case checking the plugin's healthz endpoint fails.

There might be cases when the server is not yet ready to accept connections.

## Type of Change

- [X] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


plugins/runner/runner.go
```diff
@@ -92,22 +92,24 @@ func (r *pluginRunner) WaitReady(ctx context.Context) error {
...
+					return fmt.Errorf("failed to set plugin server client: %w", err)
+				}
 			}
 
 			// Check for plugin server once container is ready
-			// TODO: add retry mechanism
 			resp, err := r.client.GetHealthzWithResponse(ctx)
 			if err != nil {
-				return fmt.Errorf("failed to get plugin server healthz: %w", err)
+				fmt.Printf("failed to get plugin server healthz: %s, retrying...", err.Error())
```
ramizpolic, line 111: we should use logger here

zsoltkacsandi (approved):

PR #1821, opened 2024-06-19
feat(utils): set stdin explicitly for command

zsoltkacsandi:
## Description

If there is no stdin set for the command, the standard library attempts to use the /dev/null device, which is not always available.

https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/os/exec/exec.go;drc=f7f266c88598398dcf32b448bcea2100e1702630;l=506

## Type of Change

- [ ] Bug Fix
- [X] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/utils/command/command.go
+++ b/utils/command/command.go
@@ -57,7 +57,8 @@ func (c *Command) Start(ctx context.Context) (*State, error) {
 		cmd.Dir = c.WorkDir
 	}
 
-	var stdOut, stdErr bytes.Buffer
+	var stdIn, stdOut, stdErr bytes.Buffer
+	cmd.Stdin = &stdIn
 	cmd.Stdout = &stdOut
 	cmd.Stderr = &stdErr
 
@@ -65,6 +66,7 @@ func (c *Command) Start(ctx context.Context) (*State, error) {
 
 	return &State{
 		cmd:    cmd,
+		Stdin:  &stdIn,
 		StdOut: &stdOut,
 		StdErr: &stdErr,
 	}, err
@@ -76,6 +78,7 @@ func (c *Command) String() string {
 
 type State struct {
 	cmd    *exec.Cmd
+	Stdin  *bytes.Buffer
 	StdOut *bytes.Buffer
 	StdErr *bytes.Buffer
 }


utils/command/command.go
```diff
@@ -76,6 +78,7 @@ func (c *Command) String() string {
 
 type State struct {
 	cmd    *exec.Cmd
+	Stdin  *bytes.Buffer
```
ramizpolic, line 81: could be a bit counterintuitive since we dont use stdin anywhere from the state itself. `State` can serve for read/watch-only for now

zsoltkacsandi: I would agree if there was not a `(s *State) Kill()` function on the `State` struct, and that is for process manipulation.

I am totally fine with not exposing the stdin, but the abstraction that this package provides does not feel right to me.

The end user of this library cannot access the underlying `exec.Command` in any way (it is a private field of the `State` struct), so it hides all of the capabilities that the `exec.Command` provides. 

I see three options:
- not use this package at all where we need the options of exec.Command that this package does not expose
- refactor this package (including all the places where it is used)
- keep stdin "hidden", but that wouldn't solve the problem if any of the future users of this package tries to access the stdin of a process (which is a common use case)

zsoltkacsandi: Just one additional note for the above:

if you take a look at any of the methods on the `State` struct, it wraps the corresponding method of the underlying command, and does not do anything else.

Question if this abstraction provides any value.

ramizpolic: both `Kill()` and `Wait()` are lifecycle ops and do not really change the state of what the command is doing, while writing to `Stdin` can change the behaviour of the underlying command. 

From the abstraction level, it does provide value to _quickly_ and _explicitly_ spawn and handle commands. The reason why I suggested to keep stdin private is because we don't have commands that require input and we want to discourage having them as they would complicate the overall flow. I am aware that it can happen that we need stdin, but in most cases there are alternatives to using stdin (env, flags, etc). Plus, its difficult to test the flow when you have inputs to deal with.

zsoltkacsandi: > From the abstraction level, it does provide value to quickly and explicitly spawn and handle commands.

So does `os/exec.Command`:

```
cmd := exec.CommandContext(
  ctx, 
  args[0], 
  args[1:]
)
cmd.Env = env
cmd.Dir = workDir

cmd.Wait()

pid := cmd.ProcessState.Pid()
exited := cmd.ProcessState.Exited()
cmd.Process.Kill()
```

Versus `utils.command`:

```
cmd := command.Command{
	Cmd:     args[0],
	Args:    args[1:],
	Env:     env,
	WorkDir: workDir,
}
	
state, _ := cmd.Start(ctx)

commandState.Wait()

pid := state.Pid()
exited := state.Exited()
state.Kill()
```

It's pretty much the same amount line of code, and you have to provide the same arguments with exactly the same API.

> reason why I suggested to remove stdin from State is because we don't have commands that require input and we want to discourage having them as they would complicate the overall flow

As I said, I agree with that. What I wanted to point out is that there is a question that IMO needs to be answered: what is the purpose of this package, and that's why I proposed the three options above:

> not use this package at all where we need the options of exec.Command that this package does not expose

a, if it is to be used where we want to spawn commands and just observe them (in a read-only way, and only for that use case), it is fine, but we should communicate that clearly, because it may cause confusion among contributors

> refactor this package (including all the places where it is used)

b, if we want to provide a general and consistent way to run and handle external commands in VMClarity, then
  1, we need to expose the same API that the underlying command has or
  2, we shouldn't hide the underlying command

> keep stdin "hidden", but that wouldn't solve the problem if any of the future users of this package tries to access the stdin of a process (which is a common use case)

Setting the stdin internally because we now have one use case where it causes a problem (binary plugin handler) does not make sense in my opinion.

ramizpolic: > what is the purpose of the package?

To simplify working with commands that don't require sophisticated flow control. If the command requires more advanced control, it makes sense to fall back to native command package. We are not trying to replace it but simplify common ops. 

There is no need to overengineer utils package, it's opinionated by default. I share your concerns, but imo switching back to native command everywhere or refactoring/consolidating interface will take time and will provide very little value. We can expose the underlying command as it would address some of the points you mentioned, but I wouldn't pursue it further than that. Hope this clarifies it.

PR #1822, opened 2024-06-19
ci(ci): add eslint

adamtagscherer:
## Description

This PR adds eslint, prettifier and react testing library to the frontend codebase.

## Type of Change

- [ ] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [x] Other (please describe)

## Checklist

### DIFF
--- a/ui/vite.config.ts
+++ b/ui/vite.config.ts
@@ -34,5 +34,9 @@ export default defineConfig(({ mode }) => {
                 },
             },
         },
+        test: {
+            globals: true,
+            environment: 'happy-dom'
+        }
     }
 });
--- a/.github/workflows/reusable-verification.yml
+++ b/.github/workflows/reusable-verification.yml
@@ -82,6 +82,24 @@ jobs:
           make lint-actions
           echo "::remove-matcher owner=actionlint::"
 
+  javascript:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+
+      - name: Install modules
+        working-directory: ./ui
+        run: npm install
+
+      - name: Run ESLint
+        working-directory: ./ui
+        run: npm run lint
+
+      - name: Run tests
+        working-directory: ./ui
+        run: npm run test
+
   go:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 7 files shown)


misi-bp (dismissed):
I am not familiar with some parts of the unit test configs set and the configuration opportunities, but if it works, then amen.

.github/workflows/unit-testing-js.yaml
```diff
@@ -0,0 +1,19 @@
+name: Unit Tests on JavaScript
```
csatib02, line 1: I think this can also be part of the `reusable-verification`, since we also run linters and tests after each-other in case of golang as part of the verification.

.github/workflows/ci.yml
```diff
@@ -18,6 +18,11 @@ jobs:
     uses: ./.github/workflows/reusable-verification.yml
     secrets: inherit
 
+  unit-test-js:
```
csatib02, line 21: Referring to my other comment, this can be removed, in case you agree with me. 😁

csatib02 (commented):
I don't know much about frontend, but I'm in the middle of restructing the whole pipeline for VMClarity, and the only part that can be left intact is the reusable-verification workflow. 😁

.github/workflows/unit-testing-js.yaml
adamtagscherer: Sure, let's do that.

csatib02 (approved):
LGTM(ci)!

akijakya (approved):

ui/vite.config.ts
```diff
@@ -34,5 +34,9 @@ export default defineConfig(({ mode }) => {
                 },
             },
         },
+        test: {
+            globals: true,
+            environment: 'happy-dom'
```
gallotamas, line 39: _(no need to change anything happy-dom is a great choice I'm just adding a remark here)_
I guess you checked out the different environments but if not then then keep in mind that if you run into compatibility issues then you might have to switch to jsdom
A rough rule of thumb is that happy-dom might be faster (depending on your use case) but jsdom might be more complete (simulates more of what a browser would).

gallotamas (approved):
great work

ui/vite.config.ts
adamtagscherer: Thanks, I'll keep that in mind! :)

PR #1824, opened 2024-06-20
fix(plugin): fix plugin family runner race condition

zsoltkacsandi:
## Description

If an error occurs, we send the result back to the family manager on a channel and return from the function:

```
if err := rr.WaitReady(ctx); err != nil {
  s.sendResults(retResults, fmt.Errorf("failed to wait for plugin scanner to be ready: %w", err))
  return
}
```

As part of the return, the deferred `Stop` and `Remove` lifecycle functions begin to run:

```
defer func() {
	if err := rr.Stop(ctx); err != nil {
		s.logger.WithError(err).Errorf("failed to stop runner")
	}

	// TODO: add short wait before removing to respect container shutdown procedure

	if err := rr.Remove(ctx); err != nil {
		s.logger.WithError(err).Errorf("failed to remove runner")
	}
}() //nolint:errcheck
```

Once the family manager gets the result in the result channel, it does not really care if the underlying goroutines are finished or not:

https://github.com/openclarity/vmclarity/blob/07500346ae7c80f1201992cdb3c22d0b6377d708/scanner/families/manager.go#L148

This results in a race condition, once the family manager gets all the results it expects, the program will terminate regardless of whether the plugins' Stop() and Destroy() functions have finished or not.

We need to call the `Exit` and `Destroy` lifecycle functions before sending back the result.
 

## Type of Change

- [X] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/scanner/families/plugins/runner/runner.go
+++ b/scanner/families/plugins/runner/runner.go
@@ -78,7 +78,8 @@ func (s *Scanner) Run(ctx context.Context, sourceType utils.SourceType, userInpu
 			s.sendResults(retResults, fmt.Errorf("failed to create plugin runner: %w", err))
 			return
 		}
-		defer func() {
+
+		cleanup := func(ctx context.Context) {
 			if err := rr.Stop(ctx); err != nil {
 				s.logger.WithError(err).Errorf("failed to stop runner")
 			}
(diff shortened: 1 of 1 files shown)


akijakya:
Nice catch! This behavior can be found in other families and should be fixed there as well...

Just a thought: what if the cleanup func is added to the `Scanner` struct, and the `sendResults` is refactored in a way that it calls `s.cleanup` first, it can even be renamed to `finish()` or something?

zsoltkacsandi:
> Just a thought: what if the cleanup func is added to the `Scanner` struct, and the `sendResults`

That can work too, the reason why I chose this approach, because it's more expressive.
If someone reads the code, you can read as a sentence, `first we are cleaning up, then we are sending back the result, and then returning from the function`. The intent is clear.

To make part of the `sendResults` feels like creating a side effect. The name of the function says that I am sending back a result. There would be a hidden behaviour that you wouldn't expect without taking a look into the `sendResults` function.

>  it can even be renamed to finish() or something?

Totally agree, we can use a more descriptive name here. finishRunner or something like that?

akijakya:
> Totally agree, we can use a more descriptive name here. finishRunner or something like that?

`finishScannerRun`?

ramizpolic (approved):

PR #1843, opened 2024-06-24
feat(api): generate frontend oapi code

adamtagscherer:
## Description

- Generate frontend SDK.
- Added UI related makefile targets.
- Updated docs.

- Changing to react queries will be in a follow-up PR.

## Type of Change

- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


CONTRIBUTING.md
```diff
@@ -153,24 +152,30 @@ header.
...
 - `make lint-helm` lints the Helm chart.
 
 `make fix` is also provided which can automatically resolve lint issues such as formatting.
 
 ### Unit tests
 
-`make test` can be used run all the unit tests in the repo. Alternatively you can use the standard go test CLI to run a
-specific package or test like:
+`make test` to run both go and js unit tests.
+Altrenatively you can run them separetly with:
```
adamtagscherer, line 163: typo

ui/src/api/index.ts
adamtagscherer, line 1: we don't reexport things in an `index.ts` file so far nowhere in the project AFAIK, let's not start it now
we can discuss it though

ui/src/components/WrappingTextBoxWithEllipsis/wrapping-text-box-with-ellipsis.scss
```diff
@@ -1,6 +1,6 @@
 .wrapping-text-box-with-ellipsis-wrapper {
     display: flex;
-    align-items: end;
```
adamtagscherer, line 3: ? :D

csatib02: ne szemtelenkedj :D

Makefile
```diff
@@ -318,10 +342,15 @@ docker-scanner-plugins: ## Build scanner plugin container images
 ##@ Code generation
 
 .PHONY: gen
-gen: gen-api gen-bicep gen-helm-docs ## Generating all code, manifests, docs
+gen: gen-api-js gen-api-go gen-bicep gen-helm-docs ## Generating all code, manifests, docs
+
+.PHONY: gen-api-js
+gen-api-js: ## Generating Javascript library for API specification
+	$(info Generating API for UI code ...)
+	@(cd ui && npx @openapitools/openapi-generator-cli generate -i ../api/openapi.yaml -g typescript-axios -o ./src/api/generated --openapi-normalizer SET_TAGS_FOR_ALL_OPERATIONS=VMClarity)
```
gallotamas, line 350: This should be added to package.json as an npm command to make it easy for UI developers to (re)generate the api and hide the complexity of generating the api. It doesn't belong to the make file in my opinion.
Here you should only run the appropriate npm command e.g. `cd ui && npm run generate-api`
Probably it should be added as a postinstall script to the npm package.json as well because the UI can't be built if you don't run it before the build. e.g. `"postinstall": npm run generate-api"`

gallotamas (changes requested):
There are 2 important things that are missing from this PR.

- VMClarity has 2 APIs that are used from the UI and we are only generating code to one of them here. Please add code generation for the UI Backend as well: https://github.com/openclarity/vmclarity/blob/main/uibackend/openapi.yaml
- we should add npm commands to (re)generate code while working on the frontend. See my comment above for more details

ui/src/api/generated/base.ts
```diff
@@ -0,0 +1,86 @@
+/* tslint:disable */
```
gallotamas, line 1: I think we shouldn't commit any of the generated files. If the openapi doesn't change then the output should always be the same and if it does we don't really care about the changes unless it breaks the CI but then an error will be thrown and you won't be able to ship your code until you fix it.

ui/package.json
```diff
@@ -49,10 +52,12 @@
   },
   "scripts": {
     "start": "vite",
-    "build": "tsc && vite build",
+    "build": "tsc  && vite build",
```
csatib02, line 55: there is an additional space character here that should be removed!

csatib02 (approved):
Small small issue but overall Looks Good To Me! {Rocket}

akijakya (approved):
🐳

gallotamas (approved):
great job! 🎉

PR #1828, opened 2024-06-23
feat(api): add summary to package findings

ramizpolic:
## Description

Adds summary field to findings. Registers a new processor for findings to calculate vulnerability severity count for applicable finding types (i.e. only for packages).

Closes https://github.com/openclarity/vmclarity/issues/1040

## Type of Change

- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/openclarity/vmclarity/blob/main/CONTRIBUTING.md)
- [x] Existing issues have been referenced (where applicable)
- [x] I have verified this change is not present in other open pull requests
- [x] Functionality is documented
- [x] All code style checks pass
- [x] New code contribution is covered by automated tests
- [x] All new and existing tests pass

PR #1848, opened 2024-06-25
ci(workflow): enable community contributions

csatib02:
## Description

- This PR enables community contributions, by dropping `ghcr` uploads on pull requests, and instead uses artifacts.
- If there is a release ghcr.io will be used, no change was made there.
- In case of pull requests, artifacts are created and used for e2e testing.
- No comments will be posted under PR's with the images, instead the build artifacts can be used for further manual testing.
- Minor design changes.

## Type of Change

- [ ] Bug Fix
- [X] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


.github/workflows/reusable-end-to-end-testing.yml
```diff
@@ -7,63 +7,14 @@ on:
         required: true
         type: string
         description: 'Image tag to build and push.'
-      registry_name:
-        required: false
-        type: string
-        description: 'Registry name used for container image names. Default is `ghcr.io/openclarity`.'
-        default: ghcr.io/openclarity
```
ramizpolic, line 14: we could keep this field to use in `Run end to end tests` step

.github/workflows/build-and-push-components-release.yml
```diff
@@ -0,0 +1,112 @@
...
+    inputs:
+      image_tag:
+        required: true
+        type: string
+        description: 'Image tag to build and push.'
+      registry_name:
+        required: false
+        type: string
+        description: 'Name of the image to publish'
+        default: ghcr.io/wewanttocontributetovmclarity
```
ramizpolic, line 14:
```suggestion
        default: ghcr.io/openclarity
```

.github/workflows/reusable-end-to-end-testing.yml
csatib02: As far as I know we only run e2e tests as part of the CI run, I deleted it because ghcr will never be used on CI runs.

adamtagscherer:
1. What will happen if we want to wire in the cloud e2e tests? Will we have the `latest` tagged images on the `main` branch? If I see it correctly then we build and push to `ghcr` images only when creating a release.

> No comments will be posted under PR's with the images, instead the build artifacts can be used for further manual testing.

2. How can I use the build artifacts for manual testing? For example if I want to deploy the whole system on AWS and test my changes there, how do I overwrite the images in the cloudformation file?
I guess we have to build the images locally, upload them to `DockerHub` and use those in the cloudformation file. I have no objections against it, if the others agree that it's a good trade off for enabling the members to contribute.

csatib02:
@adamtagscherer

> 1. What will happen if we want to wire in the cloud e2e tests? Will we have the `latest` tagged images on the `main` branch? If I see it correctly then we build and push to `ghcr` images only when creating a release.

We are building images when we are pushing to main, check `main-merge.yml`, those images will be tagged latest, and can be used for the e2e tests for the different environments.

> 
> > No comments will be posted under PR's with the images, instead the build artifacts can be used for further manual testing.
> 
> 2. How can I use the build artifacts for manual testing? For example if I want to deploy the whole system on AWS and test my changes there, how do I overwrite the images in the cloudformation file?
>    I guess we have to build the images locally, upload them to `DockerHub` and use those in the cloudformation file. I have no objections against it, if the others agree that it's a good trade off for enabling the members to contribute.

We can create a workflow that can be ran manually by org members, to upload images to ghcr tagged with `dev`, so it can be used for such use-cases you mentioned.
The artifacts themselves can be found on the actions page if you scroll down. The workflow creates docker images, that can be loaded to your own local docker registry or any with `docker load {image}`.

.github/workflows/reusable-end-to-end-testing.yml
csatib02: Added back this field! 👍

ramizpolic:
> 1. What will happen if we want to wire in the cloud e2e tests? Will we have the `latest` tagged images on the `main` branch? If I see it correctly then we build and push to `ghcr` images only when creating a release.
> 
> > No comments will be posted under PR's with the images, instead the build artifacts can be used for further manual testing.
> 
> 2. How can I use the build artifacts for manual testing? For example if I want to deploy the whole system on AWS and test my changes there, how do I overwrite the images in the cloudformation file?
>    I guess we have to build the images locally, upload them to `DockerHub` and use those in the cloudformation file. I have no objections against it, if the others agree that it's a good trade off for enabling the members to contribute.

@adamtagscherer raises important concerns and i would also like to add to this. Although we are not pushing PR images to GHCR, we could solve this by creating another (manually triggered) workflow that takes specific PR commit reference and pushes built artifacts for that commit to GHCR. This would allow us to test things, would also unblock community contributions, and it would also address the security concerns. My suggestion is that we add this action in a separate PR after this one is merged (it can be reused similarly to pushes to main).

I wrote a small PoC bash script that can import images from a specific GA pipeline run into local docker registry. I have tested it against GA run https://github.com/openclarity/vmclarity/actions/runs/9664752735. You need GitHub CLI to run this. Note that pulling artifacts can take some time. It can be used in case we use docker-only approach for E2E testing on cloud providers. It does not solve directly the second concern regarding CF file (unless it is extended with this script). But the suggestion above

```bash
#!/usr/bin/env bash

## Defines which image artifacts to download
GH_ACTION_RUN_ID="9660640440"
PLATFORM="arm64"

## Defines how to tag loaded VMClarity images
IMPORTED_IMAGE_REGISTRY="ghcr.io/openclarity"
IMPORTED_IMAGE_TAG="latest"

## Set path to temporary dir where artifacts will be pulled into
mkdir -p /tmp/vmclarity-images
cd  /tmp/vmclarity-images

## Download all artifacts using GitHub CLI
gh run download --repo openclarity/vmclarity ${GH_ACTIONS_RUN_ID}

## Load all VMClarity platform-specific images
for image_archive in vmclarity-*-${PLATFORM}/*.tar; do
  # load image
  image_name=$(basename $image_archive .tar)
  image_load_output=$(docker load --input "$image_archive" --quiet)
  image_id=${image_load_output#"Loaded image ID: sha256:"} # strip prefix, very hacky due to https://github.com/moby/moby/issues/40188 limitations

  # tag image
  tagged_image="$IMPORTED_IMAGE_REGISTRY/$image_name:$IMPORTED_IMAGE_TAG"
  echo "Tagging image ID=$image_id as $tagged_image"
  docker tag $image_id $tagged_image
done

## Cleanup artifacts temporary dir
cd ~
rm -rf /tmp/vmclarity-images
```

csatib02:
> @adamtagscherer raises important concerns and i would also like to add to this. Although we are not pushing PR images to GHCR, we could solve this by creating another (manually triggered) workflow that takes specific PR commit reference and pushes built artifacts for that commit to GHCR. This would allow us to test things, would also unblock community contributions, and it would also address the security concerns. My suggestion is that we add this action in a separate PR after this one is merged (it can be reused similarly to pushes to main).

> I wrote a small PoC bash script that can import images from a specific GA pipeline run into local docker registry. I have tested it against GA run https://github.com/openclarity/vmclarity/actions/runs/9660640440. You need GitHub CLI to run this. Note that pulling artifacts can take some time. It can be used in case we use docker-only approach for E2E testing on cloud providers. It does not solve directly the second concern regarding CF file (unless it is extended with this script). But the suggestion above

@ramizpolic I also agree that there should be a way to manually upload to ghcr to further test changes, let's sit together and decide on a solution that works for everyone.

We had something like this created by @zsoltkacsandi recently in #1768, I removed it in this PR since essential parts of the workflow has been reworked, and I believe we need to adhere with the changes made on it.

ramizpolic (dismissed):
LGTM, great work on this @csatib02!

ramizpolic:
lets just wait for #1828 #1794 to be merged first

adamtagscherer (approved):
LGTM :shipit:

.github/workflows/build-and-push-components-release.yml
```diff
@@ -0,0 +1,112 @@
...
+
+  build-and-push:
+    needs: prepare-build
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        targets: ${{ fromJson(needs.prepare-build.outputs.targets) }}
+        platform:
+          - linux/amd64
```
akijakya, line 54: In this current form, the images would be built parallel by platform in a matrix and then pushed separately instead of merging the two and then pushing one multi-arch image as in the original flow. 

We already spoke about this with Bence and concluded that 
- we don't even need to build for two platforms in case of producing the image for the e2e test, just `linux/amd64` and Docker Bake can handle this
- in case of releasing, since it happens rarely, we don't need to spare these minutes by building for the two Linux platforms parallel in a matrix and then combining those to a single multi-arch image to be pushed to ghcr.io, but we can leave this also to Docker Bake by setting `*.platform=linux/arm64,linux/amd64` in the CI step

akijakya (changes requested):

.github/workflows/build-and-push-components-release.yml
csatib02: I have added the changes we have discussed:

- Added a workflow that can be run on any pushed branch, and it also pushes images to `ghcr.io/openclarity` with `-dev` suffix and `latest` tag:
image

- When a branch is merged into main, images are uploaded to `ghcr` with `latest` tag:
image

- When there is a release, images are pushed to `ghcr` tagged with the `release-tag`
image

adamtagscherer (dismissed):

akijakya (dismissed):
LGTM!

.github/workflows/build-and-push-components-manual.yml
```diff
@@ -0,0 +1,14 @@
+name: Build and Push Components Manual
+
+on:
+  workflow_dispatch:
```
ramizpolic, line 4: just checking here if we did any tests that we can check for this step. would be good to include those details here. im not sure if this will work as intended, so please share more info here

csatib02: I've collected details on the recent changes here:

https://github.com/openclarity/vmclarity/pull/1848#discussion_r1657030770

ramizpolic (approved):

ramizpolic (dismissed):

ramizpolic (approved):

PR #1867, opened 2024-07-05
feat(plugin): make reportFormats param configurable

csatib02:
## Description

- Makes the `ReportFormats` field configurable.

## Type of Change

- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


e2e/kics_output_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/openclarity/vmclarity/scanner/families"
+	"github.com/openclarity/vmclarity/scanner/families/plugins"
+	"github.com/openclarity/vmclarity/scanner/families/plugins/common"
+	"github.com/openclarity/vmclarity/scanner/families/plugins/runner/config"
+	"github.com/openclarity/vmclarity/scanner/families/types"
+	"github.com/openclarity/vmclarity/scanner/utils"
+)
+
+var _ = ginkgo.Describe("Running a KICS scan", func() {
+	ginkgo.Context("which scans an openapi.yaml file and has report-formats set to sarif", func() {
```
ramizpolic, line 34: we can move this to `e2e/kics_test.go` to handle all kics-related tests in one place

plugins/store/kics/main.go
```diff
@@ -92,15 +95,21 @@ func (s *Scanner) Start(config types.Config) {
 			return
 		}
 
-		rawOutputFile := filepath.Join(os.TempDir(), "kics.json")
+		// Ensure JSON format is always included,
+		// since it's the only format that can be consumed by VMClarity
+		clientConfig.ReportFormats = ensureJSONFormat(clientConfig.ReportFormats)
```
ramizpolic, line 100: we can drop `ensureJSONFormat` function and move this part directly to `createConfig` method

plugins/store/kics/main.go
```diff
@@ -171,46 +180,105 @@ func (s *Scanner) createConfig(input *string) (*ScannerConfig, error) {
...
 }
 
-func (s *Scanner) formatOutput(rawFile, outputFile string) error {
-	file, err := os.Open(rawFile)
+func (s *Scanner) formatOutput(rawOutputDir, outputFile string, reportFormats []string) error {
+	var wg sync.WaitGroup
+	var resultMutex sync.Mutex
+	var result types.Result
+	errCh := make(chan error, len(reportFormats))
+	for _, format := range reportFormats {
```
ramizpolic, line 188: overall, nice idea! however, we are not gaining a lot of performances here by parallelising 2 output formats which are quick ops. imo there is not much gained here compared to having these ops simpler and synchronous

ramizpolic (changes requested):

plugins/store/kics/main.go
csatib02: I added it before returning the config, since if we get a config as input, and it does not have JSON included, it would be overwritten.

```go
func (s *Scanner) createConfig(scannerConfig *string) (*ScannerConfig, error) {
	config := ScannerConfig{
		PreviewLines:     3,
		// If we would place it here, it might be overwritten with the posted config
                // when we unmarshal on this.
		// ReportFormats:    []string{"json"},
		Platform:         []string{"Ansible", "CloudFormation", "Common", "Crossplane", "Dockerfile", "DockerCompose", "Knative", "Kubernetes", "OpenAPI", "Terraform", "AzureResourceManager", "GRPC", "GoogleDeploymentManager", "Buildah", "Pulumi", "ServerlessFW", "CICD"},
		MaxFileSizeFlag:  100,
		DisableSecrets:   true,
		QueryExecTimeout: 60,
		Silent:           true,
		Minimal:          true,
	}

	if scannerConfig == nil || *scannerConfig == "" {
		return &config, nil
	}

	if err := json.Unmarshal([]byte(*scannerConfig), &config); err != nil {
		return nil, fmt.Errorf("failed to unmarshal JSON config: %w", err)
	}

	// Ensure JSON format is always included,
	// since it's the only format that can be consumed by VMClarity
	config.ReportFormats = ensureJSONFormat(config.ReportFormats)

	return &config, nil
}
```

plugins/store/kics/main.go
csatib02: I do agree with you 100%, but I don't believe it can be that much simpler.

This would be the synchronous way:

```go
	var result types.Result
	for _, format := range reportFormats {
		switch format {
		case "json":
			var summaryJSON model.Summary
			err := decodeFile(filepath.Join(rawOutputDir, "kics.json"), &summaryJSON)
			if err != nil {
				return fmt.Errorf("failed to decode kics.json: %w", err)
			}

			var misconfigurations []types.Misconfiguration
			for _, q := range summaryJSON.Queries {
				for _, file := range q.Files {
					misconfigurations = append(misconfigurations, types.Misconfiguration{
						Id:          types.Ptr(file.SimilarityID),
						Location:    types.Ptr(file.FileName + "#" + strconv.Itoa(file.Line)),
						Category:    types.Ptr(q.Category + ":" + string(file.IssueType)),
						Message:     types.Ptr(file.KeyActualValue),
						Description: types.Ptr(q.Description),
						Remediation: types.Ptr(file.KeyExpectedValue),
						Severity:    types.Ptr(mapKICSSeverity[q.Severity]),
					})
				}
			}

			result.Vmclarity.Misconfigurations = &misconfigurations
			result.RawJSON = summaryJSON

		case "sarif":
			var summarySarif interface{}
			err := decodeFile(filepath.Join(rawOutputDir, "kics.sarif"), &summarySarif)
			if err != nil {
				return fmt.Errorf("failed to decode kics.sarif: %w", err)
			}

			result.RawSarif = &summarySarif
		
		default:
			return fmt.Errorf("unsupported report format: %s", format)
		}
	}

	if err := result.Export(outputFile); err != nil {
		return fmt.Errorf("failed to save KICS result: %w", err)
	}

	return nil
```

KICS supports these output formats:

```go
	"json"
	"sarif"
	"html"
	"glsast"
	"pdf"
	"sonarqube"
	"cyclonedx"
	"junit"
	"asff"
	"csv"
	"codeclimate"
```
If we need to add even more in the future, this might be a heavier operation.
Will you be okay with keeping this?
I cleaned up the code. <3

plugins/store/kics/formatter/formatter.go
ramizpolic, line 1: we can keep this one in `store` directly rather than pulling it into a separate module

ramizpolic (approved):
lgtm, thanks @csatib02!

PR #1893, opened 2024-07-09
refactor(ci): refactor build and push components

zsoltkacsandi:
## Type of Change

- [ ] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [X] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


ramizpolic (approved):

.github/workflows/build-and-push-components.yml
```diff
@@ -22,6 +22,35 @@ on:
...
+    inputs:
+      git_ref:
+        required: false
+        type: string
+        description: 'The branch, tag or SHA to build.'
+        default: ''
+      image_tag:
+        required: true
+        type: string
+        description: 'Image tag to use.'
```
akijakya, line 35:
```suggestion
        description: 'Image tag to use.'
        default: latest
```

I think the default could still be `latest`

.github/workflows/build-and-push-components.yml
```diff
@@ -22,6 +22,35 @@ on:
...
+      registry_name:
+        required: false
+        type: string
+        description: 'Name of the registry.'
+        default: ghcr.io/openclarity
+      suffix:
+        required: false
+        type: string
+        description: 'Suffix to append to the image name'
+        default: ""
```
akijakya, line 45:
```suggestion
        default: "-dev"
```

I think the default could still be `-dev`

.github/workflows/build-and-push-components.yml
```diff
@@ -22,6 +22,35 @@ on:
...
+      bake-group:
+        required: false
+        type: string
+        description: 'Name of the Docker Bake group of targets'
+        default: default
+
+permissions:
+  packages: write
```
csatib02, line 53: This workflow should have one responsibility, now it has two, I believe we should keep the manual workflow, and have these specified there.

csatib02 (changes requested):

.github/workflows/build-and-push-components.yml
zsoltkacsandi: I worked based on the previous version, we did not have suffix there: https://github.com/openclarity/vmclarity/pull/1768/files

But I am not sure why this suffix was added here.

csatib02: https://github.com/openclarity/vmclarity/pull/1848#discussion_r1656824040

.github/workflows/build-and-push-components.yml
zsoltkacsandi: It is a perfectly valid use case to trigger a workflow on multiple events: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-activity-types-and-filters-with-multiple-events, that does not mean it would have "multiple responsibilities".

If you are referring to the SRP, what exactly could be the reasons for the changes that are different in the two cases? Clarifying this would help to decide whether if it is worth extracting the manual job into a separate file, the "it should have one responsibility" is quite general and not so actionable.

.github/workflows/build-and-push-components.yml
zsoltkacsandi: Done, thank you.

.github/workflows/build-and-push-components.yml
zsoltkacsandi: Thanks, the dev suffix has been added.

.github/workflows/build-and-push-components.yml
zsoltkacsandi: > BTW, Your previous implementation of the manual build and push was something very similar to what I implemented.

My previous implementation was a quick fix for a problem that an external team had, not a well thought-out solution.

> If you look at any of the workflow files right now, it's immediately clear what it does, and how that thing is done.

If you look at this at this file, it still does the same, based on the inputs it will build and push the vmclarity components. The only change that now it can be triggered on multiple events. The inputs are the same, everything is the same.

Either way, I will put the workflow_dispatch into a separate file.

akijakya (approved):

csatib02 (approved):

ramizpolic (approved):

PR #1898, opened 2024-07-10
fix(plugin): fix plugin shutdown

zsoltkacsandi:
## Description

- introducing shutdown grace period in family manager to wait for the families to shut down properly in case of context cancellation (only needed for the plugin family)
- shutting down plugins if a panic occurs
- enabling init process for plugin containers to prevent zombie processes
- plugin family runner has been reworked to shut down the plugin container *before* it sends back the results
- passing context to Stop and Remove functions *without* cancel (otherwise in case of a context cancellation, these would be cancelled as well and wouldn't run)

## Type of Change

- [X] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


ramizpolic (dismissed):

plugins/runner/internal/runtimehandler/docker/handler.go
```diff
@@ -107,6 +108,7 @@ func (h *containerRuntimeHandler) Start(ctx context.Context) error {
 				},
 			},
 			Mounts: []mount.Mount{*scanDirMount},
+			Init:   &init,
```
ramizpolic, line 111: we can do `to.Ptr(true)`

scanner/families/plugins/runner/runner.go
```diff
@@ -67,77 +69,113 @@ func (s *Scanner) Scan(ctx context.Context, inputType common.InputType, userInpu
...
+		resChan <- res
+	}(ctx)
+
+	select {
+	case <-ctx.Done():
+		shutdownRunner(ctx)
+		return nil, errors.New("plugin context cancelled")
+	case r := <-resChan:
+		shutdownRunner(ctx)
+		return &types.ScannerResult{
```
ramizpolic, line 174: we can receive error here but it is not handled, we always assume the received result is valid

ramizpolic (changes requested):

scanner/families/plugins/runner/runner.go
zsoltkacsandi: Not exactly sure that I understand. Should I return an empty `ScannerResult` struct here?

`return &types.ScannerResult{}, errors.New("plugin context cancelled")`

ramizpolic: resChan result contains error which is not checked

zsoltkacsandi: Ah, thanks! Fixed.

ramizpolic (approved):

PR #1920, opened 2024-07-15
feat(scanner): add default clamd configuration

zsoltkacsandi:
## Type of Change

- [ ] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/cli/cmd/scan/scan.go
+++ b/cli/cmd/scan/scan.go
@@ -179,6 +179,9 @@ func loadConfig(cfgFile string) *scanner.Config {
 	err := viper.ReadInConfig()
 	cobra.CheckErr(err)
 
+	// Set defaults
+	viper.SetDefault("malware.scanners_config.clam.use_clam_daemon", true)
+
 	// Load config
 	config := &scanner.Config{}
 	err = viper.Unmarshal(config)
--- a/scanner/families/malware/clam/clam.go
+++ b/scanner/families/malware/clam/clam.go
@@ -17,6 +17,7 @@ package clam
 
 import (
 	"context"
+	_ "embed"
 	"errors"
 	"fmt"
 	"os"
@@ -37,6 +38,9 @@ import (
 )
 
 const ScannerName = "clam"
+
+//go:embed templates/clamd.conf
+var clamdConfigTemplate string
 
 type Scanner struct {
 	config config.Config
@@ -321,10 +325,7 @@ func syncFreshclamConfig(configPath, alternativeMirrorURL string) error {
 
 func createDefaultClamdConfig(config config.Config) error {
 	t := template.New("clamd.conf")
-	t, err := t.Parse(`LocalSocket {{.LocalSocket}}
-{{range .ExcludedPaths}}
-ExcludePath {{.}}
-{{end}}`)
+	t, err := t.Parse(clamdConfigTemplate)
 	if err != nil {
 		return fmt.Errorf("unable to parse template: %w", err)
 	}
--- a/scanner/families/malware/clam/clam.go
+++ b/scanner/families/malware/clam/clam.go
@@ -21,7 +21,10 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"slices"
 	"strings"
+	"text/template"
 
 	"github.com/openclarity/vmclarity/core/log"
 	"github.com/openclarity/vmclarity/scanner/common"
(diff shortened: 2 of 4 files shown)


cli/cmd/scan/scan.go
```diff
@@ -179,6 +179,9 @@ func loadConfig(cfgFile string) *scanner.Config {
 	err := viper.ReadInConfig()
 	cobra.CheckErr(err)
 
+	// Set defaults
+	viper.SetDefault("malware.scanners_config.clam.use_clam_daemon", true)
```
ramizpolic, line 183: we can replace `use_clam_daemon` with `use_native_clamscan`. its not a breaking change yet since we still havent released. this would default to using daemon for scanning. 

use another flag that defaults to daemon scan if you have a better suggestion, but we can definitely remove this part to avoid creating partial defaults scattered around the code base

scanner/families/malware/clam/clam.go
```diff
@@ -201,12 +208,19 @@ func prepareClamDaemon(ctx context.Context, config config.Config) error {
 	logger.Debugf("found clamd binary at: %s", clamDaemonPath)
 
 	// Define default clamd args to run
-	var args []string
+	args := []string{
+		"--config-file", config.ClamDaemonConfigPath,
+	}
```
ramizpolic, line 213: wont work because we create and set config path below and then append it to args again. either reverting to previous changes or moving the whole default clamd configuration code above this works

```suggestion
	// clamd config created here
	
	args := []string{
		"--config-file", config.ClamDaemonConfigPath,
	}
```

ramizpolic: we can revert to previous version actually, looks better
```suggestion
	var args [] string
```

cli/cmd/scan/scan.go
zsoltkacsandi: Indeed, now I see in the history this is a new config option. Replaced use_clam_daemon with use_native_clamscan.

scanner/families/malware/clam/config/config.go
```diff
@@ -26,9 +26,9 @@ const (
 )
 
 type Config struct {
-	// UseClamDaemon indicates that scanning should use daemon (clamdscan) command
-	// instead of native (clamscan) command.
-	UseClamDaemon                 bool     `yaml:"use_clam_daemon" mapstructure:"use_clam_daemon"`
+	// UseNativeClamscan indicates that scanning should use the native (clamscan) command
+	// instead of daemon (clamdscan) command.
+	UseNativeClamscan             bool     `yaml:"use_native_clamscan" mapstructure:"use_native_clamscan"`
```
ramizpolic, line 31: please rebase to pull the latest JSON tags

scanner/families/malware/clam/clam.go
```diff
@@ -176,7 +183,7 @@ func (s *Scanner) runClamScanWithDaemon(ctx context.Context, fsScanPath string)
 		/* If the error is that malware was found, this is not an actual error, Clamd returns
 		   a non 0 exit code when malware was found */
 		var runError utils.CmdRunError
-		if !errors.As(err, &runError) || !strings.Contains(string(runError.Stdout), constants.MalwareDetectedIndication) {
+		if !errors.As(err, &runError) || !strings.Contains(string(runError.Stdout), constants.ScanSummaryText) {
```
ramizpolic, line 186: please note that we don't display summary for clamdscan (the `--no-summary` flag is passed to not display), so we would need to remove the flag to avoid erroring here

Did you have a chance to test both clamscan and clamdscan scanners with a a) no malwares and b) one specific malware?

scanner/families/malware/clam/clam.go
```diff
@@ -188,7 +195,7 @@ func (s *Scanner) runClamScanWithDaemon(ctx context.Context, fsScanPath string)
 
 func prepareClamDaemon(ctx context.Context, config config.Config) error {
```
ramizpolic, line 196: we should either config as a reference here or switch to a method-based approach (same for `prepareFreshclam`). my suggestion is to switch this to
```suggestion
func (s *Scanner) prepareClamDaemon(ctx context.Context) error {
```

the `New` method can easily be changed to
```go
func New(ctx context.Context, _ string, config types.ScannersConfig) (families.Scanner[*types.ScannerResult], error) {
	scanner := &Scanner{
		config: config.Clam,
	}

	// Prepare freshclam to sync database data
	if err := scanner.prepareFreshclam(ctx); err != nil {
		return nil, fmt.Errorf("failed to run freshclam: %w", err)
	}

	// Prepare clam daemon
	if err := scanner.prepareClamDaemon(ctx); err != nil {
		return nil, fmt.Errorf("failed to run freshclam: %w", err)
	}

	return scanner, nil
}
```

scanner/families/malware/clam/clam.go
zsoltkacsandi: Thanks, I've fixed it.

I've tested clamscan and clamdscan scanning with no malwares and with the EICAR test file.

ramizpolic (approved):

PR #1971, opened 2024-07-31
fix(scanner): enrich vulnerability data instead of using first vulnerability

paralta:
## Description

Currently, we use the first vulnerability in merged vulnerabilities results https://github.com/openclarity/vmclarity/blob/fb54c41f560e393dc08cab39a8e65747b246638f/cli/presenter/apimodel.go#L72 which may cause inconsistencies in final results when there is more than one entry for a particular vulnerability id (CVE).

These code changes allow us to merge vulnerabilities with the same ID into a single vulnerability by enriching the vulnerability data with the data from all the different entries. 

## Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


scanner/families/vulnerabilities/types/vulnerability.go
```diff
@@ -99,60 +122,77 @@ type VulnerabilityDiff struct {
 	ASCIIDiff   string                 `json:"asciiDiff"`
 }
 
-type MergedVulnerability struct {
-	ID            string              `json:"id"` // Merged vulnerability ID used in DiffInfo - THIS IS NOT THE CVE ID
-	Vulnerability Vulnerability       `json:"vulnerability"`
-	Diffs         []VulnerabilityDiff `json:"diffs"`
-	ScannersInfo  []ScannerInfo       `json:"scanners"`
-}
+func mergeSlicesWithoutDuplicates[T comparable](slice1, slice2 []T) []T {
```
ramizpolic, line 125: consider moving this to `to` package, you can then use it as `to.UniqueSlice(append(sliceA, sliceB))`
```suggestion
// UniqueSlice returns a slice without duplicate elements.
func UniqueSlice[T comparable](items []T) []T {
	var filtered []T

	unique := make(map[T]bool, len(items))
	for _, item := range items {
		if !unique[item] {
			filtered = append(filtered, item)
			unique[item] = true
		}
	}

	return filtered
}
```

scanner/families/vulnerabilities/types/vulnerability.go
```diff
@@ -99,60 +122,77 @@ type VulnerabilityDiff struct {
...
 
-func (mv *MergedVulnerability) AppendScannerInfo(info ScannerInfo) *MergedVulnerability {
-	mv.ScannersInfo = append(mv.ScannersInfo, info)
-	return mv
-}
+	elementMap := make(map[T]bool)
+	result := []T{}
 
-func (mv *MergedVulnerability) AppendDiffInfo(diff VulnerabilityDiff) *MergedVulnerability {
-	mv.Diffs = append(mv.Diffs, diff)
```
ramizpolic, line 115: are we still using `VulnerabilityDiff`? if not, we can remove dead code

scanner/families/vulnerabilities/types/vulnerability.go
ramizpolic: you can include additional method to allow generics too which should simplify existing  `mergeSlicesWithoutDuplicates` and `mergeCVSSSlicesWithoutDuplicates` in the same way

```go
func UniqueGenericSlice[T any](items []T, getKey func(T) string) []T {
	var filtered []T

	unique := make(map[string]bool, len(items))
	for _, item := range items {
		key := getKey(item)
		if key != "" && !unique[key] {
			filtered = append(filtered, item)
			unique[key] = true
		}
	}

	return filtered
}
``` 

Usage example:
```go
mergedLinks := to.UniqueSlice(append(vulnerability.Links, otherVulnerability.Links...))
mergedCVSSs := to.UniqueGenericSlice(append(vulnerability.CVSS, otherVulnerability.CVSS...), func(cvss CVSS) string {
	b, err := json.Marshal(c)
	if err != nil {
		log.Errorf("failed to marshal cvss. c=%+v: %v", c, err)
		return "" // skip
	}

	return string(b)
})

```

ramizpolic: @paralta please consider using a better name for this method, this one might be difficult for usage, e.g. `UniqueSliceFromKey`

ramizpolic (approved):
LGTM!

